PR #5, opened 2026-08-02
test(kit): assert the event stream stays open, not just that it acked

### LINKED ISSUE #2: the subscription test never asserts that the subscription stays open
Found by an independent review seat reviewing v1a. Still present at `ae9a118`.

## The defect

`testSubscribeAcknowledgesAndStaysOpen` claims two properties in its name and asserts only one:

```swift
for try await line in stream {
    if line == .subscriptionStarted { sawAck = true; break }
}
XCTAssertTrue(sawAck, "server should acknowledge the subscription")
```

It **breaks at the acknowledgement**. A server that answered `subscription_started` and then immediately closed the connection would pass this test unchanged — which is precisely the behaviour the name promises to rule out.

That matters because "the event socket is persistent" is a load-bearing claim in this design. It is why `HerdrTransport` has two methods rather than one, why the transport plans one long-lived event channel alongside N ephemeral request channels, and it is written into the README as a measured fact. The property is real — it was verified by hand against a live server, which held a stream open through a 20-second idle — but **nothing in the suite would notice if it stopped being true.**

## Why I am filing rather than quietly fixing

This is the fifth test I have shipped this week that asserts less than its name implies. The others: a single-shot test that did two independent round trips and proved nothing; a 50% flake from one `read()` leaving a trailing newline; a test that codified a freeze as correct behaviour; and a stream-cancellation test that passed without the fix. The pattern is worth a visible record rather than a silent commit, because the failure mode is always the same — the test observes something adjacent to the property and I read the green as confirmation.

## Acceptance

Read the acknowledgement, then continue reading and require the stream to still be open after an idle interval — i.e. assert that a subsequent read neither returns nil nor throws within the window. Kill the server-side stream (or point at a stub that closes after the ack) and the test must fail. Confirm any mutation actually applies before trusting a green run.

HERDR-APP:

jerryfane:
Closes #2. Goal v1b, task 5 — independent of tasks 1–4, run in parallel per the goal.

## WHAT

`testSubscribeAcknowledgesAndStaysOpen` now asserts what its name claims. It previously broke out of its loop at the acknowledgement, so a server that answered `subscription_started` and then closed immediately would have passed.

## WHY

That the event socket is **persistent** is load-bearing: it is why `HerdrTransport` has two methods, why the transport plans one long-lived event channel alongside N ephemeral request channels, and it is stated in the README as measured fact. **Nothing in the suite would have noticed if it stopped being true.**

## AXIS — what each test pins

- `testSubscribeAcknowledgesAndStaysOpen` — **the stream is still open after an idle interval.** Not that an ack arrived; that the connection survives the window.
- `testTheStayOpenCheckCatchesAServerThatClosesAfterTheAck` — **the new assertion can actually fail.** A stub that acks then closes must be reported as ended early.

That second test exists deliberately. Without it, *"the assertion would have caught it"* is an unverified claim **about a test** — which is precisely the shape of defect this project keeps shipping. Thirteen tests this week asserted less than their names implied; the remedy is to test the detector, not to trust it.

## RESULTS

- **47 tests, 0 failures.** Build warning-clean.
- **Mutation**, with an applied-assertion that fails loudly if the target is absent: defeating early-close detection (`markEnded` made a no-op) **fails** the stub test.
- Live test **repeated 12×, 0 failures** — timing-sensitive, per goal rule.

One process note: my first mutation attempt used a `\|` filter that swift-test treats as a substring, so it matched **zero tests** and reported green. Caught and redone against the real target. That is the fourth silently-inert mutation this week, and the reason the goal requires an applied-assertion rather than a green run.

## RISK

- The idle window is **1.5s live / 1.0s stub**. Long enough to catch an immediate close, not long enough to catch a server that closes after several seconds. A longer window would be strictly better evidence at the cost of suite time; 12 clean repeat runs say the current value is not flaky.
- Proves the stream **stays open**, not that it **delivers events**. Driving a real turn to completion and asserting the event arrives is separate coverage and is not claimed here.

HERDR-APP:

### DIFF
--- a/Tests/HerdrKitTests/HerdrKitTests.swift
+++ b/Tests/HerdrKitTests/HerdrKitTests.swift
@@ -206,6 +206,16 @@ final class LiveServerTests: XCTestCase {
         )
     }
 
+    /// herdr-ios#2. The previous version broke out of the loop at the
+    /// acknowledgement, so a server that answered `subscription_started` and
+    /// then closed immediately would have passed — the exact behaviour the name
+    /// promises to rule out.
+    ///
+    /// That the event socket is PERSISTENT is load-bearing: it is why
+    /// `HerdrTransport` has two methods, why the transport plans one long-lived
+    /// event channel alongside N ephemeral request channels, and it is written
+    /// in the README as measured fact. Nothing in the suite noticed if it
+    /// stopped being true.
     func testSubscribeAcknowledgesAndStaysOpen() async throws {
         try XCTSkipIf(socketPath == nil, "no live herdr server")
         let c = try client()
@@ -215,11 +225,27 @@ final class LiveServerTests: XCTestCase {
             Subscription(.paneAgentStatusChanged, paneID: pane),
             Subscription(.paneTurnCompleted, paneID: pane),
         ])
-        var sawAck = false
-        for try await line in stream {
-            if line == .subscriptionStarted { sawAck = true; break }
-        }
-        XCTAssertTrue(sawAck, "server should acknowledge the subscription")
+
+        let seen = await observeSubscription(stream, idleWindow: 1.5)
+        XCTAssertTrue(seen.sawAck, "server should acknowledge the subscription")
+        XCTAssertFalse(
+            seen.endedEarly,
+            "the event stream must still be open after an idle interval; it closed instead"
+        )
+    }
+
+    /// Proves the assertion above can actually fail. A stub that acknowledges
+    /// and then closes is exactly the server behaviour the old test could not
+    /// distinguish from a healthy one.
+    func testTheStayOpenCheckCatchesAServerThatClosesAfterTheAck() async throws {
+        let stream = HerdrClient(transport: AckThenCloseTransport())
+            .subscribe([Subscription(.paneTurnCompleted, paneID: "p1")])
+        let seen = await observeSubscription(stream, idleWindow: 1.0)
+        XCTAssertTrue(seen.sawAck, "the stub does acknowledge")
+        XCTAssertTrue(
+            seen.endedEarly,
+            "a stream that closes after the ack must be detected as ended early"
+        )
     }
 }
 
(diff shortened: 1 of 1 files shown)


jerryfane:
> Runtime: `claude`
> Job: `local-review-herdr-ios-review-opus-18c7d424e3a095f9`

**Decision:** `failed`
**Summary:** --dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons: exit status 1
**Diagnostics:** --dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons: exit status 1

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7d4dc6423e875`

**Decision:** `approved`
**Summary:** Head ea1683b correctly pins the stated persistence axis. The snapshot occurs before consumer cancellation, so cancellation cannot produce a returned endedEarly=true; ObservationBox locking is sufficient, and the where clause consumes rather than loses non-ack lines. The finite window proves neither indefinite persistence nor event delivery, but those limitations are honestly scoped and it meaningfully catches acknowledgement followed by immediate closure. AckThenCloseTransport faithfully reproduces that observable sequence. No third acknowledgement-break pattern or malformed multi-test filter was found.

**Tests Run**
- Confirmed exact HEAD ea1683b4478ebc7ef36e35c5aecdfdb3c66d63e8 and clean worktree.
- swift build: passed; only sandbox SwiftPM cache warnings.
- Exact stub test selection executed one test and passed.
- Mutation verification: replaced markEnded with a no-op, asserted the mutation was present, and ran exactly LiveServerTests.testTheStayOpenCheckCatchesAServerThatClosesAfterTheAck; one selected test failed at the intended endedEarly assertion. Source was restored.
- LiveServerTests.testSubscribeAcknowledgesAndStaysOpen passed 10/10 in separate exact-selector XCTest processes without skips.
- swift test: passed 47/47; all live tests executed with zero skips.
- git diff --check and final clean-worktree verification passed.
- Searched repository commands and tests for zero-match backslash-pipe filters and additional acknowledgement-only subscription assertions; none found.

PR #7, opened 2026-08-02
docs: SSH transport cost, measured

jerryfane:
Goal v1b, task 2. **Findings, not code** — the measurement harness imports `SSHTransport`, which is unmerged and frozen behind #4, so a code PR would stack and wait. The findings don't need to.

## WHAT

A measured breakdown of where SSH transport cost actually goes, and what it means for the connection design.

| phase | cost | nature |
|---|---|---|
| TCP connect | 0.1 ms | negligible |
| handshake / kex | 18.6 ms | **real crypto** |
| publickey auth | 20.4 ms | **real crypto** |
| **first channel of a session** | **~100–1900 ms, variable** | **cause unknown** |
| subsequent channels | 0.1 ms | free |
| missing `TCP_NODELAY` | ~30 ms | recovered |

## WHY

Task 3 was to pool connections. Sizing that pool required knowing what it would actually hide — and the original premise (channel-open is the dominant cost) turned out to be wrong.

## The finding

**The first channel of a session is expensive; every subsequent channel on it is free.** Since `roundTrip` opens a session per request, **every request pays a once-per-session cost**.

## How it was nearly missed

Three rounds of percentile reporting found nothing — channel-open looked like a modest `p50 = 40.7 ms`. Printing the **sequence** found it in one look:

```
1804.6  0.2  0.1  0.1  0.1  0.1  0.1  0.1  0.1  0.1  0.1  0.1
```

A p50 over a bimodal distribution isn't a weak signal, it's a **confident wrong one**. A separate "pathological p95 of 1958 ms", logged for a while as an unrelated defect, dissolved into the same phenomenon.

## Ruled out

- **herdr** — a trivial echo socket with zero herdr code paid *more* (1877.8 ms vs 99.7 ms)
- **`UseDNS`** — unset, defaults off
- **GSSAPI** — `gssapiauthentication no`
- **herdr's two-write pattern** — that claim was **retracted** (compared two different programs; real figure ~2x, sub-ms). `jerryfane/herdr#29` stays parked.

## Recorded as unknown

**Why the first-channel cost varies ~100 ms to ~1900 ms between runs.** It resembles a timeout more than work, but that's untested and labelled as a hypothesis. The chase stopped because the design doesn't depend on the cause — the cost is per-session either way.

## RESULTS

No code, so no tests. Every number is reproducible against the live herdr server; the method for each is described in the document.

One correction is embedded rather than hidden: I earlier reported the 1.8 s as a *herdr-path* fact. A later run gave herdr 99.7 ms and the echo path 1877.8 ms, so the magnitude is unstable and my firm number came from a single sequence. Corrected here.

## RISK

- **Loopback is a floor**, not a cellular prediction. Stated in the document.
- **The cause is genuinely unknown.** If it turns out to be per-*connection* rather than per-*session*, task 3's design would change again — this is the fourth revision already.
- Numbers are point-in-time on one machine; the *shape* is what should be relied on, not the magnitudes.

HERDR-APP:


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7d73ae250158b`

**Summary:** Session pooling is directionally supported, but the document misidentifies a post-authentication readiness delay as an intrinsic first-channel cost and does not provide evidence for pool sizing.

**Findings**
- **The claim that every session's first channel is expensive is false. Thirty immediate post-authentication opens took 89.6-102.1 ms, but thirty sessions aged 100 ms before their first channel took only 0.28-0.54 ms. Pre-opened sessions ranged from 0.22 ms to 2594 ms.** (HIGH)
  - concrete_change: Describe this as a post-authentication readiness delay and require task 3 to prewarm or validate sessions before checkout.
  - location: docs/transport-measurements.md:17
- **The evidence supports pooling sessions rather than channels, but does not determine pool size. There are no concurrency, throughput, queueing, server-limit, or long-term lifetime measurements.** (HIGH)
  - concrete_change: Narrow the conclusion to the pooling unit and prewarming requirement; measure concurrent load and session lifetime before selecting pool size or eviction policy.
  - location: docs/transport-measurements.md:74
- **The p50 lesson conflates datasets. The displayed sequence of eleven approximately 0.1 ms samples and one 1804.6 ms sample has a median near 0.1 ms, not 40.7 ms. A median is not inherently wrong; the problem is aggregating heterogeneous session states.** (MEDIUM)
  - concrete_change: Identify the dataset that produced 40.7 ms, publish its samples, and recommend stratifying by session age and channel ordinal.
  - location: docs/transport-measurements.md:36
- **The echo comparison proves herdr is not necessary for the delay, but separate 99.7 ms and 1877.8 ms runs do not prove both magnitudes have one cause. Same-session order reversal showed whichever target opened first paid 87-104 ms and the second took 0.07-0.11 ms; the older 1.8 second outlier remains unexplained.** (MEDIUM)
  - concrete_change: Use the narrower claim that the phenomenon occurs independently of herdr and include the target-order control.
  - location: docs/transport-measurements.md:54
- **The PR does not preserve the harness, raw samples, complete sample counts, ordering, or meaningful conditional p50/p95 distributions. API-level handshake and authentication timing also does not isolate cryptographic CPU work from protocol I/O and scheduling.** (MEDIUM)
  - concrete_change: Document or check in the exact procedure and raw data, and label the rows as handshake-phase and authentication-phase latency unless cryptographic work is separately profiled.
  - location: docs/transport-measurements.md:3

**Tests Run**
- Verified exact head 93a826883fe43337adb78c5a5001fc043e923f95.
- Measured 30 immediate first-channel opens: p50 96.374 ms, p95 100.700 ms, range 89.593-102.124 ms.
- Measured 30 first-channel opens after 100 ms session age: p50 0.396 ms, p95 0.517 ms, range 0.279-0.537 ms.
- Measured eight pre-authenticated sessions: first channels ranged 0.221-2594.332 ms; second channels ranged 0.058-0.213 ms.
- Reversed herdr and echo order across ten sessions: first target took 87.314-104.091 ms; second took 0.069-0.112 ms.
- Verified one warmed session remained cheap after 1, 3, and 10 seconds idle.
- swift build passed.
- swift test passed: 47 tests, 0 failures.
- git status, git diff --check, and git diff --exit-code confirmed a clean worktree after removing the audit harness.

**Needs**
- Rewrite the central claim around session readiness and prewarming.
- Preserve reproducible raw measurements and conditional distributions.
- Measure concurrency and longer session lifetimes before selecting pool size.
- The cause of rare 585-2594 ms outliers and behavior beyond 10 seconds or across mobile networks remain unverified.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7d9904b6d37f7`

**Summary:** At 0d7a301, the correction is not consistently propagated and several causal claims exceed the measurements. Build and 47 tests pass.

**Findings**
- HIGH docs/transport-measurements.md:17,43,78,88 — The document declares the intrinsic first-channel claim false, then repeats it as a variable once-per-session cost and uses it to justify pooling. Rewrite the breakdown, unknowns, and design consequence around the observed immediate-post-authentication timing effect.
- HIGH docs/transport-measurements.md:31-37 — The pre-opened-session result was not approximately 0.22 ms overall; measured first-open outliers reached 2594 ms. The 100 ms control establishes timing correlation in one batch, not an observable readiness state or a universal cure. Report full ranges and label readiness as an untested hypothesis.
- MEDIUM docs/transport-measurements.md:48-62 — The original 40.7 ms harness retained only summary statistics for 20 consecutive opens on one shared session. Without its raw sequence or age metadata, attributing that result specifically to mixed immediate and aged states is retrospective inference. Require preserved raw samples and explicit state stratification; sequence order is diagnostic, not proof.
- MEDIUM docs/transport-measurements.md:66-67 — A separate echo run being slower proves herdr is not necessary for the delay, but does not categorically exclude an independent herdr-path contribution. Narrow the claim or cite the controlled target-order experiment.
- LOW docs/transport-measurements.md:15-16 — Timing libssh2 handshake and authentication phases does not prove all latency is cryptographic computation. Label these as measured phase latencies unless profiling supports the stronger attribution.

**Tests Run**
- Confirmed exact head 0d7a30142549a98d8cbbe9b17a97605c3e6d6dd2 and clean worktree.
- swift build passed; only environmental SwiftPM cache-permission warnings were emitted.
- swift test passed 47 tests with 0 failures, including live-server tests.
- git diff --check 93a8268..HEAD passed.
- Recovered and audited the original 40.7 ms measurement harness.

**Needs**
- Propagate the correction through all sections, report the complete controlled data, and separate measured timing effects from causal hypotheses.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7daf68fc51ceb`

**Summary:** The design consequence is materially improved, and ~96 ms is acceptably scoped as the p50 of one 30-sample batch. However, the document at 91409eb still contains the retracted 100–1900 ms first-channel claim, an incorrect pre-opened-session result, and several causal conclusions unsupported by the measurements.

**Findings**
- HIGH docs/transport-measurements.md:81-85 — Propagation remains incomplete. This section still says the first channel intrinsically costs 100–1900 ms and locates the effect in the first direct-streamlocal forward. Replace it with the controlled observation: one 30-session loopback batch showed immediate opens at 89.6–102.1 ms and 100-ms-aged opens at 0.28–0.54 ms; the mechanism remains unknown.
- HIGH docs/transport-measurements.md:32-38 — The pre-opened-session result is still incorrectly summarized as approximately 0.22 ms; that control included first-open outliers up to 2594 ms. Also, 'waiting 100 ms makes the cost vanish' and 'the session needs a moment to become ready' turn one controlled batch into a universal mechanism. Remove or report the complete pre-opened distribution and describe readiness as an operational hypothesis, not an observed state.
- MEDIUM docs/transport-measurements.md:93-104 — Sacrificial-channel prewarming is supported by the cheap subsequent-channel sequences, but 'prewarming addresses it whatever the cause' is too strong and conflicts with the falsifier immediately below. Say that prewarming addressed the observed effect in these runs and must be revalidated under the deployment conditions.
- MEDIUM docs/transport-measurements.md:51-65 — The original 40.7 ms harness preserved only summaries for 20 sequential opens on one shared session. Its raw sequence and session-age metadata were not retained, so attributing that median specifically to mixed immediate and aged states remains retrospective inference. Preserve raw samples and state metadata; sequence order is diagnostic, not proof of heterogeneous conditions or bimodality.
- MEDIUM docs/transport-measurements.md:69-70 — One echo observation exceeding one herdr observation proves herdr is not necessary for a large delay, not that herdr cannot contribute independently. Narrow the claim accordingly or cite the controlled target-order experiment.
- LOW docs/transport-measurements.md:15-16 — API phase timing measures handshake and authentication latency but does not establish that all of it is cryptographic computation. Rename these as measured phase latencies unless profiling supports the stronger attribution.

**Tests Run**
- Confirmed exact head 91409ebf62c8aa80e315dac73e9124331f0e5c6c and a clean worktree.
- swift build passed; only environmental SwiftPM cache-permission warnings were emitted.
- swift test passed 47 tests with 0 failures, including live-server tests.
- git diff --check 0d7a301..HEAD passed.
- Searched the complete document for first-channel, once-per-session, readiness, prewarming, pooling, and outlier terminology.

**Needs**
- Remove the remaining retracted passage, correct the pre-opened data, and scope causal statements to what the controlled runs actually establish.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7dbdb0eadfb40`

**Summary:** The whole-file rewrite resolves the six prior findings and accurately scopes the 96 ms figure to one 30-sample loopback batch. Two remaining statements still outrun the evidence: one absolute causal summary and a falsifier that tests ageing rather than the proposed sacrificial-channel prewarming intervention.

**Findings**
- MEDIUM docs/transport-measurements.md:30-32 — 'The cost tracks how long the session has existed, not the channel's position' is still stronger than the evidence and conflicts with the careful association-only language below it. The controls show that channel ordinal alone is insufficient and that elapsed post-authentication time was strongly associated with latency in one batch; they do not establish what the cost tracks generally. Scope this sentence to that observation.
- MEDIUM docs/transport-measurements.md:118-123 — The cheap falsifier does not test the stated design intervention. The design defines prewarming as opening and freeing a sacrificial channel, but the experiment compares immediate versus merely aged first-channel opens. It can falsify waiting, not sacrificial-channel prewarming. 'Outliers persist at the same rate' also lacks a sample count and decision threshold. Specify randomized fresh-session groups for immediate request, age-only request, and request after a sacrificial channel; retain raw timings and ages, and predefine the sample size and result that rejects prewarming. This remains runnable with a measurement harness and requires no pool implementation.

**Tests Run**
- Confirmed exact head 5bdd8a992c71dc91e9ddce321915d69e05b3d718 and a clean worktree.
- Audited the complete rewritten document and searched all absolute, causal, readiness, pooling, outlier, and falsifier wording.
- swift build passed; only environmental SwiftPM cache-permission warnings were emitted.
- swift test passed 47 tests with 0 failures, including live-server tests.
- git diff --check 91409eb..HEAD passed.

**Needs**
- Scope the remaining absolute timing statement and make the falsifier test the actual sacrificial-channel prewarming treatment with a defined rejection criterion.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7dd7d25cc968b`

**Summary:** The rewritten evidence claims are now appropriately scoped, but the proposed experiment still cannot produce an unambiguous decision in every outcome, and prewarming remains prescribed elsewhere despite being labelled untested.

**Findings**
- **The decision rules overlap and leave outcomes uncovered. For A=100ms, B=80ms, C=80ms, C simultaneously qualifies for adoption, rejection as within 20ms of A, and preference for B because C approximates B. For A=100ms, B=50ms, C=70ms with a lower outlier rate, neither adoption nor rejection applies. The meaning of approximately equal and the p95 calculation convention are also undefined.** (high)
  - concrete_change: Replace the predicates with an ordered, exhaustive decision tree; make the 20ms boundary exclusive on one side; define equality tolerance and the p95 estimator; enumerate the selected result for every relationship among A, B, and C.
  - location: docs/transport-measurements.md:152
- **The retained fields can calculate a p95 and outlier rate only after latency is defined. The document does not specify whether latency means channel-open duration, full request round-trip, or another interval. The decision also ignores how long the sacrificial channel itself takes, allowing C to win on request latency while imposing a large pool-replenishment delay.** (medium)
  - concrete_change: Define exact measurement start and end points, retain sacrificial-open duration or total time-to-ready, and include that cost in the selection rule.
  - location: docs/transport-measurements.md:146
- **The extra herdr connection cost is recorded but has no decidable effect: 'counts against C independently' supplies neither a threshold nor an override. The experiment can therefore satisfy the latency rule while leaving adoption unresolved on its stated second axis.** (medium)
  - concrete_change: Predeclare how connection count affects the verdict—for example, reject C when it creates an additional server connection unless it beats B by a specified margin—or explicitly make it a separately reported non-selection metric.
  - location: docs/transport-measurements.md:164
- **Prewarming is still presented as the design response, and lines 117-118 still instruct 'Prewarm before checkout,' although lines 60-66 correctly say it is unsupported and must be tested. The demotion has not propagated through the whole document.** (medium)
  - concrete_change: Describe age-gating and sacrificial-channel prewarming consistently as competing hypotheses until the experiment selects one.
  - location: docs/transport-measurements.md:56
- **The sibling memo says C must beat age-gating B by a worthwhile margin, while this document permits adoption when C is merely no worse than B and separately says a tie should prefer B. The two documents therefore prescribe different outcomes. The sibling memo also retains stale claims about intrinsic crypto cost, a 100-1900ms stable shape, and herdr being excluded.** (medium)
  - concrete_change: Make one document the canonical experiment and decision rule, then update the sibling memo to reference it and adopt this document's scoped evidence statements.
  - location: docs/connection-pool-design.md:44

**Tests Run**
- swift build — passed; only SwiftPM cache-permission environment warnings
- swift test — passed, 47 tests and 0 failures; live-server tests executed
- git diff --check 5bdd8a9..HEAD — passed
- Verified clean worktree at a01f53b and audited the complete measurement document plus task/3-pool-design at 0a3a758

**Needs**
- An ordered and exhaustive experiment decision rule with defined measurement boundaries, statistical conventions, and treatment-cost thresholds
- Consistent hypothesis-only wording for prewarming across PR #7 and the sibling pool-design memo

jerryfane:
## Head fd2800e — the tree I asked you to check was ambiguous, and I found it

I asked you to verify the tree was genuinely exhaustive because I had "checked it
myself, which is the weakest form of checking available." It was the weakest form
of checking available, and it missed this.

The vetoes were written as fallback steps — *"fall back to the other effective
arm"* — without saying whether the vetoes re-applied to the fallback. Both
readings are defensible and they disagree. **3,072 of 23,328 input combinations
(13.2%) resolve two ways.** So the procedure was not single-valued, while
carrying a paragraph asserting it was.

I found it by enumerating it rather than reading it again.
`scripts/check-decision-rule.py` is committed: it models both the superseded
fallback form and the current one and reports ambiguity counts for each.

**The rule is now flat rather than nested.** Each candidate qualifies on all its
conditions or does not — effective, tail improved, and for C, replenishment
affordable — and then an ordered selection runs over whatever qualified. The
vetoes became qualification conditions precisely so there is no fallback whose
semantics need defining. Same enumeration: 0 ambiguous.

Both cases from your a01f53b verdict still give one answer each, and both give
**B**, unchanged from what I claimed last round.

The thing worth taking from this beyond the rule: "exhaustive" is a property with
23,328 witnesses on this input space, and I had been asserting it from
agreement-with-myself. That is the same shape as the tests this project keeps
catching — a claim whose check does not reach the claim.

HERDR-APP:

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7de8549ea44d1`

**Summary:** The flat control flow is single-valued and both prior reviewer cases resolve uniquely to B. However, the current qualification semantics can reject an ideal treatment, the retained-data contract omits a required metric, and the sibling memo still contradicts the canonical evidence.

**Findings**
- **Requiring a strictly lower >500 ms outlier rate makes both candidates ineligible whenever A records zero outliers, even if they remove the readiness delay completely. For A95=100 ms, B95=C95=5 ms, and all three outlier rates zero, the rule adopts neither and declares the readiness hypothesis failed. The controlled ~96 ms batch could naturally produce exactly this zero-outlier baseline.** (high)
  - concrete_change: Make tail behavior a non-regression condition when A has no measurable tail, or condition strict improvement on A having baseline outliers. Predeclare how sampling uncertainty affects the comparison rather than treating one event in n=100 as conclusive.
  - location: docs/transport-measurements.md:188
- **The retention contract says to retain arm, singular latency, session age, and sequence index, but the decision requires both t_request and C's t_ready p95. Unless t_ready is retained per sample, the replenishment qualification cannot be independently recomputed from the promised raw data.** (medium)
  - concrete_change: Explicitly retain t_request and t_ready for every sample, including the complete sacrificial open-and-free duration for C.
  - location: docs/transport-measurements.md:154
- **The historical model does not cover all defensible readings of the superseded ordered rule. It returns immediately after an outlier fallback, so a fallback to C never reaches the later replenishment veto. A faithful sequential reading disagrees with the script's outcome set in 1,152 of 23,328 sampled inputs. The ambiguity count remains 3,072 under that reading, so the stated 13.2% happens to survive, but the claim that the script models both forms is too strong. The script also only prints discrepancies and always exits successfully.** (medium)
  - concrete_change: Model the additional sequential reading or scope the claim to the two encoded interpretations, and make unexpected ambiguity/count assertions produce a nonzero exit.
  - location: scripts/check-decision-rule.py:20
- **The sibling memo still says the readiness delay is 'Not herdr.' The canonical document establishes only that herdr is not necessary for a large delay and explicitly says the two single observations cannot exclude an independent herdr contribution.** (medium)
  - concrete_change: Replace the absolute exclusion with the canonical scoped claim that herdr is not necessary, while its independent contribution remains unmeasured.
  - location: task/3-pool-design:docs/connection-pool-design.md:216

**Tests Run**
- python3 scripts/check-decision-rule.py — 23,328 combinations; published model reported 3,072 ambiguous (13.2%), flat model reported zero, and both reviewer cases resolved to B
- Independent 23,328-case enumeration — confirmed the 13.2% ambiguity count under a sequential fallback reading, found 1,152 exact outcome-set differences from the committed historical model, and confirmed the flat rule is single-valued
- Semantic decision probes — confirmed zero baseline outliers force neither to qualify despite large p95 improvements
- swift build — passed; only SwiftPM user-cache permission warnings
- swift test — passed, 47 tests and 0 failures; live-server tests executed
- git diff --check a01f53b..HEAD — passed
- Audited the complete canonical document and task/3-pool-design at 67dd818; worktree remained clean

**Needs**
- Revise the tail qualification so zero baseline outliers do not veto an otherwise effective treatment
- Retain both decision metrics explicitly
- Correct or narrow the enumerator's historical-model claim
- Propagate the canonical herdr scope into the sibling memo

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7df28f3caeace`

**Summary:** The zero-outlier hole, retention contract, executable regression check, and sibling memo wording are fixed. Leaving an existing multi-second tail unresolved is acceptable if it remains explicitly reported as a separate open problem. Two documentation/rule inconsistencies remain.

**Findings**
- **The rationale says a one-outlier difference at n=100 is not evidence in either direction, but the rule makes one additional observed outlier a hard veto. For A95=100, B95=5, C95=100 with counts A=0 and B=1, the only effective treatment is rejected solely because of that single event.** (medium)
  - concrete_change: Pre-register a tail non-inferiority tolerance or inference rule consistent with the stated sampling rationale and encode its boundary cases in the checker. Alternatively, if zero observed regressions is deliberately required, remove the claim that one event is not evidence and document the conservative policy explicitly.
  - location: docs/transport-measurements.md:191-203; scripts/check-decision-rule.py:28-29
- **The reporting obligation says the arm was adopted for its median, but the decision procedure never uses a median; it qualifies and selects using p95.** (low)
  - concrete_change: Replace “median” with “p95” so the reported rationale matches the actual decision metric.
  - location: docs/transport-measurements.md:205-208

**Tests Run**
- Confirmed exact head 953ace7fb97225dec31adebfc64f0268869e5cba and clean worktree.
- python3 scripts/check-decision-rule.py: 23,328 combinations, 3,072 historical ambiguities (13.2%), zero current ambiguities, exit 0.
- Applied-asserted mutation restoring strict tail improvement: regression case failed and script exited 1.
- Independently exercised zero-count, one-extra-outlier, unchanged-tail, replenishment-cap, and A95 boundary cases.
- Inspected sibling memo at e06d635: it uses the weaker herdr claim and refers to this document for the selection rule without duplicating thresholds.
- swift build: passed.
- swift test: 47 tests passed, 0 failures, 0 skips; live socket tests executed.
- git diff --check fd2800e..953ace7: passed.

**Needs**
- Align the one-event tail boundary with the document's sampling rationale and checker.
- Correct the median/p95 wording.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7dfdd4294e330`

**Summary:** The previous boundary cases are fixed and mutation-pinned, and the p95 wording is corrected. The cheap experiment may adopt a p95 improvement while reporting the tail as unresolved; that limitation need not block adoption. However, the new +1 rule still presents an operational tolerance as statistically justified and retains stale claims that the tail was not worsened or improved.

**Findings**
- **The revision moves the unsupported evidentiary boundary from one event to two. At n=100 per arm, 0 versus 2 outliers is not a clear regression (two-sided Fisher exact p≈0.50), yet the rule vetoes it and the checker calls it “clear.” The table simultaneously labels A+1 as “tail not worsened,” and line 232 still calls the criterion “tail improved,” although the document correctly admits the experiment cannot establish either claim.** (medium)
  - concrete_change: Describe +1 as a preregistered operational tolerance rather than an evidence threshold. Rename the table row and line 232 to “observed outlier count within tolerance,” and change the checker message to “regression beyond tolerance.” If statistical evidence is intended instead, specify an inference/non-inferiority rule and increase n accordingly.
  - location: docs/transport-measurements.md:191-213,232; scripts/check-decision-rule.py:125

**Tests Run**
- Confirmed exact head 5c46347a04cbf8de45e22a17ba38f8e0d5f65f15 and clean worktree.
- python3 scripts/check-decision-rule.py: 23,328 combinations, 3,072 historical ambiguities (13.2%), zero current ambiguities, exit 0.
- Applied-asserted TAIL_TOLERANCE=0 mutation: one-extra and non-zero-baseline tolerance cases failed, exit 1.
- Applied-asserted TAIL_TOLERANCE=2 mutation: two-extra and past-tolerance cases failed, exit 1.
- Applied-asserted replenishment-cap removal: cap boundary failed, exit 1.
- Computed exact fixed-margin tail comparisons for 0 versus 1–5 observed events; 0 versus 2 has two-sided p≈0.497.
- swift build passed.
- swift test: 47 tests passed, 0 failures, 0 skips; live socket tests executed.
- git diff --check origin/main...HEAD passed.

**Needs**
- Align the tail criterion's labels and justification with its actual role as an operational tolerance, not proof of non-regression.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e013d2e5b151`

**Summary:** The operational-cutoff reframing is not fully propagated. Three remaining statements still overstate what the experiment or decision rule can establish.

**Findings**
- **The document still calls the tail condition “non-regression,” and the checker says the rationale “requires” the tolerance. Both contradict the corrected explanation that +1 is an arbitrary preregistered operational cutoff that establishes neither non-regression nor safety.** (medium)
  - concrete_change: Describe it consistently as an outlier-count tolerance or operational cutoff, including the checker comment.
  - location: docs/transport-measurements.md:194; scripts/check-decision-rule.py:126
- **The claim that n=100 cannot support any evidence-based tail criterion in either direction is too absolute. A sufficiently large difference can be statistically supported with 100 samples. Likewise, saying non-inferiority would require n “far above 100” is unsupported without specifying the baseline rate, margin, alpha, and power. “Regressions gross enough” also treats observed differences as established regressions.** (medium)
  - concrete_change: Narrow the claim to the actual limitation: n=100 does not justify the chosen +1 boundary or a useful rare-tail non-inferiority guarantee. Call counts beyond the cutoff observed differences, and omit the sample-size claim unless backed by a defined power calculation.
  - location: docs/transport-measurements.md:211
- **The selection procedure still treats an operational adoption decision as a causal hypothesis test. “Neither qualifies” can result from the arbitrary tail cutoff or C’s replenishment-cost cap even when a treatment substantially lowers p95, so it does not establish that the readiness hypothesis failed or that the mechanism is unknown. Similarly, A95 ≤20 only means the delay did not reproduce above the threshold in this run.** (medium)
  - concrete_change: Record which operational qualification failed and reserve “hypothesis failed” for a separately defined analysis. Scope the A95 statement to this run and preregistered threshold.
  - location: docs/transport-measurements.md:233

**Tests Run**
- swift build passed; only SwiftPM cache-access warnings were emitted
- swift test passed: 47 tests, 0 failures, 0 skips; live tests executed
- python3 scripts/check-decision-rule.py passed: 23,328 combinations, 3,072 superseded ambiguities, 0 current ambiguities
- Mutation probes for tail tolerances 0 and 2 and removal of the replenishment cap each produced the expected non-zero checker exit
- git diff --check origin/main...HEAD passed
- Verified exact HEAD 5ba64cb76227f4043dad516eb97a357404c88852 and a clean detached worktree

**Needs**
- Revise the three remaining statistical and causal overclaims before approval.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e0c4f7e6319f`

**Summary:** The prior terminology fixes are consistently propagated, and the mutation harness behaves honestly. However, two evidentiary overclaims and the unchanged causal interpretation of the operational decision rule remain.

**Findings**
- **The statement that n=100 “does not support a useful non-inferiority guarantee” still outruns the evidence. Whether a guarantee is useful or supportable depends on the baseline rate, non-inferiority margin, alpha, and power—the exact parameters the following paragraph says have not been defined.** (medium)
  - concrete_change: Limit the claim to the supported point that n=100 does not justify the +1 cutoff. Defer any non-inferiority adequacy claim until its parameters and power calculation exist.
  - location: docs/transport-measurements.md:216
- **The decision procedure still converts an operational adoption outcome into a causal conclusion. Neither arm can qualify solely because of the arbitrary outlier cutoff or C’s replenishment-cost cap even when its p95 improves substantially; that does not establish that the readiness hypothesis failed or that the mechanism is unknown. Likewise, A95 ≤20 only says the delay did not reproduce above the operational threshold in this run.** (medium)
  - concrete_change: Record which operational condition prevented adoption and scope the A95 result to this run. Reserve “hypothesis failed” for a separately defined analysis of treatment effect.
  - location: docs/transport-measurements.md:240
- **The 20 ms margin is justified as roughly the perceptual floor for a person, but no measurement or source supports that human-noticeability claim. It is another load-bearing cutoff presented as evidence-derived when the document only establishes that it was chosen in advance.** (medium)
  - concrete_change: Declare 20 ms as a preregistered operational margin, or provide evidence supporting the perceptual threshold.
  - location: docs/transport-measurements.md:182

**Tests Run**
- python3 scripts/check-decision-rule.py passed: 23,328 combinations, 3,072 superseded ambiguities, 0 current ambiguities
- Mutation harness absent-target probe returned INVALID with status 2
- Mutation harness compile-error probe returned INVALID with status 2 and restored the source
- Mutation harness zero-test filter probe returned INVALID with status 2
- Mutation harness passing-test probe returned SURVIVED with status 1
- Mutation harness assertion-failure probe returned KILLED with status 0
- swift build passed; only sandbox-related SwiftPM cache warnings were emitted
- swift test passed: 47 tests, 0 failures, 0 skips; live tests executed
- git diff --check origin/main...HEAD passed
- Verified exact HEAD 19f34cfccbaa0b6ff34fbfb95933400e039371a0 and a clean detached worktree

**Needs**
- Revise the remaining statistical, perceptual, and causal claims before approval.
- The harness’s 300-second ESCAPED timeout path was inspected but not exercised to completion.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e1135401929b`

**Summary:** The consolidated cutoff table fixes the named document findings, and steps 3–4 remain operational rather than causal. Two residual claims still contradict that framing: one stale checker comment and an under-scoped tail reporting obligation.

**Findings**
- **The checker still says n=100 does not provide “a non-inferiority guarantee for events this rare.” That is the same unsupported directional claim removed from the document: without a baseline rate, margin, alpha, and power, adequacy cannot be asserted either way.** (medium)
  - concrete_change: Limit the comment to the supported statement that n=100 does not justify this +1 boundary; mirror the document’s parameter-dependent wording.
  - location: scripts/check-decision-rule.py:18
- **The tail reporting obligation is narrower than the preceding disclaimer. It only keeps the tail open when A already has outliers and the adopted arm does not reduce them. Therefore zero outliers in A followed by one in the adopted arm receives no warning, while an apparent reduction in tiny counts can escape the obligation even though the procedure explicitly establishes no tail improvement. The operational result can still be read causally as “tail solved.”** (medium)
  - concrete_change: State unconditionally that this procedure makes no tail inference. Always report the raw arm counts, and require a separate powered analysis before claiming improvement or non-regression, regardless of whether the adopted arm’s observed count rises, falls, or remains zero.
  - location: docs/transport-measurements.md:250

**Tests Run**
- python3 scripts/check-decision-rule.py passed: 23,328 combinations, 3,072 superseded ambiguities, 0 current ambiguities
- swift build passed; only sandbox-related SwiftPM cache warnings were emitted
- swift test passed: 47 tests, 0 failures, 0 skips; live tests executed
- git diff --check origin/main...HEAD passed
- Audited the complete decision procedure and reporting section for causal and evidentiary language
- Verified exact HEAD d12c302faf51f2eec6a9fbf81655a1106f1e0c22 and a clean detached worktree

**Needs**
- Align the checker comment with the parameter-dependent document wording.
- Make the no-tail-inference reporting obligation unconditional.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e1a88449205e`

**Summary:** The prior checker and unconditional-tail fixes are correct, and the decision checker passes. Three surviving statements still promote operational choices into evidentiary or causal claims.

**Findings**
- **The p95 operational cutoff is still described as treatment effectiveness.** (medium)
  - concrete_change: Rename these to observational terms such as “p95 margin,” “p95-qualified candidate,” and “candidate meeting the p95 cutoff.” State symmetrically that adoption is an operational selection from this batch, not a causal effectiveness finding.
  - details: The memo labels the arbitrary 20 ms rule an “effectiveness margin,” calls candidates “effective,” and records “effectiveness” as a blocking condition. The checker likewise says “perfect treatment” and “only effective arm.” Yet the memo correctly states that no treatment-effect analysis is performed. Adoption can therefore still be read as evidence that a treatment works, while only non-adoption receives an explicit non-causal disclaimer.
  - location: docs/transport-measurements.md:189,206,273; scripts/check-decision-rule.py:31,127-130
- **The memo declares a readiness transition required before the experiment establishes that requirement.** (medium)
  - concrete_change: Make the requirement conditional: if the request-path delay reproduces, evaluate an eligibility treatment or validation before checkout; do not assert that a readiness transition is already established.
  - details: Existing measurements show an association, pre-opened sessions still had multi-second outliers, and the decision procedure can adopt neither when the delay does not reproduce. Calling a transition required is therefore stronger than the evidence and conflicts with the instruction not to implement either candidate before running the experiment.
  - location: docs/transport-measurements.md:56-68,119-121
- **The claim that C's added herdr load is small has no measured denominator.** (medium)
  - concrete_change: Report only the exact per-insertion cost and state that its per-checkout magnitude depends on measured pool reuse or session lifetime.
  - details: Correcting the cost from per-checkout to per-insertion fixes the accounting, but magnitude still depends on the unmeasured checkout-to-insertion ratio. With little session reuse, one sacrificial connection per insertion may not be small.
  - location: docs/transport-measurements.md:322-327

**Tests Run**
- python3 scripts/check-decision-rule.py — passed: 23,328 combinations, 3,072 historical ambiguities, 0 current ambiguities
- In-memory mutation TAIL_TOLERANCE=0 — checker exited non-zero and caught both affected boundary cases
- export PATH=/opt/swift/usr/bin:$PATH; swift build — passed; only sandbox cache warnings
- export PATH=/opt/swift/usr/bin:$PATH; swift test — 47 passed, 0 failures, 0 skips; live tests executed
- git diff --check d12c302...HEAD — passed
- python3 -m py_compile scripts/check-decision-rule.py — passed

**Needs**
- Historical latency figures were not independently reproduced because this PR does not retain the original measurement harness or raw samples.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e23936489d5c`

**Summary:** The decision section and checker outputs now use symmetric, observational p95 language, and the prior load claim is accurately retracted. Two surrounding sections still contradict those corrections.

**Findings**
- **The earlier design section still declares the readiness conclusion unconditionally.** (medium)
  - concrete_change: Apply the conditional/observational wording here too: if the delay reproduces, B and C are candidates; the experiment records their p95 outcomes and may select B, C, or neither. Replace “whether waiting is enough” with the measured condition, such as “request p95 after 100 ms idle.”
  - details: It says pooling cannot fix the issue, something must make every session ready, and the experiment selects between the two treatments. The arm table likewise says B determines whether waiting is “enough.” This contradicts lines 119-125, where the need is correctly conditional, and the decision procedure, which may adopt neither and establishes no treatment effect.
  - location: docs/transport-measurements.md:56-68,147-151
- **The proposed checkout-to-insertion output cannot come from the experiment as specified.** (medium)
  - concrete_change: State that the ratio remains unknown and must be measured in a pool prototype or representative workload. If this experiment retains connection counts, describe them only as exact treatment accounting, not as an observed reuse ratio.
  - details: Every arm starts 100 fresh sessions and measures one real request per session, while the document explicitly requires no pool implementation. That run has no representative session reuse, eviction, pool sizing, or UI demand; its insertion/checkout ratio is either undefined or mechanically 1:1. It therefore cannot measure the denominator the paragraph correctly says determines C's per-checkout load.
  - location: docs/transport-measurements.md:341-353

**Tests Run**
- python3 scripts/check-decision-rule.py — passed: 23,328 combinations, 3,072 historical ambiguities, 0 current ambiguities
- export PATH=/opt/swift/usr/bin:$PATH; swift build — passed; only sandbox cache warnings
- export PATH=/opt/swift/usr/bin:$PATH; swift test — 47 passed, 0 failures, 0 skips; live tests executed
- git diff --check 0dc67e5...HEAD — passed
- Cross-file causal-language sweep over the memo, checker, and mutation harness

**Needs**
- Propagate the conditional readiness wording to the earlier design section.
- Move checkout-to-insertion reuse measurement to an experiment that actually exercises a pool and representative demand.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e294885719da`

**Summary:** The prior ratio and unconditional-design findings are fixed, and the executable decision rule is single-valued. Three remaining statements still outrun the evidence or method.

**Findings**
- **The document still reifies an unobserved readiness state** (medium)
  - change: Use observational language throughout the table, section title, and unknowns—for example, immediate-versus-aged channel-open latency—and ask what explains that difference without asserting ready/not-ready states.
  - detail: The table calls the measured latency a readiness delay, labels aged sessions ready, and line 104 asks why a fresh session is not ready. The batches establish an immediate-versus-aged latency association, not a separate readiness state or mechanism; this contradicts the association-only qualification in lines 34–49.
  - file: docs/transport-measurements.md
  - line: 19
- **The causal disclaimer overcorrects from the selector to the experiment** (medium)
  - change: Scope the disclaimer to the operational selector: it neither estimates a treatment effect nor quantifies uncertainty. Preserve the possibility of a separate randomized contrast analysis using the retained samples.
  - detail: The preregistered cutoff rule is not an effect estimator and adoption alone does not establish effectiveness. However, the experiment randomly assigns concrete interventions across fresh sessions, so its retained arm data can supply causal evidence if analyzed with uncertainty. Saying adoption is not evidence and that no analysis in the document could establish effectiveness is too absolute.
  - file: docs/transport-measurements.md
  - line: 188
- **Randomization does not guarantee equal exposure to drift** (low)
  - change: State the in-expectation limitation and retain enough ordering or timestamp data to inspect temporal imbalance.
  - detail: Randomized interleaving reduces systematic ordering confounds and balances time-varying load in expectation; it does not ensure machine load and thermal drift hit all three arms alike in one 300-session run.
  - file: docs/transport-measurements.md
  - line: 157

**Tests Run**
- python3 scripts/check-decision-rule.py passed: 23,328 combinations, 3,072 historical ambiguities, 0 current ambiguities
- swift build passed
- swift test passed: 47 tests, 0 failures, 0 skips
- git diff --check origin/main...HEAD passed

**Needs**
- Replace the remaining readiness-state terminology with the latency association actually measured.
- Limit the no-causal-inference statement to the cutoff selector rather than the randomized experiment and its data.
- Qualify the randomization claim as balance in expectation.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e430e7135792`

**Summary:** The committed CSV is structurally complete and independently reproduces the reported p95s, outlier counts, C readiness cost, and adoption of B. The post-run error policy is clearly disclosed. Four integrity and documentation gaps remain.

**Findings**
- **The analyzer does not enforce the preregistered dataset size** (medium)
  - change: Put samples-per-arm and the 500 ms outlier threshold in the shared executable contract. Before computing or printing a decision, require exactly 300 unique sequences, exactly 100 rows for each of A/B/C, no unknown arms, and every required numeric field.
  - detail: The rule is defined for 100 trials per arm, but the analyzer accepts any positive count. Removing one B row produced n=99 and the analyzer still exited zero and adopted B. A live three-trial smoke likewise printed a decision before its descriptive pass later crashed. OUTLIER_MS is also duplicated locally rather than imported, so importing flat() does not make every decision input drift-proof.
  - file: scripts/analyze-readiness.py
  - line: 34
- **A response line is counted as success without validating the ping result** (medium)
  - change: After capturing the end timestamp, validate the complete line as a successful response with id readiness and record protocol-error or malformed-response rows as errors. Either rerun the experiment or explicitly retain this limitation if successful request execution is not load-bearing.
  - detail: The harness stops at any newline and records t_request without decoding the response, checking the request id, or rejecting a server error envelope. Therefore zero error rows proves only that libssh2 delivered lines; the committed CSV cannot show that the 300 measured requests received successful ping results.
  - file: scripts/readiness-experiment.swift
  - line: 192
- **The Results provenance claim and distinct-range description exceed the analyzer and raw data** (medium)
  - change: Have the analyzer emit every cited descriptive statistic, including cluster membership and exceptions, or weaken the provenance claim. Describe the main clusters with explicit counts and exceptions rather than calling their ranges distinct.
  - detail: The analyzer does not print every number in the section: it omits the 51.5-second duration, filtered ranges, and outlier sequence. More importantly, the raw CSV contradicts visibly distinct ranges: A's >90 ms range is 140.249–177.864 ms, C includes 140.787 ms and therefore overlaps A, and B includes 1375.922 ms. Only 31/33 B samples and 15/16 C samples above 90 ms fall within 100.4–101.4 ms.
  - file: docs/transport-measurements.md
  - line: 403
- **The connection-design section still describes an unrun experiment** (low)
  - change: Rewrite this paragraph in past tense: state that the request-path run reproduced the operational margin and selected B, while retaining the non-causal and single-run limitations.
  - detail: This section says the experiment can adopt neither, asks which transition it will decide, and says not to implement before running it, although the Results section records that it ran and selected B.
  - file: docs/transport-measurements.md
  - line: 136

**Tests Run**
- Independent CSV audit: 300 unique sequences, 100 rows per arm, no missing fields or error rows
- Independent sorted-value recomputation matched A/B/C p95 165.366/100.910/100.520 ms, outliers 0/1/0, and C t_ready p95 105.163 ms
- python3 scripts/analyze-readiness.py passed and selected B
- python3 scripts/check-decision-rule.py passed: 23,328 combinations and 0 current ambiguities
- Truncated-CSV probe removed one B row; analyzer still exited zero and selected B at n=99
- Readiness harness compiled with a writable module-cache override and completed a three-trial live smoke; its analyzer printed B before crashing on the undersized half-split
- swift build passed
- swift test passed: 47 tests, 0 failures, 0 skips
- git diff --check origin/main...HEAD passed

**Needs**
- Make the analyzer fail closed on the complete preregistered dataset contract and centralize all decision inputs.
- Validate successful protocol responses in the measurement harness.
- Correct the Results provenance/range claims and stale pre-run design paragraph.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e4f341f2c5f6`

**Summary:** Re-reviewed exact head c6d55d1. Run-2 statistics independently match the CSV and the harness now validates responses, but the selector treats the two readiness costs asymmetrically, so its adoption of B is not yet justified by its own cost rationale.

**Findings**
- HIGH — docs/transport-measurements.md:195-198,229-250: t_ready is defined for both B and C as pool-slot-blocking replenishment cost, but the 50 ms cap applies only to C. Run 2's B t_ready p95 is 100.084 ms while C's is 125.25 ms; applying the stated affordability criterion symmetrically makes neither qualify, rather than adopting B. Split wall-clock eligibility delay from active herdr/server work and justify separate criteria, or apply the replenishment cap to both arms and transparently rerun/reclassify the decision.
- MEDIUM — docs/transport-measurements.md:136-142,402-405,458-460: run 1 cannot support the claim that the rule selected B in two validated runs. Its rows may contain error or mismatched envelopes, and the new error rule would produce no decision if any did; those responses were not retained, so this is unknowable. State that run 2 adopted B and that the timing-only analysis of the unvalidated run-1 lines also outputs B, without presenting run 1 as an independent confirmed decision.
- MEDIUM — scripts/analyze-readiness.py:34-67: the new shape guard checks counts and non-empty strings but not numeric validity or finiteness. Replacing one t_request value with nan preserved 300 rows, emitted max=nan, and still exited 0 with decision B. Parse and validate the complete schema up front, require finite non-negative timings and the expected sequence set, then add malformed numeric regression probes.
- LOW — docs/transport-measurements.md:12-16: the opening batch note still attributes run 1's A range of approximately 140.2–177.9 ms to the run presented below. Run 2 spans 143.5–1855.4 ms and has a 173.87 ms p95. Label the old range as run 1 or update the note to the validated run-2 figures.

**Tests Run**
- Independent nearest-rank recomputation matched both CSVs and all published run-2 medians, p95s, maxima, outlier counts, t_ready statistics, margins, and decision B
- python3 scripts/analyze-readiness.py scripts/readiness-experiment-results.csv passed
- python3 scripts/check-decision-rule.py passed: 23,328 combinations, current rule unambiguous
- One-row truncation probe refused with NO DECISION and exit 1
- Three-row smoke probe refused with NO DECISION and exit 1
- NaN timing probe incorrectly emitted decision B and exited 0
- Injected error-row probe refused with NO DECISION and exit 1
- Harness compiled with writable module caches and completed a live three-trial validated smoke; its output was correctly refused as undersized
- swift build passed
- swift test passed: 47 tests, 0 failures, 0 skips
- git diff --check origin/main...HEAD passed

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e5728dd21d74`

**Summary:** The committed run still selects B and the independently recomputed symmetric interpretation selects neither. Run 1 is now honestly demoted and the current dataset is numerically sound, but the new rationale overstates what was established, and two analyzer validation paths do not implement their advertised guarantees.

**Findings**
- **The post-run basis for treating B and C asymmetrically remains an unsupported design conclusion.** (high)
  - concrete_change: Retain passive-wait versus active-work as a declared operational preference, but describe deterministic timing and absorption by pool sizing as hypotheses for the pool prototype. Require that prototype to measure slot occupancy, cold-start latency, and burst behavior before treating the age gate as absorbed.
  - evidence: B is passive waiting and C performs active server work, but B's delay is not deterministic: usleep establishes a minimum delay and can overshoot under scheduling, suspension, or load. Nor does this experiment show that pool sizing absorbs that occupancy; cold start, bursts, pool size, eviction, and request demand were not measured. Those claims remain load-bearing in the recommendation despite the document correctly admitting that B's adoption is not robust to a symmetric cap.
  - location: docs/transport-measurements.md:443
- **The printed symmetric-cap sensitivity is not an actual recomputation of the symmetric rule.** (medium)
  - concrete_change: Implement the symmetric selector alongside flat() in the canonical decision-rule module, including all qualification conditions and ordering, then have the analyzer print the two actual outcomes. Add enumerated or boundary cases where B or C is independently disqualified before the readiness cap.
  - evidence: It compares only B and C t_ready against the cap. On a valid 3x100 synthetic dataset where B already fails the request-margin condition and C qualifies, both the committed and symmetric rules select C, but the analyzer prints that the symmetric outcome differs. The current raw dataset happens to produce the correct 'neither' result, but the promised sensitivity beside every decision is not generally valid.
  - location: scripts/analyze-readiness.py:145
- **Dataset validation still accepts a missing A-arm t_ready value.** (medium)
  - concrete_change: Require an explicit finite, non-negative t_ready_ms for every arm, including zero for A, and reject missing values rather than granting A a special exemption.
  - evidence: Replacing an A row's explicit zero t_ready_ms with an empty field still exits successfully and emits decision B. This contradicts the retained-data schema and the claim that every timing must parse as finite and non-negative.
  - location: scripts/analyze-readiness.py:61

**Tests Run**
- Verified exact head 669e7a608d811805ae39138495ed282919db382f and a clean worktree.
- Independently recomputed the committed result from run 2: B; symmetric-cap interpretation: neither.
- Ran scripts/analyze-readiness.py against the committed CSV; reported statistics and committed decision matched the raw data.
- Verified the exact NaN probe is rejected with exit 1.
- Verified a missing A-arm t_ready_ms is incorrectly accepted with exit 0.
- Ran an exact interleaved 3x100 synthetic dataset that exposed incorrect symmetric-sensitivity reporting.
- Ran scripts/check-decision-rule.py: all assertions passed across 23,328 combinations.
- Ran swift build successfully.
- Ran swift test: 47 tests passed, 0 failures, 0 skips.
- Ran git diff --check against both c6d55d1...HEAD and origin/main...HEAD successfully.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e5d2e7c9db19`

**Summary:** The executable fixes are correct: the canonical analyzer selects B, the full symmetric selector selects neither, the prior false-divergence probe now reports C under both rules, and missing A t_ready is rejected. The rewritten asymmetry bullet is honest, but its two retracted conclusions still survive in the closing design paragraph.

**Findings**
- **The design conclusion still calls the age gate's slot-occupancy cost deterministic.** (high)
  - concrete_change: Replace the deterministic-cost claim with the observed run-2 distribution and state that whether the wait is predictable and absorbed is assigned to the pool prototype's occupancy, cold-start, and burst measurements.
  - evidence: Line 498 says the pool design carries a 'deterministic ~100 ms slot-occupancy cost,' directly contradicting lines 450-458, which correctly establish that usleep supplies only a minimum and that scheduling, suspension, load, occupancy, cold start, bursts, eviction, and demand remain unmeasured. This is the prior high finding surviving outside the rewritten bullet.
  - location: docs/transport-measurements.md:495
- **Run 1 is promoted back into a decision after being correctly declared undecidable.** (medium)
  - concrete_change: Scope adoption and sensitivity conclusions to the validated run. Describe run 1 only as an unvalidated timing-only projection that outputs B, not as a run in which B or C was adopted or rejected.
  - evidence: Lines 406-409 say run 1's responses were not validated, so its decision is unknowable and only a timing-only projection outputs B. Lines 462-464 then claim no reading across both runs prefers C, and lines 499-500 say C was not adopted under any reading of either run. Under the current error rule, run 1 has no adoption outcome to compare.
  - location: docs/transport-measurements.md:460

**Tests Run**
- Verified exact head 13e8eba2a078a93492eb2999714d8c1b79417064 with a clean worktree.
- Ran git diff --check successfully.
- Ran scripts/check-decision-rule.py: 23,328 combinations, current flat rule unambiguous, all canonical and symmetric boundary assertions passed.
- Ran scripts/analyze-readiness.py on the committed 300-row dataset: preregistered selector returned B; full symmetric selector returned neither; all reported statistics matched the document.
- Ran the prior synthetic boundary directly: when B fails the request margin and C qualifies, both flat and flat_symmetric returned C.
- Emptied an A-row t_ready through a process-substitution probe; analyzer refused with exit 1 and emitted no decision.
- Ran swift build successfully.
- Ran swift test: 47 tests passed, 0 failures, 0 skips; all live tests executed.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7e6396ee5fec1`

**Decision:** `approved`
**Summary:** Approved at 0575a9a. The closing paragraph now reports the observed run-2 distribution without treating it as constant, assigns predictability and absorption to prototype measurements, and confines every adoption/sensitivity conclusion to validated run 2. Run 1 is consistently labelled as timing-only and non-decisional. No blocking claims or executable-rule defects remain.

**Tests Run**
- Verified exact head 0575a9af2ab150c07437690ec26e5957b569ce2a with a clean worktree.
- Ran git diff --check origin/main...HEAD successfully.
- Independently recomputed B t_ready from the CSV: median 100.065 ms and nearest-rank p95 100.084 ms, correctly rendered as 100.06 and 100.08.
- Swept docs and scripts for run-1 decisional language and deterministic/constant/absorbed cost claims; remaining occurrences are explicitly retracted, negated, descriptive, or assigned to future prototype verification.
- Ran scripts/check-decision-rule.py: 23,328 combinations, current flat rule unambiguous, all boundary assertions passed.
- Ran scripts/analyze-readiness.py on the committed 300-row dataset: preregistered outcome B and symmetric outcome neither, matching the document.
- Ran swift build successfully.
- Ran swift test: 47 tests passed, 0 failures, 0 skips; all live tests executed.

PR #8, opened 2026-08-02
docs: connection pool design

jerryfane:
Goal v1b, task 3 — **design memo, not an implementation.** Task 3's code is gated on task 1 receiving a clean verdict; this records the shape now so that when it ungates the design is settled by measurement rather than argued a fifth time.

Coordinator authorised this to run in parallel with #4's fix rounds.

## The design

**Pool authenticated SSH sessions. Open a fresh channel per request. Do not pool channels.**

| object | cost | pool it? |
|---|---|---|
| session | ~39 ms crypto **+** 100–1900 ms first-channel | **yes** |
| channel | 0.1 ms after the first | **no** — nothing to amortise, and herdr's command socket is single-shot |

That asymmetry is the entire design, and it was neither obvious nor guessable — it came out of measurement (#7).

## The four wrong turns, recorded

More instructive than the answer:

1. **"Pool channels, evict at ~3 s"** — channel-open is free after the first.
2. **"The residual is herdr's two-write pattern"** — an echo socket with no herdr code paid *more*. `herdr#29` stays parked.
3. **"Session setup is a timer to remove"** — ~39 ms of it is real cryptography.
4. **"The first channel costs 1.8 s"** — the magnitude varies 100–1900 ms; only the *shape* is stable.

## Eviction — the inherited number does not transfer

The original ~3 s threshold came from herdr's `INITIAL_REQUEST_TIMEOUT`, which governs **herdr's socket**, not sshd's session. Reusing it would inherit a justification that doesn't apply — the same error as reusing a percentile whose distribution doesn't apply.

Default to **validate-on-borrow rather than evict-on-age** until there's a number with its own basis. A phone suspends in the background, so a pooled session must be assumed dead on foreground rather than probed optimistically.

## Stated condition under which this memo is wrong

The first-channel cost's **cause is unknown**. The design doesn't depend on it — the cost is per-session either way — which is why the investigation stopped.

But if it is ever explained and turns out to be per-**connection** rather than per-**session**, **this design is wrong and needs a fifth revision.** Recorded so the condition is visible rather than discovered.

## Verification, when implemented

The axis is **not** "a pool returns a session." It is **"the second request to a host does not pay the first-channel cost."** A test that borrows twice and compares the two costs is the guard; one that checks a session comes back proves nothing.

## RISK

- Design only — no code, no tests, nothing verified by execution.
- Rests entirely on #7's measurements. If those are wrong, this is wrong.
- Pool sizing is deliberately left open: it should be driven by UI concurrency demand, which is a product question, not by the latency figures.

HERDR-APP:

### DIFF
--- a/docs/connection-pool-design.md
+++ b/docs/connection-pool-design.md
@@ -24,13 +24,32 @@ channels.**
   **Consequence: pooling alone is insufficient.** A freshly authenticated session
   handed straight out still pays.
 
-  **Readiness, defined executably** — "prewarmed" is not a design until it names
-  a transition. A session becomes eligible for checkout when a **sacrificial
-  `direct-streamlocal` channel has been opened and freed on it**. That is the
-  operation whose cost is being avoided, so performing it once is what proves
-  readiness; an age gate would be a proxy for it and would need its own
-  measurement to justify a number. Prewarm on insertion, not on borrow, so the
-  cost lands off the request path.
+  **Readiness is not settled, and this memo previously pretended it was.** It
+  named the sacrificial `direct-streamlocal` open as the transition and
+  dismissed an age gate as "a proxy that would need its own measurement" — while
+  the two lines above report the measurement that an age gate *does* have and a
+  sacrificial channel does not. Nothing measured so far opens one.
+
+  Two candidate transitions, to be decided by the three-arm experiment in
+  `transport-measurements.md` before implementation, not after:
+
+  - **Age gate** — a session is eligible once it is 100 ms old. Has supporting
+    data (0.28–0.54 ms against ~96 ms) and costs nothing at the server.
+  - **Sacrificial channel** — a session is eligible once one channel has been
+    opened and freed on it. Exercises the exact operation being avoided, but
+    **doubles the connections herdr sees per checkout**, and herdr reaps
+    connections that open and go idle (`INITIAL_REQUEST_TIMEOUT`), so it
+    manufactures the condition the pool exists to avoid.
+
+  The rule fixed in advance: **the sacrificial channel is adopted only if it
+  beats the age gate by a margin worth its server-side cost.** If they tie, the
+  age gate wins on simplicity. Whichever is chosen, prewarm **on insertion**, not
+  on borrow, so the cost lands off the request path — that part holds either way,
+  because it follows from where the cost is paid rather than from what removes it.
+
+  **Failure handling for the transition:** a session whose prewarm fails is
+  discarded, never inserted, and never counted toward pool capacity. A prewarm
+  that fails is evidence about the session, not a step to retry on it.
 - Channels cost 0.1 ms **on a ready session** — there is nothing to amortise — and herdr's command socket is **single-shot**, so every request
   needs its own channel regardless.
 
@@ -102,7 +121,21 @@ that already landed.
 So:
 
 - **A separate pre-send liveness probe**, not the user's request doubling as the
-  check.
+  check. Specified, because "a probe" is not implementable:
+  - **Operation:** open a `direct-streamlocal` channel to herdr's socket and
+    send `{"id":"<probe>","method":"ping","params":{}}`.
+  - **Success:** a decodable response envelope carrying the same `id`. Not
+    "bytes arrived", and not "the channel opened" — herdr's socket accepts a
+    connection before it commits to answering, so a channel that opens proves
+    the tunnel and nothing about the server.
+  - **Timeout:** 1 s. Above herdr's observed round trip by a wide margin and far
+    below any interval a user would wait through. Stated as a chosen bound, not
+    a measured one.
+  - **On failure or timeout:** discard the session, close it, replace it. Never
+    probe twice, and never send the user's request down a session that failed —
+    a second probe is the same optimism the delivery rule below refuses.
+  - **`ping` specifically** because it changes nothing. A probe with a side
+    effect would make every checkout an unlogged write.
 - **Discard and replace** a session that fails the probe; never nurse it.
 - **Delivery phases are defined by observed write progress, not by error
   names.** `libssh2_channel_write_ex` returns a positive accepted-byte count or
@@ -111,6 +144,23 @@ So:
     accepted for this request. Safe to retry automatically.
   - **possibly-sent** — any positive byte count was accepted, even partially.
     Not safe to retry, regardless of what error followed.
+
+  Two return values are easy to misread as progress, and both have bitten this
+  codebase's read path already:
+
+  - **`EAGAIN` is not a failure and not progress.** It means resume the *same*
+    attempt with the identical buffer and length. libssh2 keeps a partially
+    transmitted packet in its pending-send state and only returns a positive
+    payload count once the whole channel-data packet has cleared, so a request
+    can sit in EAGAIN with bytes already on the wire. Restarting the write from
+    a different offset corrupts the stream.
+  - **A zero return is not completion.** It means the channel window had no
+    capacity, so nothing was accepted — no progress, and the request stays
+    definitely-not-sent unless a previous call already returned positive.
+
+  Cumulative progress is therefore tracked **per request across every write**,
+  not per call: a request spanning several writes becomes possibly-sent at the
+  first positive return and stays that way.
 
   Tracking accepted bytes per request is what makes the boundary decidable; an
   error code alone cannot supply it, because the same code can follow either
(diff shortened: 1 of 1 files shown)


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7d81a2129a5f2`

**Summary:** Pooling authenticated sessions remains the right direction, and channels should remain per-request. However, the memo's stated causal premise was disproved by the #7 controls, and it omits load-bearing lease, validation, retry, and verification semantics needed before implementation.

**Findings**
- **The design still treats 100-1900 ms as an intrinsic once-per-session first-channel cost. The #7 audit found that 30 immediate first opens had p50 about 96 ms, while 30 sessions aged 100 ms before opening any channel had first-open p50 about 0.40 ms. The cost follows post-authentication readiness, not necessarily the first channel. The recorded fourth correction and per-session/per-connection falsifier therefore remain wrong.** (HIGH)
  - concrete_change: Retain session pooling, but describe pooled objects as authenticated and prewarmed or readiness-validated sessions. Replace the stated falsifier with the already-cheap control: compare immediate and aged first-channel opens before any channel has warmed the session.
  - location: docs/connection-pool-design.md:13
- **The memo sizes the pool against concurrent demand without specifying whether a session is leased exclusively. Multiple concurrent libssh2 calls using one session require synchronization; official guidance says only one thread at a time may use a session and the lock must span an operation through success or error ([libssh2 guidance](https://libssh2.org/mail/libssh2-devel-archive-2019-02/0010.shtml)). A long-lived subscription makes this distinction especially important.** (HIGH)
  - concrete_change: Define an exclusive session lease: one in-flight request per pooled session, return it only after the channel is freed, and use a dedicated session for each persistent stream. If multiplexing is intended, specify complete per-session serialization instead.
  - location: docs/connection-pool-design.md:66
- **Validate-on-borrow is not inherently safe. If validation is the user's request, failure after a possible write has ambiguous delivery and automatic retry can duplicate an action. If validation is separate, the first visible request still pays validation and replacement latency, and the connection can die immediately afterward.** (HIGH)
  - concrete_change: Specify a separate pre-send liveness probe, discard-and-replace behavior, and error phases that distinguish definitely-not-sent from possibly-sent. Never replay a possibly-sent mutation without an idempotency or attempt identifier. Proactively invalidate all sessions on foreground instead of waiting for the first borrow.
  - location: docs/connection-pool-design.md:57
- **Comparing latency of two borrows is neither deterministic nor specific to pooling. Variance can swamp the comparison, and two newly authenticated but aged sessions can both open cheap first channels, allowing a no-pool implementation to pass.** (MEDIUM)
  - concrete_change: Test structurally with an instrumented session factory: two sequential requests must authenticate one session and open two channels; concurrent requests must use distinct exclusive leases; a dead session must be discarded and replaced. Keep latency comparison as a benchmark, not the correctness gate.
  - location: docs/connection-pool-design.md:84
- **Session lifetime remains insufficiently evidenced. The prior control only established cheap reuse through 10 seconds on loopback. ClientAliveInterval and ClientAliveCountMax do not impose a maximum idle age on a responsive session; they terminate an unresponsive client after missed probes ([OpenSSH documentation](https://man.openbsd.org/sshd_config)).** (MEDIUM)
  - concrete_change: Measure reuse across realistic foreground idle intervals and an actual iOS suspend/resume or network transition. Treat foreground as a pool-generation invalidation until those measurements justify another policy.
  - location: docs/connection-pool-design.md:49
- **The third recorded correction still labels the measured handshake and authentication phases as real cryptographic work. API-level phase timing includes network I/O, server scheduling, and cryptography; the measurements did not isolate cryptographic CPU work.** (MEDIUM)
  - concrete_change: Call these handshake-phase and authentication-phase latency. The pooling argument only needs them to be setup latency, not proof that all 39 ms is irreducible cryptography.
  - location: docs/connection-pool-design.md:16

**Tests Run**
- Verified exact head ec2d5af56537fa05ff1bb84051cbaa3d81804342 and confirmed the PR adds only docs/connection-pool-design.md.
- Audited the memo against PR #7's recorded evidence and the controlled age, target-order, and idle-session results from its review.
- swift build passed.
- swift test passed: 47 tests, 0 failures; live socket tests executed.
- git diff --check passed and the worktree remained clean.

**Needs**
- Rewrite the memo around authenticated-session readiness rather than an intrinsic first-channel penalty.
- Define exclusive leasing, stream ownership, liveness validation, and ambiguous-delivery retry rules before implementation.
- Replace the timing-only verification axis with deterministic structural tests and separate performance benchmarks.
- Actual iOS suspension, cellular/NAT lifetime, and implementation behavior could not be verified from this document-only PR; PR #4 is not present on this branch.

jerryfane:
> Agent: `herdr-ios-review-sol-temp-local-review-herdr-ios-review-sol-18c7d9907c73afac`
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7d9907c73afac`

**Summary:** Exclusive leasing is now clear, and pane subscriptions can remain one dedicated stream session because the existing API batches many pane entries into one events.subscribe call. Two implementation-critical contracts remain underspecified, and several sentences still encode the retracted intrinsic-first-channel premise.

**Findings**
- **Define an executable readiness transition and test the first checkout** (P1)
  - body: “Prewarmed or readiness-validated” does not say what makes a new session eligible: a measured age gate, a sacrificial direct-streamlocal channel open/free, or some other probe. The aged-session falsifier validates the premise but does not define production behavior. Specify one transition, its failure handling, and a guard proving the first user request from a newly populated pool avoids the immediate-open delay. Lines 134–137 currently test only the second request and would pass an implementation that exposes one cold request.
  - file: docs/connection-pool-design.md
  - line: 24
- **Define delivery phases from observed write progress, not error names** (P1)
  - body: The boundary is determinable conservatively, but the memo only demands it without defining it. libssh2_channel_write_ex reports a positive accepted-byte count or a negative status; EAGAIN must resume with the same buffer and can occur while an SSH packet is partially transmitted. Therefore error codes alone do not establish delivery. Specify that failures before any user-payload write are definitely-not-sent; after any positive payload-write return, later write/read/close failures are possibly-sent; EAGAIN remains inside the same attempt; and only a valid response confirms completion. This must track cumulative progress for requests spanning multiple writes. See the [official API contract](https://libssh2.org/libssh2_channel_write_ex.html) and [libssh2 1.11.0 transport behavior](https://github.com/libssh2/libssh2/blob/libssh2-1.11.0/src/transport.c#L758-L823).
  - file: docs/connection-pool-design.md
  - line: 100
- **Remove remaining intrinsic-first-channel claims** (P2)
  - body: Lines 27, 37, 44–46, and 109 still say channels become cheap “after the first” or call “first expensive, rest free” the stable shape, contradicting lines 19–22 where an aged session’s first channel is already cheap. Lines 134–137 inherit the same premise. Replace these with readiness-based wording: channel opens are cheap after the session’s readiness delay or readiness transition, regardless of ordinal position.
  - file: docs/connection-pool-design.md
  - line: 27

**Tests Run**
- Verified HEAD is exactly 3d3b1acf5aee55a55e6e4c073c9d0f52ceb15762 and the worktree is clean.
- git diff --check 497bda5..3d3b1ac passed.
- Compared the complete memo against main and the prior ec2d5af revision; PR changes only docs/connection-pool-design.md.
- Audited the installed libssh2 1.11.0 headers and upstream 1.11.0 channel, transport, and keepalive implementations. No Swift tests were run because this PR changes documentation only.

jerryfane:
> Agent: `herdr-ios-review-sol-temp-local-review-herdr-ios-review-sol-18c7daf6bc1c702b`
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7daf6bc1c702b`

**Summary:** The propagation cleanup is complete, and the accepted-byte delivery boundary is conservative under libssh2's partial-write behavior. However, readiness remains insufficiently justified and its verification would miss a missing prewarm.

**Findings**
- [P1] docs/connection-pool-design.md:27-33 selects a sacrificial direct-streamlocal channel without measuring it against the age gate. Lines 19-21 already report that 100 ms of ageing makes the first channel cheap, contradicting the claim that an age gate lacks measurement. A sacrificial direct-streamlocal open also creates a real connection to herdr's Unix socket. Compare readiness latency, variance, and server-side effects before declaring this the settled transition.
- [P1] docs/connection-pool-design.md:148-153 retains the old verification axis. If insertion-time prewarming is accidentally removed, the first request is slow and the second cheap, so the proposed two-borrow comparison still passes. The guard must assert that the first real request after a session becomes eligible avoids readiness cost, with removal of the prewarm transition mutation-verified.
- [P1] docs/connection-pool-design.md:71-74 and 104-106 still leave validate-on-borrow non-executable: 'a separate pre-send liveness probe' names neither the operation nor the acknowledgement that proves the session live. Define the probe, success condition, timeout, and discard behavior.
- [P2] The delivery boundary itself holds: libssh2 retains partial raw-socket sends and returns EAGAIN until the complete SSH channel-data packet clears pending-send state; only then does channel_write return a positive payload count. However, the memo should also specify that EAGAIN retries the identical buffer and length, and that a zero return is possible when the channel window has no capacity and means no progress, not completion. See [libssh2 channel source](https://github.com/libssh2/libssh2/blob/libssh2-1.11.0/src/channel.c#L2327-L2488) and [transport source](https://github.com/libssh2/libssh2/blob/libssh2-1.11.0/src/transport.c#L759-L921).

**Tests Run**
- git diff --check 497bda5..7330e95 — passed
- Repository-wide stale-premise grep — passed; no remaining 'after the first', 'first expensive', 'rest free', or 'free after the first' wording
- Inspected installed libssh2 1.11.0 documentation and tagged channel/transport source for partial-write, EAGAIN, zero-return, and pending-send semantics
- No runtime suite run; this is a documentation-only change

jerryfane:
> Agent: `herdr-ios-review-sol-temp-local-review-herdr-ios-review-sol-18c7dd7d53447a04`
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7dd7d53447a04`

**Summary:** The verification and delivery-boundary corrections are sound, and herdr does accept/enqueue a connection before reading or committing to a response. Two design decisions remain underspecified.

**Findings**
- P1 docs/connection-pool-design.md:38-46 — The readiness adoption rule is still not executable. “Beats the age gate by a margin worth its server-side cost” has no threshold, while the referenced a01f53b measurement rule uses undefined “C ≈ B.” The cost unit is also wrong: insertion-time sacrificial prewarming adds one herdr connection per newly inserted session, whereas the mandatory liveness probe adds one per checkout. Define the C-versus-B threshold and evaluate costs per insertion and per borrow before claiming the rule is fixed.
- P1 docs/connection-pool-design.md:123-138 — The probe still lacks a complete state machine. Specify newline-delimited request/response framing; whether success requires a matching-id pong or also admits a matching-id error envelope; that the one-second deadline covers channel opening, the complete write, one complete response line, and decoding; successful channel close/free before opening the user-request channel; and a finite replacement-attempt policy when replacement sessions also fail. Herdr waits for a newline and otherwise keeps the accepted connection until INITIAL_REQUEST_TIMEOUT, so these omissions materially change behavior.

**Tests Run**
- export PATH=/opt/swift/usr/bin:$PATH && swift test — 47 tests passed with 0 failures
- git diff --check 497bda5..0a3a758 — passed
- Read-only verification against local herdr d293951f: src/api/server.rs accepts and spawns a connection handler before reading the initial newline-delimited request; INITIAL_REQUEST_TIMEOUT is 5 seconds; ping returns a matching-id pong response

jerryfane:
> Agent: `herdr-ios-review-sol-temp-local-review-herdr-ios-review-sol-18c7de601e46580f`
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7de601e46580f`

**Summary:** Both prior P1s are addressed, but the checkout retry timing still needs one bounded, implementable policy before merge.

**Findings**
- **Bound the complete pre-send checkout loop, not each probe independently.** (P1)
  - detail: The 1-second probe deadline and 3-attempt cap compose to 3 seconds before the request begins, while acquiring or preparing replacement sessions is not covered and can extend that further. Use one 1-second deadline across checkout validation, replacement acquisition, and all probes, with at most 2 attempts total: the initial session plus one ready replacement. A fast stale-session failure can recover once; a slow failure consumes the shared budget and avoids retrying a likely backend-wide outage. If no ready replacement exists, fail with backend unavailable and replenish asynchronously. The real request channel should then receive its own separate operation deadline because its delivery semantics differ.
  - file: docs/connection-pool-design.md
  - line: 152

**Tests Run**
- git diff --check 0a3a758..67dd818 passed
- Manually compared connection-pool-design.md at 67dd818 with the canonical transport-measurements.md at 7a6836e
- Verified the worktree remained clean

**Needs**
- PR #7 must land before PR #8, or PR #8 must be rebased afterward, because the canonical document referenced by this memo is absent from the PR #8 tree.

jerryfane:
> Agent: `herdr-ios-review-sol-temp-local-review-herdr-ios-review-sol-18c7df2923e6c0f6`
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7df2923e6c0f6`

**Decision:** `approved`
**Summary:** Approved at e06d635. The prior P1 is resolved: one 1-second deadline now bounds the entire pre-send checkout, retry is limited to one already-ready replacement, and the request retains a separate deadline to preserve delivery-ambiguity semantics. The herdr causality claim is now appropriately qualified.

**Tests Run**
- git diff --check 497bda5...e06d635 passed
- Compared the memo against PR #7 head 953ace7's canonical transport-measurements document
- Verified the PR #7 and PR #8 heads merge without a textual conflict

**Needs**
- Merge PR #7 at 953ace7 before PR #8, or rebase PR #8 onto it, because docs/connection-pool-design.md references docs/transport-measurements.md and that file is absent from PR #8's current base.

PR #11, opened 2026-08-02
feat(kit): streamFailed + RecoveryExecutor — the policy meets the wire

jerryfane:
Task 6, per the coordinator-approved plan and its four binding conditions.

**WHAT**

- **`ClientEvent.streamFailed(pane:from:)`** — requirement one. The wire is N
  independent per-pane persistent connections, so one dying was previously
  *inexpressible*: the pane went silent while the client believed itself
  connected. Handled by `dropAndReadmit` — admission's inverse fused with a
  fresh admission in ONE authority section, so a concurrent discovery cannot
  double the replacement, and re-subscription rides the same atomic gate rather
  than getting its own machinery.
- **`RecoveryExecutor`** (an actor) — maps every provenance-bound action to
  `ExecutorTransport` calls and mints `ClientEvent`s from outcomes. This is
  where the policy's AttemptID bindings are finally **enforced**: a subscribe or
  reconnect bound to a retired attempt is refused at execution, observably.

**THE FOUR CONDITIONS**

1. *Delayed callbacks provably rejected* — a late completion is discarded **at
   the transport** (`testADelayedCompletionFromARetiredAttemptIsDiscardedAtTheTransport`,
   which also asserts no resync and no stream follow it); a stale subscribe
   opens nothing and is recorded (`testASubscribeBoundToARetiredAttemptIsRejectedAtExecution`);
   a stale reconnect neither dials nor disturbs the current attempt
   (`testAStalePlansReconnectCannotWedgeTheExecutor`).
2. *Policy unchanged* beyond the approved `streamFailed` addition. One
   **comment** in `SessionRecovery.swift` was corrected (behaviour untouched):
   it implied duplicate terminations land outside the ledger, which a probe
   refuted. Flagged rather than assumed.
3. *Exclusions honoured* — no UI, Keychain, reachability, or macOS.
4. *Resync-before-subscribe guarded* — `testReconnectAdoptionResyncsBeforeAnyStreamOpensDespiteMemory`.
   Note the cold-start version of this assertion let the mutation survive
   **vacuously** (empty memory subscribes nothing early); the armed version
   drives a reconnect with knowledge aboard, and the mutation kills against it.

**ADVERSARIAL SWEEP BEFORE FIRST REVIEW** — three lenses over the new code, 17
findings, 15 confirmed by independent refutation, all folded. Two HIGHs were the
actor-reentrancy class the policy could not have, each reproduced 4/4:

- **Wedge**: a stale plan's `.reconnect` cancelled the *current* attempt's
  pending dial; the silent cancellation return left nothing scheduled.
- **Leak**: an unconditional `resourcedAttempt` clear after a suspension point
  destroyed an interleaved plan's claim, after which every teardown closed a
  resource-less attempt while the live one's streams leaked forever.

Also folded: the exactly-once termination contract is now **enforced** by a
latch (a double-fire produced three opens for one death); transport teardown,
the backoff delay, and "no work on the discarded attempt" were all unpinned and
now carry killed mutations; `LiveHerdrTransport` violated the seam requirements
it was cited as proving implementable (attempt-blind `closeAll`, termination on
deliberate teardown).

**RESULTS** — 120 tests, 0 failures across repeated full-suite runs; executor
suite 3/3; mutations killed: reconnect-unbound, unconditional-claim-clear,
backoff-ignored, no-closeall, no-once-latch, work-on-discarded, door-only-binding.

**RISK / open**

- **A PRE-EXISTING FLAKE, NOT FROM THIS PR**, reported rather than absorbed:
  `SSHTransportTests.testHostKeyRejectionThroughStreamNeverClosesADescriptorTwice`
  fails roughly 1 run in 10 under full-suite load with `SSH handshake failed
  (-43)` instead of the expected rejection — local `sshd` refusing under 25
  rapid connections plus parallel load. It is merged task-1 code; fixing it
  inside an executor PR is the quiet scope expansion condition 3 warns about, so
  it is escalated separately. Expect it to appear in a review run.
- Two of my own new tests failed honestly during development and are recorded in
  the commit: a deadlocked interleaving regression, and a force-unwrap that
  turned a real mutation kill into signal 4 (harness reads INVALID) — the same
  crash-instead-of-fail class the sweep found elsewhere.
- `rejections` grows unbounded; it is a diagnostics surface bounded by the
  rarity of rejections, and nothing drains it. Stated, not solved.
- The executor exposes event injection points; nothing wires them to
  `NWPathMonitor` or app lifecycle — that is the macOS half, excluded.

HERDR-APP:

---

**Task 7 cross-link (per directive 28002):** the admitted-vs-open policy round is filed as #12. This PR adds the mandated observation surface (`paneStatus`) and does not redefine admission.

### DIFF
--- a/.gitmoot/REVIEW-HEADER.md
+++ b/.gitmoot/REVIEW-HEADER.md
@@ -4,14 +4,41 @@ Prepend verbatim to every review dispatch in this lane.
 
 ## 1. Establish which code you reviewed, and NAME it
 
-**Archive the PR's current head yourself and review that.** Do not trust the
-job's pinned SHA, and do not review the shared worktree
-(`/root/.gitmoot/worktrees/jerryfane--herdr-ios/review-pr-<N>-<id>`) without
-first confirming its HEAD matches the head you intend.
+**Read the commit through local plumbing rather than trusting the working
+tree.** This needs no network and no auth, so it cannot fail closed:
 
-**State in `Tests Run`: the exact head you archived, and the diff range you
-examined.** That line is load-bearing evidence, not a formality — it is
-currently the only channel that reports which code a verdict actually covers.
+```
+git cat-file -t <sha>                              # the gate: object present?
+git diff $(git merge-base <sha> origin/main)..<sha>
+git show <sha>:<path>
+```
+
+Do not review the shared worktree
+(`/root/.gitmoot/worktrees/jerryfane--herdr-ios/review-pr-<N>-<id>`) without
+first confirming its HEAD matches the head you intend. If you *can* reach the
+PR's live head, archive it yourself and review that; if you cannot, review the
+pinned SHA through the plumbing above and say which you did.
+
+**State in `Tests Run` BOTH the SHA you reviewed AND the worktree HEAD you
+found**, even when they agree — especially when they agree.
+
+That pairing is the whole mechanism, and it is why this is a requirement rather
+than advice: it turns an instruction into a **detector**. A reviewer that
+quietly skips the plumbing is caught by the mismatch in its own output, at
+verdict time, instead of by someone diffing afterwards — or never. *An
+instruction that cannot detect its own violation is a hope.* Apply that test to
+any review contract written in this lane, not just this one.
+
+**Two traps that have already fired here:**
+
+- `gitmoot job show`'s verdict summary begins with the **worktree** SHA. Never
+  parse that field for currency; it displays the stale commit by construction.
+- A `policy: read-only` reviewer has no network and no valid auth. A brief
+  demanding verification against the PR's *live* head therefore fails closed
+  forever — six blocked rounds elsewhere proved it. The plumbing path above is
+  the reason this header does not have that failure mode. Do not "fix" it by
+  granting the reviewer network: that expands privilege on the least-privileged
+  party to perform a check the merge gate already does without a race.
 
 **The job record is not evidence.** At execution the job's `head_sha` is
 overwritten from the worktree, so a stale checkout produces a record that agrees
@@ -24,8 +51,18 @@ review task id stable rather than head-derived, so the worktree keyed by it is
 reused and never re-synced. On this repo, five consecutive dispatches recorded a
 head two commits stale while the verdicts were correct. The correctness came
 from *somewhere* — reviewer habit, or an explicitly-passed `--head-sha`; the
-lane's own data cannot distinguish them. A compensating behaviour nobody
-designed as a safeguard is not a safeguard. This header makes it one.
+lane's own data could not distinguish them at the time. A compensating
+behaviour nobody designed as a safeguard is not a safeguard. This header makes
+it one.
+
+That question has since been settled, and the answer removes the more
+comfortable option: a dispatch passing `--head-sha c340f2b` against a reused
+task recorded `e02e40ba`, three rounds stale. **`--head-sha` is not a
+mitigation.** Reviewer habit — this header — is the only thing standing
+between a reused worktree and a confident verdict on the wrong commit. On #11
+that mattered: at the stale head the guard under review still existed, so a
+review of the assigned worktree would have returned clean on a head containing
+a live race.
 
 ## 2. Standing verification clauses
 
--- a/.gitmoot/REVIEW-HEADER.md
+++ b/.gitmoot/REVIEW-HEADER.md
@@ -35,6 +35,28 @@ case — a PR elsewhere merged on exactly such a void verdict. A blob id cannot
 be produced without reading the tree it belongs to, so it is the first
 statement in this header that the coordinator did not supply the answer to.
 
+### Choosing a discriminator (coordinator-side)
+
+**It must sit on the CRITICAL PATH OF THE REASONING, not merely inside the
+diff.** Content a reviewer would read but never have reason to quote —
+comment prose, an identifier cited by line number rather than by name, a hash
+nothing depends on — produces FALSE NEGATIVES: elsewhere this check missed
+three times on a review that had demonstrably read the right head. Prefer an
+exact count the reviewer must verify to reach a verdict, an identifier central
+to the change, or a label with zero occurrences at the parent commit.
+
+**Classify the result in THREE states, never two:**
+
+| | |
+|---|---|
+| ENGAGED | the verdict engages the discriminator → confirmed |
+| INCONCLUSIVE | no engagement, but other head-specific evidence is present — line-number citations, answers to structural questions, findings naming code absent at the parent. **Look closer; do not re-run and do not accuse.** |
+| NO EVIDENCE | no engagement and nothing else head-specific → treat as unverified |
+
+A two-state classifier folds UNCERTAIN into ACCUSATORY, and the remedy for a
+suspected void verdict is another round — so the error lands in the expensive
+direction *and* trains distrust of a reviewer that was correct.
+
 That pairing is the whole mechanism, and it is why this is a requirement rather
 than advice: it turns an instruction into a **detector**. A reviewer that
 quietly skips the plumbing is caught by the mismatch in its own output, at
--- a/.gitmoot/REVIEW-HEADER.md
+++ b/.gitmoot/REVIEW-HEADER.md
@@ -22,6 +22,19 @@ pinned SHA through the plumbing above and say which you did.
 **State in `Tests Run` BOTH the SHA you reviewed AND the worktree HEAD you
 found**, even when they agree — especially when they agree.
 
+**And report the CONTENT DISCRIMINATOR the dispatch asks for.** Each dispatch
+names a file and asks for `git rev-parse <sha>:<path>` — a blob id — plus the
+verbatim text of a specified line. Do not infer these from the dispatch; it
+never contains them.
+
+That requirement exists because naming the SHA is not independent evidence: the
+brief HANDS you the SHA. A reviewer that reads a stale worktree and quotes the
+pinned SHA satisfies every check above without ever touching the code under
+review, and under gitmoot#1415 that is the common case rather than an edge
+case — a PR elsewhere merged on exactly such a void verdict. A blob id cannot
+be produced without reading the tree it belongs to, so it is the first
+statement in this header that the coordinator did not supply the answer to.
+
 That pairing is the whole mechanism, and it is why this is a requirement rather
 than advice: it turns an instruction into a **detector**. A reviewer that
 quietly skips the plumbing is caught by the mismatch in its own output, at
--- a/.gitmoot/REVIEW-HEADER.md
+++ b/.gitmoot/REVIEW-HEADER.md
@@ -1,0 +1,39 @@
+# Standing review header — jerryfane/herdr-ios
+
+Prepend verbatim to every review dispatch in this lane.
+
+## 1. Establish which code you reviewed, and NAME it
+
+**Archive the PR's current head yourself and review that.** Do not trust the
+job's pinned SHA, and do not review the shared worktree
+(`/root/.gitmoot/worktrees/jerryfane--herdr-ios/review-pr-<N>-<id>`) without
+first confirming its HEAD matches the head you intend.
+
+**State in `Tests Run`: the exact head you archived, and the diff range you
+examined.** That line is load-bearing evidence, not a formality — it is
+currently the only channel that reports which code a verdict actually covers.
+
+**Why this is a requirement and not a courtesy:** C1 of gitmoot#1354 made the
+review task id stable rather than head-derived, so the worktree keyed by it is
+reused and never re-synced. On this repo, five consecutive dispatches recorded a
+head two commits stale while the verdicts were correct. The correctness came
+from *somewhere* — reviewer habit, or an explicitly-passed `--head-sha`; the
+lane's own data cannot distinguish them. A compensating behaviour nobody
+designed as a safeguard is not a safeguard. This header makes it one.
+
+## 2. Standing verification clauses
+
+- Mutants must BUILD. A mutation that fails to compile proves nothing about the
+  tests; classify it INVALID, not KILLED.
+- A crash (signal 4) is not a kill. An assertion followed by a force-unwrap
+  turns a real failure into an unclassifiable run — use `XCTUnwrap`.
+- Check the PRECONDITION arms the hazard. A test whose setup never reaches the
+  condition under test passes for reasons unrelated to its name; on this repo
+  that has been the cause more often than a weak assertion.
+- Restore per-file after probing; leave the worktree clean.
+
+## 3. What this lane wants from a review
+
+Disagreement, with a demonstration. Every round on #10 and #11 that found
+something real found it by running a probe, not by reading. A verdict that
+agrees is worth less than one that reproduces.
--- a/.gitmoot/REVIEW-HEADER.md
+++ b/.gitmoot/REVIEW-HEADER.md
@@ -13,6 +13,12 @@ first confirming its HEAD matches the head you intend.
 examined.** That line is load-bearing evidence, not a formality — it is
 currently the only channel that reports which code a verdict actually covers.
 
+**The job record is not evidence.** At execution the job's `head_sha` is
+overwritten from the worktree, so a stale checkout produces a record that agrees
+with it — self-consistent and wrong. Only two surfaces can be trusted: the
+worktree HEAD observed *before* the verdict, and the head you name in your own
+verdict body.
+
 **Why this is a requirement and not a courtesy:** C1 of gitmoot#1354 made the
 review task id stable rather than head-derived, so the worktree keyed by it is
 reused and never re-synced. On this repo, five consecutive dispatches recorded a
--- a/.gitmoot/REVIEW-HEADER.md
+++ b/.gitmoot/REVIEW-HEADER.md
@@ -40,11 +40,20 @@ any review contract written in this lane, not just this one.
   granting the reviewer network: that expands privilege on the least-privileged
   party to perform a check the merge gate already does without a race.
 
-**The job record is not evidence.** At execution the job's `head_sha` is
-overwritten from the worktree, so a stale checkout produces a record that agrees
-with it — self-consistent and wrong. Only two surfaces can be trusted: the
-worktree HEAD observed *before* the verdict, and the head you name in your own
-verdict body.
+**The job record is not evidence, and it is not a second witness.** The value is
+stored correctly when the job is queued and REPLACED FROM THE WORKTREE during
+execution (measured: `queued → 3b04cd0`, `running → e02e40ba`, same job). So a
+stale checkout produces a record that agrees with it — self-consistent and
+wrong.
+
+The trap is subtler than "the record can be stale": *record agrees with
+worktree* looks like corroboration and is not. The record is DERIVED from the
+worktree, so the two agree by construction, including when both are wrong.
+Counting that as two surfaces is reading one fact twice and calling it two
+witnesses — a derived value is never independent evidence of the thing it was
+derived from. **There are two independent surfaces:** (record ≡ worktree), and
+the head the verdict names. Compare both against the PR head; agreement between
+*those* means something.
 
 **Why this is a requirement and not a courtesy:** C1 of gitmoot#1354 made the
 review task id stable rather than head-derived, so the worktree keyed by it is
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -412,7 +412,13 @@ public actor RecoveryExecutor {
                 // Back off between consecutive failures for THIS pane on THIS
                 // attempt, through the same seam the dial uses so tests pin it.
                 await sleeper(Double(failures) * 0.25)
-                guard attempt == state.currentAttempt else { return }
+                guard attempt == state.currentAttempt else {
+                    record(rejection: (
+                        action: "openStream(\(pane))",
+                        reason: "attempt retired during retry backoff; retry abandoned"
+                    ))
+                    return
+                }
             }
             do {
                 // The exactly-once termination contract is ENFORCED here, not
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -374,18 +374,18 @@ public actor RecoveryExecutor {
 
     private func open(panes: Set<String>, for attempt: AttemptID) async {
         for pane in panes.sorted() {
-            // Re-checked PER PANE, because every openStream await is an actor
-            // suspension point: a retirement can interleave mid-loop, and a
-            // door-only check let the remaining panes open for the retired
-            // attempt AFTER its closeAll had run — orphan streams nothing would
-            // ever close. Reproduced by test before this guard existed.
-            guard attempt == state.currentAttempt else {
-                record(rejection: (
-                    action: "openStream(\(pane))",
-                    reason: "attempt retired mid-subscribe"
-                ))
-                return
-            }
+            // NO pre-await binding check here, deliberately, and this is a
+            // removal rather than an omission: one lived here and was found
+            // UNREACHABLE AS A DECIDER. Actor reentrancy happens only at
+            // suspension points, and between the previous iteration's
+            // post-await reap (which returns) and this position there is none —
+            // so any retirement a pre-check could see was already seen by the
+            // reap. Its mutation survived precisely because the reap covered
+            // every case. Two mechanisms for one property, neither individually
+            // pinnable: the trap this project has now hit four times, and the
+            // fourth was my own fix for the third.
+            //
+            // The single mechanism is the RECHECK-AND-REAP after the open below.
             let failures = openFailures[FailureKey(attempt: attempt, pane: pane)] ?? 0
             guard failures < Self.openFailureCap else {
                 record(rejection: (
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -376,9 +376,15 @@ public actor RecoveryExecutor {
         for pane in panes.sorted() {
             // NO pre-await binding check at the top of the iteration. The
             // invariant this loop keeps is stated once and held per site:
-            // EVERY SUSPENSION POINT IN THIS BODY IS FOLLOWED BY ITS OWN
-            // ATTEMPT RECHECK. There are exactly three, and each has a distinct
-            // correction because each leaves the world in a different state:
+            // EVERY SUSPENSION POINT THAT CAN REACH THE TOP OF THE NEXT
+            // ITERATION IS FOLLOWED BY ITS OWN ATTEMPT RECHECK.
+            //
+            // The qualifier is load-bearing. The body contains FOUR awaits; the
+            // fourth is `transport.closeAll` inside the reap below, which is
+            // followed unconditionally by `return`, so its resumption cannot
+            // reach the loop head and it needs no check. Three can, and each has
+            // a distinct correction because each leaves the world in a
+            // different state:
             //   1. the backoff `sleeper` below       -> plain return, nothing built yet
             //   2. `transport.openStream` succeeding  -> RECHECK-AND-REAP; a
             //      registration now exists past the teardown that would have
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -374,18 +374,26 @@ public actor RecoveryExecutor {
 
     private func open(panes: Set<String>, for attempt: AttemptID) async {
         for pane in panes.sorted() {
-            // NO pre-await binding check here, deliberately, and this is a
-            // removal rather than an omission: one lived here and was found
-            // UNREACHABLE AS A DECIDER. Actor reentrancy happens only at
-            // suspension points, and between the previous iteration's
-            // post-await reap (which returns) and this position there is none —
-            // so any retirement a pre-check could see was already seen by the
-            // reap. Its mutation survived precisely because the reap covered
-            // every case. Two mechanisms for one property, neither individually
-            // pinnable: the trap this project has now hit four times, and the
-            // fourth was my own fix for the third.
+            // NO pre-await binding check at the top of the iteration. The
+            // invariant this loop keeps is stated once and held per site:
+            // EVERY SUSPENSION POINT IN THIS BODY IS FOLLOWED BY ITS OWN
+            // ATTEMPT RECHECK. There are exactly three, and each has a distinct
+            // correction because each leaves the world in a different state:
+            //   1. the backoff `sleeper` below       -> plain return, nothing built yet
+            //   2. `transport.openStream` succeeding  -> RECHECK-AND-REAP; a
+            //      registration now exists past the teardown that would have
+            //      closed it, so returning without reaping leaks it
+            //   3. `handle(.streamFailed)` in the catch -> plain return; the
+            //      throwing open left no registration to reap
             //
-            // The single mechanism is the RECHECK-AND-REAP after the open below.
+            // A top-of-iteration guard was tried and removed as unreachable,
+            // and that removal was WRONG — round seven built a probe that
+            // walked straight through the hole. The reasoning failed by
+            // enumerating only the success path: the catch path also suspends,
+            // and it did not pass through the reap on its way back to here. The
+            // per-site rule is what makes that enumerable at all; a single
+            // top-of-loop check invites exactly the mistake I made, because it
+            // reads as covering suspensions it never sees.
             let failures = openFailures[FailureKey(attempt: attempt, pane: pane)] ?? 0
             guard failures < Self.openFailureCap else {
                 record(rejection: (
@@ -460,6 +468,25 @@ public actor RecoveryExecutor {
                     lastOpenErrors[FailureKey(attempt: attempt, pane: pane)] = String(describing: error)
                 }
                 await handle(.streamFailed(pane: pane, from: attempt))
+                // Suspension point 3. `handle` re-enters the policy, which
+                // typically dials a REPLACEMENT stream for this pane and
+                // suspends again inside it — a long, reliably-hit window with
+                // this loop parked in the middle of it. A retirement delivered
+                // there is invisible to every other check: the reap above is on
+                // the success path, and the next iteration's backoff guard only
+                // runs for a pane that has already failed once.
+                //
+                // Nothing to reap: `openStream` threw, and the seam's contract
+                // is that a throwing open leaves no live registration. Streams
+                // this attempt opened for EARLIER panes were closed by the
+                // teardown's own closeAll while this await was suspended.
+                guard attempt == state.currentAttempt else {
+                    record(rejection: (
+                        action: "openStream(\(pane))",
+                        reason: "attempt retired while handling an open failure; remaining panes abandoned"
+                    ))
+                    return
+                }
             }
         }
     }
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -191,6 +191,18 @@ public actor RecoveryExecutor {
 
     private func open(panes: Set<String>, for attempt: AttemptID) async {
         for pane in panes.sorted() {
+            // Re-checked PER PANE, because every openStream await is an actor
+            // suspension point: a retirement can interleave mid-loop, and a
+            // door-only check let the remaining panes open for the retired
+            // attempt AFTER its closeAll had run — orphan streams nothing would
+            // ever close. Reproduced by test before this guard existed.
+            guard attempt == state.currentAttempt else {
+                rejections.append((
+                    action: "openStream(\(pane))",
+                    reason: "attempt retired mid-subscribe"
+                ))
+                return
+            }
             do {
                 try await transport.openStream(pane: pane, for: attempt) { [weak self] in
                     // One termination, one death event — the exactly-once
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -432,14 +432,22 @@ public actor RecoveryExecutor {
                     await transport.closeAll(for: attempt)
                     return
                 }
-                openFailures[FailureKey(attempt: attempt, pane: pane)] = nil
-                lastOpenErrors[FailureKey(attempt: attempt, pane: pane)] = nil
                 // Published only if the stream is still alive: a termination
                 // firing while openStream was suspended already processed the
                 // death, and publishing here anyway reported a dead stream as
                 // .open — after exhaustion, permanently.
+                //
+                // The bookkeeping clears INSIDE the successful branch, not
+                // before the decision: clearing first erased exhausted retry
+                // state, so a pane that had burned every attempt reported
+                // admittedNotOpen(attempts: 0, lastError: nil) — the status
+                // surface telling the truth about openness while lying about
+                // WHY — and a retained orphan callback could then restart
+                // registration past the supposedly exhausted cap.
                 if once.publishIfUnterminated() {
                     openStreams[pane] = OpenRegistration(attempt: attempt, registration: registration)
+                    openFailures[FailureKey(attempt: attempt, pane: pane)] = nil
+                    lastOpenErrors[FailureKey(attempt: attempt, pane: pane)] = nil
                 }
             } catch {
                 // A stream that cannot OPEN is a death the same as one that
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -47,15 +47,38 @@ public protocol ExecutorTransport: Sendable {
     func discard(attempt: AttemptID) async
 }
 
-/// One-shot claim for a stream's termination callback.
+/// Lifecycle latch for one stream registration: coordinates the termination
+/// callback with the openStream return, atomically.
+///
+/// Two states short of this were two lies: a termination firing WHILE
+/// openStream was suspended still saw the success path publish the pane as
+/// open afterwards (falsely `.open` after exhaustion), and nothing tied a
+/// termination to the registration it belonged to.
 final class TerminationLatch: @unchecked Sendable {
+    private enum Lifecycle {
+        case fresh
+        case published
+        case terminated
+    }
+
     private let lock = NSLock()
-    private var claimed = false
-    /// True exactly once.
+    private var lifecycle: Lifecycle = .fresh
+
+    /// Termination side: true exactly once, whether or not the open ever
+    /// published — at-most-once is unconditional.
     func claim() -> Bool {
         lock.lock(); defer { lock.unlock() }
-        if claimed { return false }
-        claimed = true
+        if case .terminated = lifecycle { return false }
+        lifecycle = .terminated
+        return true
+    }
+
+    /// Open-return side: true iff the stream is still alive to publish. A
+    /// registration whose termination already fired must NOT surface as open.
+    func publishIfUnterminated() -> Bool {
+        lock.lock(); defer { lock.unlock() }
+        guard case .fresh = lifecycle else { return false }
+        lifecycle = .published
         return true
     }
 }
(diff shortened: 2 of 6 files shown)


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f05b91730f1f`

**Summary:** Changes requested at e02e40b. The core attempt binding is sound under the covered cases, but teardown/open interleavings and repeated stream-open failures can still leak resources or create an unbounded retry storm.

**Findings**
- HIGH — Sources/HerdrKit/RecoveryExecutor.swift:250 checks attempt validity only before awaiting openStream. A concurrent closeAll can finish while that call is suspended, after which the call can publish a retired attempt's stream. The regression at Tests/HerdrKitTests/RecoveryExecutorTests.swift:272 misses this because it asserts only that the second pane stays closed; its first pane is recorded after closeAll. The live adapter has the same register-after-close window at lines 491-516. Make registration versus teardown atomic or recheck and reap after openStream returns, and assert no retired stream appears after closeAll.
- HIGH — Sources/HerdrKit/RecoveryExecutor.swift:279 immediately converts an openStream throw into streamFailed, whose policy readmits the pane and immediately invokes openStream again. A persistently failing conformer therefore creates an unbounded, delay-free retry loop. Add attempt-bound per-pane backoff or escalate repeated open failures through transport recovery, with a deterministic always-failing test.
- MEDIUM — ExecutorTransport does not define the ownership boundary when openStream installs a callback/resource and then throws. The catch path starts a replacement with a new TerminationLatch; a later first invocation of the failed registration's old callback can then retire the replacement ledger entry and open a third stream. Require that throwing leaves no live registration and no future callback, or return an explicit registration token/handle; test an old callback firing after a failed registration.
- MEDIUM — RecoveryExecutor.rejections at line 91 is append-only and has no cap or drain operation. A long-running client receiving stale plans or duplicate callbacks grows it indefinitely. Use a bounded diagnostics buffer with a dropped count, or expose an atomic drain.
- LOW — RecoveryExecutorTests.swift uses NSLock.lock/unlock directly from async functions at lines 94, 505, and 512. The current compiler emits six warnings and -warnings-as-errors fails; these become errors in Swift 6 language mode. Replace them with scoped synchronous locking or an async-safe state holder.

**Tests Run**
- swift build — passed
- Focused mid-subscribe retirement test — passed, but does not assert the in-flight first stream is reaped
- RecoveryExecutor test selection — 13 passed
- swift test with writable module caches — 120 passed, 2 environment-dependent skips, 0 failures
- ThreadSanitizer RecoveryExecutor selection — 13 passed with no race diagnostics
- swift test -Xswiftc -warnings-as-errors focused build — failed on six async-context NSLock diagnostics

**Needs**
- Close the register-after-teardown window and strengthen the fake/live seam verification.
- Bound or escalate persistent stream-open retries.
- Define and test failed-registration callback ownership.
- Bound or drain rejection diagnostics.
- Remove the new Swift concurrency warnings.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f1624d4011b3`

**Summary:** Changes requested at 9e47e6a. The five prior findings are mechanically addressed, but retry exhaustion creates a false subscription ledger, and transport-level failures can replace an attempt without closing its resources.

**Findings**
- HIGH — RecoveryExecutor.swift:263-280 routes connect and snapshot failures directly through transportFailed; that policy emits only reconnect. dial() then overwrites resourcedAttempt with the replacement, leaving any surviving failed-attempt connection or streams unreachable. A temporary snapshot-failure probe reproduced this: attempt B connected without any closeAll(A). Route current failures through attempt-scoped teardown before installing the replacement claim, while preserving stale-failure rejection.
- HIGH — RecoveryExecutor.swift:309-315 stops retrying after five failures but leaves the pane in SessionRecovery's subscribedPanes ledger. The ledger is documented as panes with open subscriptions, subsequent snapshots deduplicate the pane, and no event retries it until unrelated whole-transport recovery. This recreates the silent-pane condition while reporting the pane as watched. It must be resolved before merge through the required policy review, either with admitted/open states, an exhaustion transition that retracts and schedules recovery, or explicit transport-wide escalation.
- MEDIUM — RecoveryExecutor.swift:102 retains openFailures entries for every failed (attempt,pane) forever. Successful opens clear only their own key; teardown, discard and attempt replacement clear none, and a late failure after retirement also inserts a stale key. Repeated failed attempts therefore create another unbounded diagnostics/state structure. Reject stale catches and retire all counters belonging to completed attempts.
- MEDIUM — The contract-violation test at RecoveryExecutorTests.swift:520 is vacuous on its named axis. ScriptedTransport returns before recording every failed open, so opens == 0 regardless of how many attempts occur; its terminators dictionary also overwrites callbacks by pane, preventing an older registration's closure from being selected. Track attempted registrations separately and retain callbacks per registration. Also change onTermination SHOULD fire to a binding MUST for every non-close terminal stream end: the latch enforces at-most-once, not the at-least-once notification required to prevent silent panes.

**Tests Run**
- Exact head 9e47e6a tested from an isolated archive
- swift build — passed
- swift test — 124 tests passed, 0 failures
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests — 17 tests passed
- git diff --check e02e40b..9e47e6a — passed
- Temporary snapshot-failure regression probe — failed as predicted because closeAll was never called for the replaced attempt

**Needs**
- Run the admitted-versus-open ledger change through the coordinator-required policy review before merging.
- Close failed attempt resources before replacement and pin connect, snapshot, and external transportFailed paths.
- Retire per-attempt open-failure counters.
- Make the seam test observe registration attempts and old callbacks rather than successful-open logging.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f24bd73c66f1`

**Summary:** Changes requested at eea9405. Directive 28002's interim observation is present and the previous resource/counter/seam fixes hold, but paneStatus can still report the opposite of reality under two valid stream-lifecycle interleavings.

**Findings**
- HIGH — RecoveryExecutor.swift models actual streams as an unprovenanced Set<String>. First, streamDied removes a pane before validating the callback's AttemptID; a temporary probe opened p1 on B, delivered A's delayed termination, and B changed from .open to .admittedNotOpen despite B's stream remaining live. Second, if onTermination fires while openStream is suspended and the call later returns, line 402 unconditionally inserts the pane and clears its failures; a second probe exhausted replacement attempts, then the terminated original call returned and paneStatus falsely became .open. Bind observed opens to attempt plus registration generation, let stale callbacks remove only their matching registration, and atomically coordinate termination with successful-return publication through the latch/lifecycle state.
- MEDIUM — The retry boundary is not pinned. RecoveryExecutorTests permits openFailureCap + 1 attempts, and mutating failures < openFailureCap to failures <= openFailureCap left the entire 20-test executor selection green. The cap represents five failed registrations, so assert the exact attempt and backoff counts rather than allowing the off-by-one.

**Tests Run**
- Exact head eea9405 tested from a clean isolated archive
- swift build — passed
- swift test — 127 tests passed, 0 failures
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests — 20 tests passed
- git diff --check 9e47e6a..eea9405 — passed
- Temporary stale-attempt termination probe — failed: live B status became admittedNotOpen
- Temporary termination-before-open-return probe — failed: a dead registration reported open
- Mutation failures < cap to failures <= cap — RecoveryExecutor selection still passed all 20 tests

**Needs**
- Make paneStatus lifecycle and provenance safe without changing task 7's fenced admission semantics.
- Pin the exact open-failure cap boundary.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f3920dbd122b`

**Summary:** The stale-registration status flip and off-by-one retry assertions are fixed, but the termination-during-open path still erases exhausted retry state, and its regression test can miss the intended interleaving.

**Findings**
- **A terminated registration clears the exhausted retry state** (MEDIUM)
  - detail: After openStream returns, openFailures and lastOpenErrors are cleared before publishIfUnterminated decides whether the registration survived. Replaying the shipped interleaving with stronger assertions produced admittedNotOpen(attempts: 0, lastError: nil) after five replacement failures. Firing one retained orphan callback then restarted registration attempts beyond the supposedly exhausted cap. Clear failure bookkeeping only inside the successful publish branch. With that change, the strengthened probe retained attempts == openFailureCap, retained the error, and did not restart.
  - file: Sources/HerdrKit/RecoveryExecutor.swift
  - line: 435
- **The false-open regression does not establish that termination was registered** (MEDIUM)
  - detail: The test waits only until fetchSnapshot is logged, but that log occurs before apply/openStream registers the terminator. killStream can therefore be a no-op, after which releasing the stall publishes a legitimately live stream. I observed this schedule during the probe. Wait for an observable registration count—or make killStream return success and assert it—before firing the death. The test should also assert the cap count and last error, not merely the admittedNotOpen case.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 745

**Tests Run**
- git diff --check eea9405..9409bf6 — passed
- swift build at archived head 9409bf6 — passed
- swift test at archived head 9409bf6 — 129 tests, 0 failures, 1 environment-dependent skip
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests — 22 selected tests passed
- Shipped false-open regression repeated 20 times — passed, but does not prove its registration precondition
- Strengthened deterministic false-open probe — failed at 9409bf6: attempts 0 instead of 5, lastError nil, and an orphan callback restarted opens
- Same strengthened probe with failure resets moved into the successful-publish branch — passed

**Needs**
- Preserve failure counters and the last error when publishIfUnterminated rejects a terminated registration.
- Make the regression wait for and assert registration before firing termination, then assert retained retry diagnostics and that no post-cap attempts restart.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f437e5b6646d`

**Summary:** The production lifecycle fix is correct, and the registration/death precondition is now real. One claimed regression axis remains untested: the orphan-restart check observes a log entry that restarted failing registrations never emit.

**Findings**
- **The no-restart assertion cannot observe restarted failing registrations** (MEDIUM)
  - detail: With onlyFirstOpenSucceeds enabled, every replacement registers its terminator and then throws at line 181, before .openAttempt is recorded at line 183. Therefore attemptsBefore == attemptsAfter remains true even if the orphan callback restarts five registrations. I isolated this by reintroducing the clear-before-decision defect and neutralizing the separate status assertions: the shipped openAttempt equality still passed, while registrationCount exposed growth from the expected 4 registrations to 9. Record .openAttempt before the early throw, or assert directly that registrationCount decreases by exactly the killed callback and never grows.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 779

**Tests Run**
- Reviewed an archive of exact head bac420187085057f237dd5b28aaae015974a374e; the assigned worktree remained at e02e40b
- git diff --check 9409bf6..bac4201 — passed
- swift build and swift test from archived head bac4201 — 129 tests passed, 0 failures, 0 skips
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests from archived head bac4201 — 22 selected tests passed
- Isolated orphan-restart mutation probe — shipped openAttempt equality passed despite five restarted registrations; registrationCount assertion failed with 9 registrations instead of the expected 4

**Needs**
- Make the orphan-restart assertion observe every registration attempt, including the early-throw path.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f4dbe3ec0f1e`

**Summary:** The registration-count fix is correct and the production code remains stable. Two test-harness defects remain: another interleaving test does not prove it entered its named window, and a direct behavioral mutation is converted from an assertion failure into signal 4.

**Findings**
- **The mid-subscribe regression can pass without entering the loop** (MEDIUM)
  - detail: testARetirementMidSubscribeLoopStopsTheRemainingStreams waits for fetchSnapshot, but the fake logs that before returning the snapshot or registering the stalled first stream. I added a post-log snapshot delay and removed the production per-pane binding guard; the mutation built and the test passed because networkChanged retired A before the loop began. Adding registrationCount("a-first") > 0 failed with zero. Wait for the first pane's registration before delivering networkChanged, and assert the kill/window setup directly.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 399
- **The stale-discard regression turns a real kill into signal 4** (MEDIUM)
  - detail: Removing transport.discard compiled and correctly failed the assertion at line 316, but the subsequent firstIndex(...)! crashed with signal 4. Under the standing harness contract this mutation is INVALID, not KILLED, so the test cannot provide classified evidence for its primary behavior. Replace the unwrap with try XCTUnwrap and use the returned index only after it succeeds. The remaining snapshotIndex! at line 266 should receive the same audit.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 324

**Tests Run**
- Observed assigned worktree HEAD e02e40ba3340f31557d42df934b47b4d4106574e before verdict; it remained stale
- Archived exact current branch head 1f279963d448c4f0988905932d36bb507a3268f6, independently matched by task/6-executor and origin/task/6-executor; requested f2c60ec9c1bfae66aa72c9b68e608d352697b1f4 is its ancestor
- Examined diff range bac420187085057f237dd5b28aaae015974a374e..1f279963d448c4f0988905932d36bb507a3268f6; git diff --check passed
- swift build and swift test from archived head 1f279963 — 129 tests, 0 failures, 1 environment-dependent skip
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests from archived head 1f279963 — 22 selected tests passed
- Compiled delayed-snapshot plus removed-per-pane-guard probe — named mid-subscribe test passed without entering the first open; added registration precondition failed with count 0
- Compiled no-discard probe — focused test reported its expected assertion failure, then crashed with signal 4 at discardIndex!

**Needs**
- Arm the mid-subscribe test on an observed first-stream registration rather than the earlier snapshot log.
- Replace the demonstrated force unwrap with XCTUnwrap and audit the remaining force unwrap in this test file.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f58a7bd32a49`

**Summary:** The prior fixes are present, but removing the per-pane binding guard exposed a live catch-path race. A failed open can re-enter through streamFailed, suspend during retry, and then continue opening the next pane after the attempt was retired.

**Findings**
- **Catch-path reentry opens the next pane for a retired attempt** (high)
  - concrete_change: Restore the per-pane current-attempt guard, or guard immediately after await handle(.streamFailed) before the loop can continue. Add the gated catch-path retirement sequence as a regression test.
  - detail: The success path passes through the post-open reap, but the catch path awaits handle(.streamFailed) and then continues the outer loop without rechecking the attempt. A compiled deterministic probe failed: while the nested retry was suspended, networkChanged retired attempt A; after release, the outer loop opened its next pane for A. Restoring the per-pane guard made the same probe pass.
  - file: Sources/HerdrKit/RecoveryExecutor.swift
  - line: 462

**Tests Run**
- Assigned worktree HEAD observed before verdict: e02e40ba3340f31557d42df934b47b4d4106574e; worktree remained clean and stale.
- Independently archived and reviewed current branch head 72b1fccadeb286c1006511fe54d57f3f80424dd1; local and origin-tracking branch refs agreed.
- Examined diff 1f279963d448c4f0988905932d36bb507a3268f6..72b1fccadeb286c1006511fe54d57f3f80424dd1; git diff --check passed.
- Clean archived head: swift build && swift test — 129 tests, 0 failures, 2 skips.
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests — 22 selected tests passed, 0 failures.
- Compiled gated catch-path probe on 72b1fcc failed with a stale open of the next pane; the same probe passed after restoring a current-attempt guard in the disposable archive.
- Force-unwrap audit found no remaining firstIndex, reconnect-attempt, currentAttempt, first, or last force unwraps in RecoveryExecutorTests. Diagnostic changes were confined to a disposable archive; the assigned worktree remained clean.

**Needs**
- Close the catch-path retirement window and add the demonstrated regression before approval.

jerryfane:
HERDR-APP: correction to the round-8 dispatch, posted before the verdict rather than after.

The dispatch asked the reviewer to verify claim (a), "exactly three suspension points in `open(panes:for:)`". That count is wrong as stated. There are **four** awaits in the loop body; the fourth is `transport.closeAll` inside the post-open reap, followed unconditionally by `return`, so it cannot reach the top of the next iteration and needs no recheck.

The invariant is unchanged and so is the code's behaviour — but the imprecision is in exactly the direction that produced last round's HIGH finding (an unenumerated path), so it should not stand in the record uncorrected. The comment at `RecoveryExecutor.swift` now states the qualifier explicitly: *every suspension point that can reach the top of the next iteration* is followed by its own recheck, four awaits, three qualifying.

Pushed as 64623b0 (comment only; no behaviour change, 24/24 executor tests unchanged). Review head for round 8 is now 64623b0 rather than c340f2b — the diff between them is one comment block.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f676b0ca2805`

**Summary:** The production catch-path and backoff guards are correctly placed, but the new catch-path regression does not provide reliable evidence: it arms itself through transport behavior the production contract forbids and its guard-removal mutant can false-pass.

**Findings**
- **Catch-path regression depends on a forbidden orphan registration and lacks a completion signal** (medium)
  - concrete_change: Track open invocations separately from live terminators and remove the throwing call's terminator before propagating its error. Replace the fixed sleep with a deterministic completion condition that waits until either the catch-guard rejection or the forbidden next-pane attempt is observed, then assert the former occurred and the latter did not.
  - detail: ScriptedTransport appends the terminator before the first catch-window open throws, then registrationCount >= 2 counts that forbidden surviving registration plus the retry. ExecutorTransport explicitly requires a throwing open to leave no live registration or future callback. A compiled probe that merely removed the first terminator before throwing caused the unmutated test to fail its precondition: the valid retry produced registrationCount 1, never 2. Separately, the compiled catch-guard-removal mutant passed once before failing on eight subsequent runs; the fixed 300 ms sleep can inspect the negative assertion before the faulty outer loop resumes.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 498
- **The stated suspension-point inventory omits the reap's closeAll await** (low)
  - concrete_change: Scope the invariant to suspension points whose continuation can reach another loop iteration, and explicitly account for the terminal closeAll await.
  - detail: The c340f2b comment says the body contains exactly three suspension points, but await transport.closeAll is a fourth. It is harmless because it is followed unconditionally by return, but the invariant as written is broader than the enumeration.
  - file: Sources/HerdrKit/RecoveryExecutor.swift
  - line: 379

**Tests Run**
- Assigned worktree HEAD observed before verdict: e02e40ba3340f31557d42df934b47b4d4106574e; it remained clean and was not used as the review source.
- At review start, independently resolved and archived exact head c340f2be443c09edf1aff6da4522f09b64851556; both local and origin-tracking task/6-executor refs agreed then. Examined full diff 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..c340f2be443c09edf1aff6da4522f09b64851556 and round delta 72b1fccadeb286c1006511fe54d57f3f80424dd1..c340f2be443c09edf1aff6da4522f09b64851556; git diff --check passed.
- The branch advanced during review, so this verdict intentionally covers the archived c340f2be443c09edf1aff6da4522f09b64851556, not the later refs.
- Clean archived head: swift build && swift test — 131 tests, 0 failures, 2 skips.
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests — 24 selected tests, 0 failures.
- Compiled catch-guard-removal mutant: the named regression passed once, then failed on eight subsequent runs; no compile error or crash.
- Compiled contract-compliance probe retaining production code but removing the throwing fake's terminator: the named regression failed its registrationCount >= 2 precondition.
- Compiled backoff-guard-removal mutant: the named backoff regression failed with 2 opens versus the expected 1.
- All diagnostic edits were confined to a disposable archive and restored byte-for-byte; assigned worktree remained clean.

**Needs**
- Make the catch-window fake obey ExecutorTransport's throwing-open contract and synchronize the regression on completion rather than elapsed time.
- Correct the c340f2b suspension inventory wording.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f79b95dcdbe6`

**Summary:** The production guards and outcome checks are sound, but the catch-path regression still has an attempt-crossing stall race. The unmutated test failed 1 of 20 focused runs, so the suite is not deterministic yet.

**Findings**
- **Pane-scoped catch stall can park both the retired and replacement attempts** (medium)
  - concrete_change: Stall exactly the second catch-window entry, not every later entry, and preferably key/release the gate by attempt or registration so B cannot re-arm A's stall.
  - detail: catchPathPhase stalls every phase >= 2 using stalledPanes keyed only by pane. After attempt A is observed parked, networkChanged starts B. The test can release A, then B's phase-3 open reinserts the same pane into the shared stall before A resumes; both calls remain parked and neither the guard rejection nor forbidden open appears. The unmutated regression reproduced this timeout 1/20. The compiled guard-removal mutant also reached this third outcome 1/11 instead of the intended forbidden-open failure.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 257
- **installsThenThrows retains the same callback twice** (low)
  - concrete_change: For installsThenThrows, retain the registration appended at entry and do not append it again.
  - detail: openStream already appends onTermination at entry, then the installsThenThrows branch appends the identical closure again before throwing. The deliberate violation therefore creates two retained callbacks per failed open while its test and documentation describe one.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 276
- **Catch-window comment still claims registrationCount counts entered opens** (low)
  - concrete_change: Retitle the comment in terms of openEntryCount and the single live retry registration.
  - detail: The comment says registrationCount counts opens entered, contradicting the new openEntryCount separation and the regression's explicit assertion that a throwing open drops its registration.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 249

**Tests Run**
- Reviewed SHA 3b04cd05768d4a992e6fa26c09dea76fee2cfdfb through local plumbing and a git archive; assigned worktree HEAD was e02e40ba3340f31557d42df934b47b4d4106574e.
- git cat-file -t 3b04cd05768d4a992e6fa26c09dea76fee2cfdfb returned commit; inspected diff 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..3b04cd05768d4a992e6fa26c09dea76fee2cfdfb and round delta 64623b0..3b04cd0 using local objects; git diff --check passed.
- Clean archived head: swift build && swift test — 131 tests, 0 failures.
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutorTests — 24 selected tests, 0 failures.
- Unmutated catch-path regression repeated 20 times — 19 passed, 1 failed after five seconds with neither outcome reached.
- Compiled catch-guard-removal mutant repeated 11 times — 10 failed on the intended forbidden open, 1 timed out with neither outcome reached; no compile error or crash.
- Compiled retry-backoff-guard-removal mutant failed on the intended 2-versus-1 open assertion.
- Audited deliberate violation flags: installsThenThrows and onlyFirstOpenSucceeds are each enabled only by their intended test.
- All probe edits were restored byte-for-byte; assigned worktree remained clean.

**Needs**
- Make the catch-window stall attempt- or invocation-specific so the unmutated regression is deterministic.
- Remove the duplicate installsThenThrows registration and correct the stale registrationCount comment.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f84c806d79b4`

**Summary:** Production executor behavior passed review, but the catch-path test seam is still coupled to the ordinary stall seam. A compiled probe reproduced the resulting false precondition deterministically, so the mutation evidence is not yet trustworthy under composed scenarios.

**Findings**
- **The ordinary release permanently pre-releases the supposedly private catch window** (medium)
  - body: release(_:) always inserts the pane into catchReleased, while armCatchPathWindow(_:) neither creates fresh latch state nor clears the previous release. I exercised a normal stalled open, released it to .open, then armed the catch window on the same pane and killed the stream. Entry 2 did not park because the unrelated earlier release had already opened its latch. The compiled probe failed on the intended assertion 5/5. This also confirms that a second catch window on the same pane cannot be represented reliably: catchReleased remains set and openCounts continues beyond phase 2.
  - concrete_change: Give the catch window its own per-arm or per-invocation release handle. Ordinary release(_:) should affect only stalledPanes; arming a catch window should allocate fresh phase and latch state, and a dedicated catch release should target that invocation. Add regressions for ordinary release followed by a catch window and, preferably, two successive catch windows on one pane.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 59

**Tests Run**
- Local plumbing reviewed commit f5b43c9651dbc911d299c73b1d661da1dc7ba8bc; assigned worktree HEAD found before and after review was e02e40ba3340f31557d42df934b47b4d4106574e. Examined diff 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..f5b43c9651dbc911d299c73b1d661da1dc7ba8bc and round delta 3b04cd0..f5b43c9.
- Content discriminator: Sources/HerdrKit/RecoveryExecutor.swift blob 7438b5285d9c745a876eedf9ae9baba3707efd6e.
- Content discriminator: Tests/HerdrKitTests/RecoveryExecutorTests.swift blob f51fa7c80dc2550f235f303801aacaed3ae0adf0.
- Content discriminator, release(_:) last body line verbatim: "        lock.withLock { _ = stalledPanes.remove(pane); _ = catchReleased.insert(pane) }"
- git diff --check 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..f5b43c9651dbc911d299c73b1d661da1dc7ba8bc — passed.
- swift build — passed.
- CLANG_MODULE_CACHE_PATH=/tmp/herdr-pr11-clang-module-cache-f5b43c9 swift test — 131 tests, 0 failures, 1 environment-dependent descriptor-canary skip. The initial run without the writable cache failed closed in PaneSnapshotAccessTests because /root/.cache was read-only; rerun passed.
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutor — 24 tests, 0 failures.
- Compiled probe testAnOrdinaryStallReleaseCannotPreReleaseAFutureCatchWindow — failed on the intended assertion 5/5, with no crash or timeout.
- Probe changes were confined to a temporary archive; assigned worktree remained clean.

**Needs**
- Isolate the catch-window latch from release(_:) and make each arm a fresh invocation before relying on this seam's mutation results.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7f944af8c3962`

**Summary:** The ordinary-release coupling is fixed, but the catch generation still lacks release provenance, and the reap regression still observes a precondition that precedes its hazard. Both were reproduced with compiled probes. No new production-executor defect was found; two listed one-shot seams, releaseAndKill and killDuringFirstSuccessfulOpen, currently have zero call sites.

**Findings**
- **A delayed release is rebound to the current catch generation** (medium)
  - body: releaseCatchWindow(_:) copies catchArm[pane] when the call executes; it does not carry the generation whose window it was created to release. Therefore the monotonicity argument is false: a generation-1 release delayed until generation 2 is current assigns catchReleasedArm = 2 and opens generation 2. A compiled two-generation probe parked generation 2, delivered the stale generation-1 release, and failed on the intended misrelease assertion 6/6.
  - concrete_change: Have armCatchPathWindow return an invocation token or latch and require releaseCatchWindow to receive that identity. Each parked open must retain its own latch so overlapping or delayed releases cannot be rebound through the current pane state. Add the delayed-generation regression.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 131
- **The reap regression waits for snapshot logging before its stream exists** (medium)
  - body: testAnOpenCompletingAfterTeardownIsReaped waits until fetchSnapshot appears in the log, but fetchSnapshot records that operation before returning and before openStream registers its terminator. A diagnostic transport that suspended immediately after recording the snapshot made the shipped wait settle with registrationCount("only") == 0; the direct precondition probe failed 3/3. This permits a scheduler interleaving where the baseline or an unrelated mutant fails because no late-open window existed, corrupting kill attribution.
  - concrete_change: Wait for and assert the actual registration for "only" before retiring the attempt, as the other death/open-window regressions do. Prefer an attempt-provenanced registration observable if the fake later permits multiple attempts to register the same pane before retirement.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 954

**Tests Run**
- Local plumbing reviewed d61683971ab6e53a1ed0903390fa067c957abf9e; assigned worktree HEAD was e02e40ba3340f31557d42df934b47b4d4106574e before and after review. Examined 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..d61683971ab6e53a1ed0903390fa067c957abf9e and round delta f5b43c9..d616839.
- Content discriminator 1: catchReleased was replaced by catchReleasedArm; release(_:) now touches only stalledPanes.
- Content discriminator 2: RecoveryExecutorTests contains exactly 25 test methods. The 26-test RecoveryExecutor filter includes the separate LiveRecoveryExecutorTests method.
- Content discriminator 3: releaseCatchWindow call sites are line 648 in testARetirementDuringAFailedOpensHandlingStopsTheRemainingStreams, line 703 in testOrdinaryReleaseCannotOpenAParkedCatchWindow, and line 720 in testASecondCatchWindowOnTheSamePaneArmsIndependently.
- git diff --check 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..d61683971ab6e53a1ed0903390fa067c957abf9e — passed.
- swift build — passed.
- swift test with a writable Clang module cache — 133 tests, 0 failures.
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutor — 26 tests, 0 failures.
- Existing CatchWindow seam tests — 2 tests per round, 5/5 rounds passed.
- Delayed generation-release probe: first version was INVALID due to await inside an XCTest autoclosure and was excluded; corrected probe built and failed on the intended assertion 6/6 without crashes or timeouts.
- Snapshot-precondition probe built and failed directly on registrationCount == 0 after the shipped fetchSnapshot wait 3/3.
- All diagnostic edits were confined to a temporary archive; the assigned worktree remained clean.

**Needs**
- Bind catch releases to the arm that created them rather than consulting the current generation.
- Make the reap regression wait for the actual open registration before retiring its attempt.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7fa487010f640`

**Summary:** The release-provenance fix and reap precondition are correct, and removing the two dead termination seams did not remove production coverage. One shared-slot race remains: phase classification and CatchWindow capture are separate operations, allowing an old invocation to park the newly armed window and make retryIsParked pass for the wrong reason.

**Findings**
- **Phase classification is not atomically bound to its CatchWindow** (medium)
  - body: openStream first calls catchPathPhase(pane), then separately reads catchWindows[pane] at line 312. If armCatchPathWindow replaces the current object between those lock acquisitions, a phase-2 invocation from the old arm parks the new CatchWindow. A deterministic hook paused after phase classification, armed a replacement, then resumed: first.isParked was false, second.isParked was true, and retryIsParked returned true. The compiled probe failed both intended assertions 6/6. Thus the new seam tests can report an armed current window even though the parked invocation belongs to the previous arm.
  - concrete_change: Return the phase and CatchWindow reference together from one locked operation, for example catchPathStep(_:) -> (phase: Int, window: CatchWindow?). openStream must use that captured object without consulting catchWindows again. Add the in-flight re-arm regression asserting the old invocation parks only its original object and cannot make the replacement satisfy retryIsParked.
  - file: Tests/HerdrKitTests/RecoveryExecutorTests.swift
  - line: 303

**Tests Run**
- Local plumbing reviewed 969ad35fa50a94c88f757fae23cae984d5ca94ba; assigned worktree HEAD was e02e40ba3340f31557d42df934b47b4d4106574e before and after review. Examined 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..969ad35fa50a94c88f757fae23cae984d5ca94ba and round delta d616839..969ad35.
- Content discriminator 1: armCatchPathWindow returns ScriptedTransport.CatchWindow. The phase-2 open stores a local CatchWindow reference fetched from catchWindows[pane] after phase classification; that delayed lookup is the finding.
- Content discriminator 2: RecoveryExecutorTests contains exactly 26 test methods; swift test --filter RecoveryExecutor executes 27 because it also includes LiveRecoveryExecutorTests.testExecutorColdStartsAgainstTheLiveServer.
- Content discriminator 3: releaseAndKill and killDuringFirstSuccessfulOpen were removed with killOnRegister and succeedOnceThenKill. Both zero-call-site seams fired a registered terminator inside openStream before it returned, retaining unused contract-violating machinery rather than merely dead helpers.
- Reviewed test-file blob: 008abe212fcdeffed042e7ab70dfbca807474c97.
- git diff --check 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..969ad35fa50a94c88f757fae23cae984d5ca94ba — passed.
- swift build — passed.
- swift test with a writable Clang module cache — 134 tests, 0 failures, 1 environment-dependent descriptor-canary skip.
- swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutor — 27 tests, 0 failures.
- In-flight phase/re-arm probe built and failed both intended ownership assertions 6/6, without crashes or timeouts.
- Mutation replacing if once.publishIfUnterminated() with if true built and was KILLED by testATerminationDuringTheOpenCannotPublishAFalseOpen with status open, confirming the remaining stall-plus-external-kill route covers termination before openStream returns.
- All diagnostic edits were confined to temporary archives; the assigned worktree remained clean.

**Needs**
- Atomically bind each catch-path phase decision to the CatchWindow object it belongs to.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7fb761124cbed`

**Summary:** The Round 13 fix correctly binds phase classification to its captured window, but two test-harness integrity defects remain: arm membership is still decided after entry, and a compiling no-op release mutation escapes by hanging.

**Findings**
- **An invocation already in flight can consume a catch window armed after it entered.** (medium)
  - concrete_change: Capture the current CatchWindow and claim its phase atomically with openStream entry/registration, preferably keeping the phase on the captured window. An arm installed after entry must not classify that existing invocation.
  - evidence: openStream records entry and registers its terminator under one lock, but catchPathStep classifies it under a later acquisition at line 335. A deterministic hook in that real lock gap reproduced 6/6: the pre-arm invocation consumed phase 1 and threw; the first genuinely post-arm invocation became phase 2, parked, and then succeeded. This violates the stated 'entries since this arm' model and can make retryIsParked observe the wrong invocation.
  - location: Tests/HerdrKitTests/RecoveryExecutorTests.swift:310
- **Catch-window tests can hang instead of failing when release is broken.** (medium)
  - concrete_change: Make the parked loop cancellation-aware, positively wait for and assert the release outcome, then cancel the task as unconditional independent teardown before awaiting it. Apply the same pattern to every catch-window test that currently releases and immediately awaits.
  - evidence: Replacing CatchWindow.release() with a compiling no-op made testOrdinaryReleaseCannotOpenAParkedCatchWindow time out at both 5 and 20 seconds. The test calls release and then awaits the parked task through that same mechanism, with no positive release assertion or independent teardown, so the mutation is ESCAPED rather than KILLED.
  - location: Tests/HerdrKitTests/RecoveryExecutorTests.swift:713

**Tests Run**
- Reviewed 777f048e17afdebe75641538d58fcd7bf69df33a through local object plumbing and clean git archives; assigned worktree HEAD before and after review was the stale e02e40ba3340f31557d42df934b47b4d4106574e and remained clean.
- Examined full PR range 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..777f048e17afdebe75641538d58fcd7bf69df33a and incremental range 969ad35..777f048; git diff --check passed.
- Content discriminator: private func catchPathStep(_ pane: String) -> (phase: Int, window: CatchWindow?) returns the phase and its captured CatchWindow from one lock acquisition.
- Content discriminator: the regression installs afterPhaseClassification; it is required to expose the otherwise unobservable post-classification/pre-park point where a replacement arm is introduced.
- Content discriminator: the regression releases both first and replacement.value; releasing only first left the mutant parked on the replacement and made the run hang/ESCAPE.
- RecoveryExecutorTests contains exactly 27 test methods; the RecoveryExecutor filter contains 28 including one LiveRecoveryExecutorTests method. Test-file blob: aaa984da2dc56ffd69c44a43fb7caa367d87d3d8.
- swift build passed; warning-as-errors compilation and the Round 13 regression passed; the shipped regression passed 6/6 total runs.
- swift test --filter RecoveryExecutor passed 28/28.
- Full swift test passed 135 tests with 0 failures and 1 environment-dependent descriptor-canary skip.
- Pre-arm classification probe built and failed on all three intended assertions in 6/6 runs.
- CatchWindow.release no-op mutant built; focused test timed out at both 5 seconds and 20 seconds, classified ESCAPED rather than KILLED.

**Needs**
- Bind catch-window phase membership at open entry so an older in-flight invocation cannot consume a later arm.
- Give catch-window regressions positive release assertions and cancellation-aware independent teardown so broken-release mutations fail rather than hang.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7fc62582efbd4`

**Summary:** Both Round 13 fixes are correct, and the production executor’s three resumable suspension paths remain independently covered. One remaining harness hang exists: the synchronous phase-1 expectThrow call can park forever under a compiling off-by-one phase mutation.

**Findings**
- **The catch-window tests still have an unbounded await before they obtain a cancellable task handle.** (medium)
  - concrete_change: Make the phase-1 expectation bounded and cancellation-backed: run the open in a task, positively observe completion versus parking, fail if it parks or times out, and always cancel/release before awaiting the task. Apply through expectThrow or a catch-window-specific helper so an off-by-one phase mutation produces an assertion failure rather than a hang.
  - evidence: Changing CatchWindow.claimPhase() from returning phase to returning phase + 1 builds successfully. The first open is then classified as phase 2 and parks inside expectThrow at line 697. Because the test awaits expectThrow synchronously, it cannot release the window or cancel the call; the focused test timed out after 30 seconds and is therefore ESCAPED, not KILLED. All five catch-window tests use this same phase-1 pattern.
  - location: Tests/HerdrKitTests/RecoveryExecutorTests.swift:396

**Tests Run**
- Reviewed 2c301291a27809065aacef6feaac1ff0f646d3f2 through local object plumbing and clean archives; assigned worktree HEAD before and after review was the stale e02e40ba3340f31557d42df934b47b4d4106574e and remained clean.
- Examined full PR range 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..2c301291a27809065aacef6feaac1ff0f646d3f2 and incremental range 777f048..2c30129; git diff --check passed.
- Content discriminator: the phase counter is now CatchWindow.phase, claimed by CatchWindow.claimPhase() during the open-entry lock; this removes the pane-keyed catchPhase dictionary and its separate entry/classification acquisition.
- Content discriminator: after release, every catch-window test positively waits for and asserts that its window unparked, then cancels its task as unconditional teardown before awaiting it.
- Content discriminator: the compiling release-is-a-noop mutation previously timed out/ESCAPED; it now fails testOrdinaryReleaseCannotOpenAParkedCatchWindow with 'release did not unpark the window' and is KILLED.
- Object evidence: RecoveryExecutor.swift blob 7438b5285d9c745a876eedf9ae9baba3707efd6e; RecoveryExecutorTests.swift blob 3e3b0e7dd686bce123bda48144315e85458901f0.
- RecoveryExecutorTests contains 27 methods; the RecoveryExecutor filter contains 28 tests including one LiveRecoveryExecutorTests method.
- swift build passed; swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutor passed 28/28; full swift test passed 135/135 with no failures or skips.
- Production audit: independently removed the retry-backoff recheck, post-open reap, and catch-path recheck. All three mutants built and were KILLED by their named behavioral regressions with direct assertion failures. I found no lost production coverage or new resource-ownership defect in RecoveryExecutor.swift.
- Hang probe: claimPhase returning phase + 1 built, entered testOrdinaryReleaseCannotOpenAParkedCatchWindow, and timed out after 30 seconds while its synchronous expectThrow remained parked; classified ESCAPED.

**Needs**
- Bound and independently tear down the phase-1 expectThrow path so a parked first entry fails instead of hanging.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7fd50a7a4d911`

**Summary:** Round 15’s hang fix is sound, and the requested production probes found no runtime defect. One diagnostics-contract coverage gap remains: drainRejections() returns and resets the overflow count, but neither behavior is tested; two compiling mutants survive.

**Findings**
- **The drain test does not verify the dropped-count contract.** (medium)
  - concrete_change: Assert drained.dropped equals the pre-drain dropped count and droppedRejections is zero afterward, preferably also verifying a second drain returns zero dropped entries.
  - evidence: testRejectionsAreBoundedAndDrainable checks the pre-drain counter, drained entry count, and empty entry buffer only. A mutant returning dropped: 0 SURVIVED, and a separate mutant omitting droppedRejections = 0 also SURVIVED; both built and passed the focused test.
  - location: Tests/HerdrKitTests/RecoveryExecutorTests.swift:1219

**Tests Run**
- Reviewed ae923bb9a9e2e5e771147cca4cb81cff2eae5cc7 through local Git plumbing; assigned worktree HEAD was stale at e02e40ba3340f31557d42df934b47b4d4106574e. Examined 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..ae923bb and the focused 2c30129..ae923bb change.
- Content discriminator: expectPhaseOneThrows replaced 5 call sites. When an open parks, it releases the CatchWindow and cancels the task before awaiting its result.
- Content discriminator: claimphase-off-by-one now builds and is KILLED by assertion failures; the focused regression exited with three failures in about seven seconds instead of hanging/ESCAPING.
- Blob verification: RecoveryExecutor.swift 7438b5285d9c745a876eedf9ae9baba3707efd6e; RecoveryExecutorTests.swift d7bd4123a4957138c57428b44a6b29a626140fb9.
- Static hang sweep was exhaustive over syntactically unbounded awaits in RecoveryExecutorTests.swift, but not every imaginable composite mutant. All active task-result waits now have release or cancellation teardown, bounded waits remain bounded, and expectThrow has zero call sites.
- streamFailed re-entry probe: removing the catch-path handle(.streamFailed) call built and was KILLED by testAPersistentlyFailingOpenCannotSpin.
- paneStatus exhaustion probe: forcing the reported attempt count to zero built and was KILLED by testATerminationDuringTheOpenCannotPublishAFalseOpen.
- Diagnostics probes: drain-loses-dropped-count and drain-retains-dropped-count both built and SURVIVED testRejectionsAreBoundedAndDrainable.
- swift build passed; swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutor passed 28/28; full swift test passed 135 tests with 0 failures and 1 documented environment-dependent SSH canary skip.
- git diff --check passed. The assigned worktree remained clean, and every temporary probe file was restored to the reviewed blob hash.

**Needs**
- Add assertions covering both the returned dropped count and its reset during drainRejections().

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c7fe0ae8e679b3`

**Decision:** `approved`
**Summary:** No—nothing left that I can find should block this merge. Round 16 fully pins the drain contract, both former survivors are real kills, the production blob is unchanged, and the final gates are clean. Further rounds are unlikely to buy proportionate confidence.

**Tests Run**
- Reviewed 8b80726d57c14c3aad5618793722cfa4be5f6a6a through local Git plumbing; assigned worktree HEAD remained stale at e02e40ba3340f31557d42df934b47b4d4106574e. Examined 61a6ba1805262d95a0d8bbdf6f7a726c1f71befd..8b80726 and incremental ae923bb..8b80726.
- Blob verification: RecoveryExecutor.swift 7438b5285d9c745a876eedf9ae9baba3707efd6e; RecoveryExecutorTests.swift d29f6713e4195e34e788ec79939dbb510e5062b0.
- Content discriminator: testRejectionsAreBoundedAndDrainable now asserts drained.dropped equals the pre-drain count, droppedRejections resets to zero, and a second drain returns neither entries nor drops. The third contract assertion establishes that dropped is a delta since the preceding drain rather than a cumulative total.
- drain-loses-dropped-count: previously SURVIVED, now compiled and KILLED by the returned-count equality (0 versus 20).
- drain-retains-dropped-count: previously SURVIVED, now compiled and KILLED by both the immediate reset check and the second-drain delta check.
- swift build passed; swift test -Xswiftc -warnings-as-errors --filter RecoveryExecutor passed 28/28.
- The first full run exposed a read-only Clang module-cache path; after redirecting it to /tmp, a live SSH acknowledgment check transiently failed under load and immediately passed in isolation. The final full swift test passed 135 tests with 0 failures and 1 environment-dependent live canary skip.
- git diff --check passed. Both mutation probes restored the reviewed blob exactly, and the assigned worktree remained clean.

PR #14, opened 2026-08-02
test(scripts): report which tests killed a mutant, and stop miscounting kills

jerryfane:
Verification-infrastructure only.

**Scope disclosure**: this branch grew after its first review round at `ce17297`. A production change (`.noteIgnoredDeath`) had accumulated here; it is **split out into #15** and reverted from this branch. A production change should not ride in on a scripts PR.

## 1. Report WHICH tests killed a mutant (`ce17297`)

A kill is a **pattern**, not a verdict. On #13 round one a compiling no-op survived the two tests written for it and was killed by an unrelated one — that split was the proof those two tests were vacuous, available a full round before any premise mutation, and thrown away because the output said only `KILLED`.

The same mutation today reports `killed by 5 of 32`. At #13 round one it would have reported **1**.

## 2. Stop miscounting kills (`0dd7f31`) — from #14 round one

**A crash after a failure was reported as KILLED** (high). A two-suite probe: suite 1 completed one assertion failure, suite 2 died with signal 4. The failure line and suite 1's count were both present, so the run reported `KILLED (1 tests)` and hid the crash. "A crash is not a kill" has been in this lane's header since #11 — my own harness violated it whenever a failure preceded the crash. `EXECUTED` now comes from the terminal summary rather than the largest component suite.

**Skipped tests read as passing ones** (low). A probe where the *intended* regression skipped and an unrelated test killed the mutant printed `killed by 1 of 2` — indistinguishable from the premise-suspect pattern, while the intended test never ran its body. This repo has conditional live-server and descriptor-canary skips. Skips are now printed separately.

Verified: a deliberate `fatalError` probe reports INVALID rather than KILLED.

## 3. Review header (`9d82763`, `74e6048`)

Premise mutation as a standing clause, then directive 30193's settled aim: **does the assertion pin the REASON, or only the OUTCOME?** Absence-shape is where the class clusters, not what it is. `9d82763` was stranded on a local branch and never shipped — #13 merged without it.

## 4. One test strengthening (`d20f30c`)

`testEachRoundTripGetsItsOwnChannel` asserted both responses were non-empty, under a name claiming each round trip gets its own channel. Any non-empty text passed. It now asserts each response carries its own request id and that the two differ.

139 tests, 0 failures.

PR #13, opened 2026-08-02
feat(kit): exhaustion retracts the admission, and the reason survives it

jerryfane:
Task 7. Closes #12, under coordinator ruling 29761 (option B).

## The gap

The ledger recorded **admission** — intent, at observe time — and nothing retracted it. A pane whose stream could not be opened after 5 attempts stayed listed while nothing watched it: subscribed in name, silent in fact. #11 could only *report* that; this closes it.

## What changed

**Exhaustion is a transition.** `.streamExhausted(pane:from:)` retracts through `dropAndReadmit(readmit: false)` — the same authority section a real death uses, already reviewed, because exhaustion is a death that happened before the stream existed. No new mechanism, and **the ledger gains no intermediate state**, which is why this did not reopen #10's exactly-once invariants.

**It does not re-admit.** The failure counter is keyed by `(attempt, pane)` and no ledger operation clears it, so re-admitting on the same attempt would hit the cap guard again — correct, but a plan churned per resync for a pane already hopeless on this connection. The pane is not abandoned: a new attempt retires the counters and adoption's `resyncAllPanes` re-admits from a fresh snapshot. That global re-admission **is** the recovery schedule; there is no per-pane retry queue and this does not add one.

**Absence is no longer one fact**, so `paneStatus` stopped deriving `.notAdmitted` from bare ledger membership:

```
absent + failures >= cap  ->  .admittedNotOpen(attempts:lastError:)
absent + no failures      ->  .notAdmitted
```

The discriminator is `openFailures`, which needed **no second store** because `paneStatus` already read it, and which survives the retraction precisely because no ledger operation touches it. It is documented at its declaration as load-bearing, with the two properties that are now contractual rather than convenient.

## Guards rose with the promotion

The paired control proves the surface *discriminates*; these prove it survives the transitions that now depend on it. All four were required by the ruling:

| case | property |
|---|---|
| exhaustion during adoption | a retired attempt's exhaustion cannot retract the live one |
| exhaustion racing a real death | retracts exactly once — both paths go through the same was-in-the-ledger check |
| re-admission after exhaustion | a new connection is a genuine fresh chance |
| attempt rollover | a pane exhausted on attempt N reports 0 attempts on N+1, not the inherited count |

The rollover case is correct *by the keying* rather than by a check, which is exactly why it is tested explicitly — it would break silently if someone re-keyed the counter.

[...]

PR #15, opened 2026-08-02
feat(kit): make an ignored stream death observable, so its guard can arm

jerryfane:
Split out of #14, which had grown past its title. This is a **production** change and should not ride in on a scripts PR.

## The finding

Directive 30217's authoring rule — *when asserting path P did not disturb state Y, also assert P ran* — applied to this suite found one:

`testExhaustionRacingARealDeathRetractsExactlyOnce` asserts that a real stream death arriving **after** exhaustion does not re-admit the pane. **Deleting the `handle(.streamFailed)` line left it passing.** The setup established absence directly, so "still absent" was true before the path ran, and the test could not distinguish *harmless* from *absent*.

Neither existing technique caught it. The absence-shape grep flagged the test and I cleared it, because the reviewer's #13 round-3 premise sweep had already probed it — but that sweep mutated the **exhaustion** premise. Nobody mutated the **death** premise. Two premises, one swept.

## Why this needed production code

The assertion was **unwritable**, not merely unwritten: ignoring the death produced no observable at all. A path with no success signal cannot be armed by any test, ever — which is a defect in the surface, not only in the test. `.noteIgnoredDeath` exists solely to make a correct no-op a fact instead of an absence.

**Its reason string names the class, not a cause**, deliberately: `dropAndReadmit` returns nil for three distinct situations — stale attempt, no live connection, no ledger entry — and my first version labelled all three "no ledger entry". That is the three-states-into-one flattening this lane has spent the day removing, committed while fixing an instance of it. Proving the path *ran* does not require knowing which of the three it was.

## Knock-on

`testStreamFailuresThatCannotBeActedOnAreInert` asserted `plan.isEmpty` three times — outcome-shaped, and unable to distinguish an inert plan from an event that never reached the policy. Each case now asserts **both**: no acting action, and the death was processed.

## Evidence

```
premise-real-death-never-delivered   SURVIVED -> KILLED
inert-plan-loses-its-note            KILLED (2 of 67)
full suite                           139 tests, 0 failures
```

### DIFF
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -225,6 +225,18 @@ public actor RecoveryExecutor {
     /// plans and duplicate callbacks are rare but unbounded over days.
     public private(set) var rejections: [(action: String, reason: String)] = []
     public private(set) var droppedRejections = 0
+
+    /// How many stream deaths arrived for a pane the ledger could not act on —
+    /// a duplicate, a late callback, or one already retracted by exhaustion.
+    ///
+    /// Deliberately a COUNTER on its own surface rather than an entry in
+    /// `rejections`: these are unbounded in principle (a conformer double-firing
+    /// terminations produces one per fire) and sharing the bounded buffer let
+    /// them evict actionable refusals, measured at 129 deaths destroying a
+    /// seeded real rejection. An observation that degrades the diagnostics it
+    /// sits next to is not an observation.
+    public private(set) var ignoredDeaths = 0
+    public private(set) var lastIgnoredDeathPane: String?
     static let rejectionCapacity = 128
 
     private func record(rejection: (action: String, reason: String)) {
@@ -339,20 +351,19 @@ public actor RecoveryExecutor {
                 await resync(for: attempt)
 
             case .noteIgnoredDeath(let pane, let from):
-                // The ONLY effect is that the no-op becomes observable. See the
-                // case's own documentation: ignoring was already correct, and
-                // was also indistinguishable from the event never arriving.
-                // The reason names the CLASS, not a cause. `dropAndReadmit`
-                // returns nil for three distinct situations — a stale attempt,
-                // no live connection, or no ledger entry — and this note cannot
-                // tell them apart. Claiming one would be the three-states-into-
-                // one flattening this lane has spent the day removing; the note
-                // exists to prove the path RAN, which does not require knowing
-                // which of the three it was.
-                record(rejection: (
-                    action: "streamFailed(\(pane))",
-                    reason: "death not actionable (stale attempt, disconnected, or no ledger entry)"
-                ))
+                // A COUNTER, NOT A REJECTION ENTRY, and that distinction is a
+                // review finding rather than a preference. Putting these in the
+                // shared 128-entry buffer meant a flood of duplicate deaths
+                // EVICTED ACTIONABLE REFUSALS: a probe seeded one real
+                // "bound to a retired attempt" rejection, then delivered 129
+                // ignored deaths, and the real entry was gone. A path added for
+                // VISIBILITY had started destroying visibility.
+                //
+                // Counting cannot evict anything and is O(1). It is also
+                // sufficient for what this exists to do — prove the path RAN —
+                // which does not require a message per occurrence.
+                ignoredDeaths += 1
+                lastIgnoredDeathPane = pane
                 _ = from
 
             case .subscribe(let panes, let on):
--- a/Sources/HerdrKit/RecoveryExecutor.swift
+++ b/Sources/HerdrKit/RecoveryExecutor.swift
@@ -338,6 +338,23 @@ public actor RecoveryExecutor {
             case .resyncAllPanes(let attempt):
                 await resync(for: attempt)
 
+            case .noteIgnoredDeath(let pane, let from):
+                // The ONLY effect is that the no-op becomes observable. See the
+                // case's own documentation: ignoring was already correct, and
+                // was also indistinguishable from the event never arriving.
+                // The reason names the CLASS, not a cause. `dropAndReadmit`
+                // returns nil for three distinct situations — a stale attempt,
+                // no live connection, or no ledger entry — and this note cannot
+                // tell them apart. Claiming one would be the three-states-into-
+                // one flattening this lane has spent the day removing; the note
+                // exists to prove the path RAN, which does not require knowing
+                // which of the three it was.
+                record(rejection: (
+                    action: "streamFailed(\(pane))",
+                    reason: "death not actionable (stale attempt, disconnected, or no ledger entry)"
+                ))
+                _ = from
+
             case .subscribe(let panes, let on):
                 // THE binding check. A plan can sit queued while the world
                 // moves; an action bound to a retired attempt must be refused
--- a/Sources/HerdrKit/SessionRecovery.swift
+++ b/Sources/HerdrKit/SessionRecovery.swift
@@ -305,6 +305,17 @@ public enum RecoveryAction: Equatable, Sendable {
     /// `Wire.swift` also models non-pane-scoped kinds this plan does not
     /// express.)
     case subscribe(Set<String>, on: AttemptID)
+
+    /// A stream death was received for a pane the ledger does not hold, and
+    /// deliberately ignored — a duplicate, a late callback, or one for a pane
+    /// already retracted by exhaustion.
+    ///
+    /// It exists ONLY to make a correct no-op observable. Ignoring was already
+    /// right; it was also indistinguishable from the event never arriving, so
+    /// the guard asserting "a death after retraction does not re-admit" passed
+    /// with the death removed entirely. A path with no success signal cannot be
+    /// armed by any test.
+    case noteIgnoredDeath(pane: String, from: AttemptID)
 }
 
 /// An ordered list of steps. Order is the point.
@@ -627,7 +638,20 @@ public struct SessionRecovery: Sendable {
             // favour, and a later snapshot corrects any miss.
             guard let readmitted = state.authority.dropAndReadmit(
                 pane, from: from, readmit: state.knownPanes.contains(pane)
-            ) else { return RecoveryPlan() }
+            ) else {
+                // A death for a pane the ledger does not hold: a duplicate, a
+                // late callback, or one for a pane already retracted by
+                // exhaustion. Correct to ignore — and until now, INVISIBLE.
+                //
+                // A path that produces no observable when it behaves correctly
+                // cannot be distinguished from a path that never ran, by any
+                // test, ever. That is not a testing inconvenience: the guard
+                // asserting "a death after retraction does not re-admit" passed
+                // with the death never delivered, and no premise mutation could
+                // have armed it without something to observe. The refusal is
+                // recorded so the no-op is a fact rather than an absence.
+                return RecoveryPlan([.noteIgnoredDeath(pane: pane, from: from)])
+            }
             return subscriptionPlan(for: readmitted)
 
         case .streamExhausted(let pane, let from):
--- a/Tests/HerdrKitTests/RecoveryExecutorTests.swift
+++ b/Tests/HerdrKitTests/RecoveryExecutorTests.swift
@@ -264,20 +264,14 @@ final class ScriptedTransport: ExecutorTransport, @unchecked Sendable {
         return seconds
     }
 
-    /// Parks every connect until released, so a test can hold a dial IN FLIGHT.
-    /// Needed because `pendingDial`'s only observable effect is whether an
-    /// in-flight dial gets CANCELLED — a landed one is discarded and shows in
-    /// the log, a cancelled one never records. Without a way to keep one in
-    /// flight, that state is unpinnable and a mutation clearing it survives.
-    private var stallConnects = false
-    func stallConnect() { lock.withLock { stallConnects = true } }
-    func releaseConnect() { lock.withLock { stallConnects = false } }
-
+    // REMOVED: the connect stall (stallConnect/releaseConnect/parkedConnects).
+    // It was added to pin `pendingDial` and has ZERO call sites now that the
+    // assertion it served was cut. Dead seams in this fake are not inert — two
+    // were removed from here in #14 for firing a terminator inside openStream
+    // while nothing called them — so it goes rather than sitting armed for
+    // whoever finds it next. `git log` has it if the pendingDial gap is ever
+    // closed.
     func connect(for attempt: AttemptID) async throws {
-        while lock.withLock({ stallConnects }) {
-            if Task.isCancelled { throw CancellationError() }
-            try? await Task.sleep(nanoseconds: 10_000_000)
-        }
         record(.connect(attempt))
         if failConnects { throw TransportError.closedBeforeResponse }
     }
@@ -1640,23 +1634,19 @@ actor SleepRecorder {
         let backedOff = await waitUntil { await backoffs.recorded().count > backoffsBefore }
         XCTAssertTrue(backedOff, "the note replaced the sleeper; nothing waits any more")
 
-        // AND THE PENDING-DIAL HANDLE SURVIVED. Its only observable effect is
-        // whether an in-flight dial gets CANCELLED when the next one starts: a
-        // cancelled dial never records a connect, a landed one does. So park a
-        // dial, deliver a note, then start another dial — the parked one must
-        // not appear.
-        transport.stallConnect()
-        await executor.handle(.networkChanged(at: Date()))
-        let parkedAttemptOptional = await executor.currentAttempt
-        let parkedAttempt = try XCTUnwrap(parkedAttemptOptional)
-        let inFlight = await waitUntil(1) { false }   // let the dial reach the stall
-        _ = inFlight
-        await executor.handle(.streamFailed(pane: "gone", from: parkedAttempt))   // another note
-        await executor.handle(.networkChanged(at: Date()))                        // supersedes it
-        transport.releaseConnect()
-        try? await Task.sleep(nanoseconds: 300_000_000)
-        XCTAssertFalse(transport.log().contains(.connect(parkedAttempt)),
-                       "the superseded dial was never cancelled; the note dropped pendingDial")
+        // NOT PINNED: pendingDial. Its only observable effect is whether an
+        // in-flight dial gets CANCELLED, and every version I wrote to observe
+        // that raced the cancellation against the stall release — flaky at
+        // first, then wrong 4 runs in 5 once I made the wait deterministic.
+        // Reporting it as uncovered rather than shipping a test that fails on
+        // correct code: an unreliable guard is worse than a named gap, because
+        // the gap is visible and the flake gets muted.
+        //
+        // A mutation clearing pendingDial from this branch therefore SURVIVES.
+        // What it would cost in production: a superseded dial is never
+        // cancelled, lands late, and is discarded — wasteful, not incorrect,
+        // since the discard path exists for exactly that. That is why I am
+        // willing to leave it named rather than keep grinding.
     }
 
     /// AXIS: after retraction, a genuine re-admission works — the pane is not
--- a/Tests/HerdrKitTests/RecoveryExecutorTests.swift
+++ b/Tests/HerdrKitTests/RecoveryExecutorTests.swift
@@ -1072,6 +1072,21 @@ final class RecoveryExecutorTests: XCTestCase {
                       "the dropped duplicate must be observable")
     }
 
+/// Observes a parked dial's cancellation. `pendingDial` holds the task that
+/// runs the sleeper, so a cancelled dial surfaces as `Task.sleep` throwing —
+/// the one place that state is directly visible.
+actor DialGate {
+    private var entered = false
+    private var cancelled = false
+    private var released = false
+    func enter() { entered = true }
+    func markCancelled() { cancelled = true }
+    func release() { released = true }
+    var hasEntered: Bool { entered }
+    var wasCancelled: Bool { cancelled }
+    var isReleased: Bool { released }
+}
+
 actor SleepRecorder {
     private var delays: [TimeInterval] = []
     func record(_ delay: TimeInterval) { delays.append(delay) }
@@ -1602,6 +1617,8 @@ actor SleepRecorder {
         let backoffsBefore = await backoffs.recorded().count
         XCTAssertGreaterThan(backoffsBefore, 0, "the sleeper never fired; its assertion would be vacuous")
         let rejectionsBefore = await executor.rejections.count
+        XCTAssertGreaterThan(rejectionsBefore, 0,
+                             "the rejection buffer is empty; its assertion below would compare 0 to 0")
 
         // The note.
         await executor.handle(.streamFailed(pane: "p", from: attempt))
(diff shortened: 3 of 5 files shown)


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c803d6d116f0a7`

**Decision:** `failed`
**Summary:** checkout head is 0871cb381e513b4f5834306b89a3dd8f8bfd1941, not review job head 0871cb3
**Diagnostics:** checkout head is 0871cb381e513b4f5834306b89a3dd8f8bfd1941, not review job head 0871cb3

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c8051b49b893f9`

**Summary:** The observable ignored-death path is correctly emitted and its class-level reason is honest, but two issues remain: ignored-death notes can evict actionable diagnostics from the shared bounded buffer, and the executor tests do not guard the claim that executing the note performs no transport work.

**Findings**
- **Ignored-death visibility consumes and can displace the actionable rejection budget** (Do not place ignored-death observations in the same eviction budget as actionable rejections. Coalesce them, use a separate bounded observation/counter surface, or otherwise guarantee they cannot evict actionable entries; add the demonstrated flood regression.)
  - evidence: Through the public event path, I seeded one "bound to a retired attempt" rejection, exhausted pane p, then delivered 129 streamFailed events for the now-unadmitted pane. The ignored-death notes evicted the actionable rejection and raised droppedRejections to 2. Thus the new supposedly observational path changes which real refusals remain available to diagnostics.
  - location: Sources/HerdrKit/RecoveryExecutor.swift:341
  - severity: medium
- **The executor half of the note's inertness contract is unguarded** (Strengthen the executor regression to positively arm the ignored-death path and assert the transport log and relevant executor state are unchanged except for the dedicated observation. The closeAll-after-record mutation should fail that test.)
  - evidence: I mutated the noteIgnoredDeath executor branch to retain the expected rejection and then call transport.closeAll(for: from). The mutation built and survived both strengthened regressions and the complete 32-test RecoveryExecutor filter. The policy test proves the marker exists, and the race test proves ledger absence plus recording, but neither proves execution performed no transport action.
  - location: Tests/HerdrKitTests/RecoveryExecutorTests.swift:1433
  - severity: medium

**Tests Run**
- Dispatch requested 5e4d0b8, but git cat-file reported that object absent locally and no local or remote ref contained it. The reachable live branch and assigned worktree both resolved to 0871cb381e513b4f5834306b89a3dd8f8bfd1941, which I archived and reviewed. Examined a2fbd9cf07b76f6e041caf93d14fda0a20955ce9..0871cb381e513b4f5834306b89a3dd8f8bfd1941.
- Content discriminator: the new action is noteIgnoredDeath(pane:from:). The executor records the exact reason "death not actionable (stale attempt, disconnected, or no ledger entry)".
- Content discriminator: the reason names a class because dropAndReadmit returns nil for three indistinguishable causes—stale attempt, disconnected state, or missing ledger entry—so the note cannot truthfully select one cause.
- Content discriminator: testStreamFailuresThatCannotBeActedOnAreInert previously asserted isEmpty for stale-attempt, unknown-pane, and closed-pane plans. Each case now asserts both that no action other than noteIgnoredDeath exists and that noteIgnoredDeath is present, proving no acting action and that the path ran.
- Verified reviewed blobs: SessionRecovery.swift 0c6112b9d729ceca4c003824a6bd621d6c179e1d; RecoveryExecutor.swift 3c03080236acc2a90b1b2388e99c76324ae4c443; RecoveryExecutorTests.swift 93a3ff468f567a3392769f3a7ca627a995a54580; SessionRecoveryTests.swift cf77281a324d3d0d6e9dafe09664079991df400e.
- Premise mutation deleting the streamFailed delivery: the strengthened executor regression failed on "the death was never delivered".
- Production mutation removing noteIgnoredDeath emission: built and killed both strengthened tests; all three policy cases and the executor race reported that the path never ran.
- Premise mutations neutralizing the stale-attempt, unknown-pane, and pane-closed setups: all three cases failed on "the plan acted"; no additional changed-test premise gap found.
- Public-path diagnostic flood probe: failed because 129 ignored deaths evicted the seeded actionable rejection and incremented droppedRejections twice.
- Executor-side-effect mutation adding closeAll after recording noteIgnoredDeath: built and survived both named regressions and all 32 RecoveryExecutor-filtered tests.
- Pristine swift build passed; full suite passed with 139 tests, 0 failures, 1 descriptor-canary skip.
- All mutations and temporary probes were restored; reviewed file hashes match the archived head, diff check passed, and the assigned worktree is clean.

**Needs**
- Isolate or coalesce ignored-death observations so they cannot evict actionable rejection diagnostics.
- Add an executor-level regression proving noteIgnoredDeath records its observation without performing any transport or recovery action.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c805f3707d677c`

**Summary:** The counter design is bounded and does not evict rejection diagnostics, but the observation-only contract remains incompletely pinned: two compiling executor-state mutations survived the relevant suite.

**Findings**
- **The ignored-death regression does not prove executor-state inertness** (medium)
  - body: Tests/HerdrKitTests/RecoveryExecutorTests.swift checks the ledger, diagnostic counter, last pane, and immediate transport-log count, but not openFailures or resource ownership. Adding `openFailures = [:]` to the noteIgnoredDeath branch built and survived all 140 tests; it changes the exhausted pane's public status from admittedNotOpen(attempts: cap, ...) to notAdmitted. Independently, adding `resourcedAttempt = nil` built and survived all 33 executor-filter tests; the next teardown would skip closeAll for the owned attempt. Add a regression that snapshots paneStatus across the note and then exercises teardown, proving the old attempt is still closed. A second live pane can similarly pin openStreams directly.

**Tests Run**
- Reviewed pinned head archived through local plumbing: 09055e93539de4fb2a80d66c314987bf19ad4637. Assigned worktree HEAD found: 0871cb381e513b4f5834306b89a3dd8f8bfd1941. Examined diff a2fbd9cf07b76f6e041caf93d14fda0a20955ce9..09055e93539de4fb2a80d66c314987bf19ad4637; git diff --check passed.
- Content discriminator: ignored deaths are recorded in RecoveryExecutor.ignoredDeaths, with lastIgnoredDeathPane retaining the most recent pane; they are not recorded in rejections or its 128-entry budget.
- Content discriminator: testExhaustionRacingARealDeathRetractsExactlyOnce now asserts transport.log().count is unchanged across delivery of the ignored death, independently of its ledger assertions.
- Content discriminator: the flood regression asserts 129 ignored deaths, exactly rejectionCapacity + 1, because exceeding the 128-entry buffer by one reproduces eviction if ignored deaths share that buffer and proves every delivered death was counted.
- Reviewed blob IDs: RecoveryExecutor.swift e54b0234048615b7740d1ef77b116431ad64b1bb; SessionRecovery.swift 0c6112b9d729ceca4c003824a6bd621d6c179e1d; RecoveryExecutorTests.swift 00f6cafb7c9e2c2f7084f195fc79a03d5ff80e4d.
- openFailures-clearing mutant: built successfully; RecoveryExecutor filter passed 33/33 and full suite passed 140/140, so it SURVIVED.
- resourcedAttempt-clearing mutant: built successfully; RecoveryExecutor filter passed 33/33, so it SURVIVED.
- openStreams-clearing mutant: built successfully; both new regressions passed, but the older stale-termination regression failed, so the suite KILLED it indirectly.
- Clean archived head: full suite passed 140 tests with 0 failures.
- All probe edits were restored; archived RecoveryExecutor.swift hashes to the reviewed blob, and the assigned worktree remains clean.

**Needs**
- Extend the ignored-death inertness regression to pin non-transport executor state, particularly paneStatus/openFailures and resourcedAttempt teardown ownership.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c80691394a2c81`

**Summary:** Two test-contract gaps remain. I found no direct production defect, but the observation-only claim is still insufficiently pinned, so this head should not merge yet.

**Findings**
- MEDIUM — Tests/HerdrKitTests/RecoveryExecutorTests.swift:1491: the ownership regression waits for any historical `.closeAll(attempt)`, not one caused by the post-death teardown. Demonstration: `resourcedAttempt = nil` is killed alone, but the same compiling mutant passes when an unrelated `transport.closeAll(for: attempt)` occurs before the ignored death. Assert a new occurrence after `networkChanged`—for example, snapshot the matching count immediately beforehand and require it to increase.
- MEDIUM — Sources/HerdrKit/RecoveryExecutor.swift:353: five independent compiling mutations in `noteIgnoredDeath` still survived all 33 executor-filter tests: clearing `lastOpenErrors`, clearing `pendingDial`, replacing `sleeper` with a no-op, clearing `state.knownPanes`, and resetting `droppedRejections`. In production these would respectively erase the reported failure cause, lose the cancellation handle for an in-flight dial, remove future retry backoff, forget panes needed for subsequent death/readmission decisions, and discard the accumulated rejection-drop delta. The regression should pin these remaining state invariants, preferably with positive preconditions proving each state is non-default before delivering the note.

**Tests Run**
- Reviewed pinned commit 7130b99b957428780985302346755eca882d983b through local plumbing and a `/tmp` archive; assigned worktree HEAD was stale at 0871cb381e513b4f5834306b89a3dd8f8bfd1941. Diff range: a2fbd9cf07b76f6e041caf93d14fda0a20955ce9..7130b99b957428780985302346755eca882d983b.
- Content verification: RecoveryExecutor.swift blob e54b0234048615b7740d1ef77b116431ad64b1bb; RecoveryExecutorTests.swift blob 87bec17c3b1ccff997d109212c08c68470e7d903.
- Content discriminator: `note-clears-openFailures` would turn an exhausted `.admittedNotOpen(attempts: cap, ...)` pane into `.notAdmitted`; it now builds and is killed by `testAFloodOfIgnoredDeathsCannotEvictActionableRejections`. `note-clears-resourcedAttempt` would make teardown skip `closeAll` for the socket-owning attempt; it now builds and is killed by `testExhaustionRacingARealDeathRetractsExactlyOnce`.
- Content discriminator: after delivering the ignored death, the race regression sends `.networkChanged`, waits for `.closeAll(attempt)`, and asserts it occurred because ownership is observable only through teardown behavior. `note-clears-openStreams` was already killed by `testAStaleTerminationCannotFlipALivePanesStatus`, which fails when the live pane becomes `admittedNotOpen`.
- Remaining-state mutation sweep: `lastOpenErrors = [:]`, `pendingDial = nil`, no-op `sleeper`, `state.knownPanes = []`, and `droppedRejections = 0` all built and survived the 33-test executor filter. `rejections = []` built and was killed by the flood regression.
- Wrong-reason probe: with both `resourcedAttempt = nil` and an unrelated earlier `closeAll(attempt)`, `testExhaustionRacingARealDeathRetractsExactlyOnce` passed, proving the final `contains` assertion can consume stale evidence.
- Clean baseline: Swift build succeeded; full suite passed, 140 tests. `git diff --check` passed. Every probe was restored; archived source blob IDs match 7130b99 and the assigned worktree remains clean.

**Needs**
- Make the teardown assertion require a new matching `closeAll` caused by `networkChanged`.
- Add armed invariance coverage for `lastOpenErrors`, `pendingDial`, `sleeper`, `knownPanes`, and `droppedRejections` across `noteIgnoredDeath`.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c808cab32ab96f`

**Summary:** The round-three fixes work, but the declared pendingDial gap is both testable and more costly than documented. One of the five new state assertions also lacks its claimed positive precondition.

**Findings**
- HIGH — Tests/HerdrKitTests/RecoveryExecutorTests.swift:1637: `pendingDial` need not remain unpinned. A deterministic test using the existing sleeper seam passed 20/20 on correct code and killed `pendingDial = nil` 5/5. It parks a delayed reconnect in a cancellation-aware sleeper, positively observes entry, delivers the ignored death, invokes `networkChanged`, and requires that exact sleeper to observe cancellation; an unconditional release cleans up the mutant. No connect stall or release race is required.
- MEDIUM — Tests/HerdrKitTests/RecoveryExecutorTests.swift:1645: the documented production cost is understated. One lost handle permits a late stale dial that is eventually discarded, but repeated ignored deaths and replacements can accumulate multiple orphan dial tasks. Because a delayed orphan may begin connecting after its attempt-scoped `closeAll` already ran, it can retain connection resources or bounded worker capacity until completion, potentially degrading availability rather than causing only one wasted discard.
- MEDIUM — Tests/HerdrKitTests/RecoveryExecutorTests.swift:1604: `rejectionsBefore` is captured but never asserted nonzero, unlike the other four surfaces. A compiling premise mutation that discarded every rejection while incrementing `droppedRejections` satisfied all existing preconditions and the test still passed with `rejectionsBefore == rejectionsAfter == 0`. Add `XCTAssertGreaterThan(rejectionsBefore, 0, ...)` so the rejection-buffer assertion pins a reason rather than an empty outcome.

**Tests Run**
- Reviewed pinned commit 191d600890a00f32369ed3aa652ca93ecfc6c459 through local Git plumbing and a `/tmp` archive. Assigned worktree HEAD was stale at 0871cb381e513b4f5834306b89a3dd8f8bfd1941. Merge base: a2fbd9cf07b76f6e041caf93d14fda0a20955ce9.
- Content blobs: Sources/HerdrKit/RecoveryExecutor.swift e54b0234048615b7740d1ef77b116431ad64b1bb; Tests/HerdrKitTests/RecoveryExecutorTests.swift a2b3ae3312adc92d918b1f36d77a45ab257dddcf.
- Content discriminator: the deliberately unpinned state is `pendingDial`; the site says clearing it leaves a superseded dial uncancelled so it lands late and is discarded—described there as wasteful but not incorrect.
- Content discriminator: the ownership assertion snapshots the count of `.closeAll(attempt)` entries and requires that count to increase after `networkChanged`, replacing historical membership via `contains`.
- Content discriminator: the removed seam is the connect-stall trio `stallConnect` / `releaseConnect` / `parkedConnects`; it had zero call sites after the attempted pendingDial test was removed, and dead fake seams are hazardous rather than inert.
- Ownership stale-evidence probe: a compiling `resourcedAttempt = nil` mutant plus an unrelated earlier `.closeAll(attempt)` was killed by the new count-increase assertion. The previous `contains` assertion accepted this exact setup.
- Five independent compiling note mutations were killed by `testTheIgnoredDeathNoteLeavesEveryOtherSurfaceAlone`: clearing `lastOpenErrors`, clearing `state.knownPanes`, resetting `droppedRejections`, replacing `sleeper` with a no-op, and clearing `rejections`.
- Pending-dial probe: the cancellation-aware sleeper regression passed 20/20 on correct code. With the compiling `pendingDial = nil` mutation, it failed on its intended cancellation assertion 5/5 and completed through unconditional release rather than hanging.
- Rejection premise probe: replacing rejection recording with drop-all behavior while keeping `droppedRejections > 0` compiled, all stated preconditions passed, and `testTheIgnoredDeathNoteLeavesEveryOtherSurfaceAlone` passed with an empty buffer.
- Clean build succeeded. Full suite passed, 141 tests. `git diff --check` passed. All probe edits were restored; archived source/test blob IDs match 191d600 and the assigned worktree remains clean.

**Needs**
- Add the deterministic sleeper-based pendingDial cancellation regression or an equivalent non-racy guard.
- Assert the rejection buffer is non-empty before comparing it across the note.
- Update the pendingDial gap assessment if it remains documented, accounting for repeatable orphan-task and resource accumulation.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c809c4768d1055`

**Summary:** There are still blocking guard gaps: the new dial assertion accepts premature cancellation as correct evidence, and two additional executor-owned states remain unpinned.

**Findings**
- [high] Tests/HerdrKitTests/RecoveryExecutorTests.swift:1679-1684 — The test proves only that the dial is eventually cancelled, not that `networkChanged` cancelled it. A compiling mutation adding `pendingDial?.cancel()` to `noteIgnoredDeath` SURVIVED the focused test and all 34 executor tests. The note cancels a legitimate delayed reconnect, but the immediately following network change masks the defect and the gate attributes stale cancellation evidence to teardown. In production this can leave the executor without its scheduled reconnect until another external event. Require a positive liveness handshake from the parked task after the note, then require cancellation only after networkChanged.
- [high] Tests/HerdrKitTests/RecoveryExecutorTests.swift:1608 — `state.consecutiveFailures` remains outside the purported every-surface assertion. Adding `state.consecutiveFailures = 0` to the note branch compiled and SURVIVED all 34 executor tests. An ignored stale death during a reconnect streak would then reset the next failure to first-attempt backoff, defeating the exponential backoff and potentially sustaining a reconnect storm. Drive the streak non-default before the note and verify the next policy delay retains it.
- [medium] Tests/HerdrKitTests/RecoveryExecutorTests.swift:1608 — The executor's jitter generator is also unpinned. A note branch that executes `_ = generator.next()` compiled and SURVIVED all 34 executor tests, so an observation changes subsequent reconnect timing. A seeded-generator assertion on the next backoff can pin this without wall-clock timing.

**Tests Run**
- Reviewed pinned SHA e976b9da95e3c26ec0ef9f95ea09e98f9eb25d98 through local Git plumbing and an archived tree. Assigned worktree HEAD was stale at 0871cb381e513b4f5834306b89a3dd8f8bfd1941; it remained clean.
- Content blobs: Sources/HerdrKit/RecoveryExecutor.swift=e54b0234048615b7740d1ef77b116431ad64b1bb; Sources/HerdrKit/SessionRecovery.swift=0c6112b9d729ceca4c003824a6bd621d6c179e1d; Tests/HerdrKitTests/RecoveryExecutorTests.swift=b40c1c3cb55d6e834334c1450ac4871a279803dc; Tests/HerdrKitTests/SessionRecoveryTests.swift=cf77281a324d3d0d6e9dafe09664079991df400e.
- Content discriminator: pendingDial is observed through the sleeper seam. The pending task itself runs the sleeper, so cancellation makes its `Task.sleep` throw before transport.connect runs; transport observation occurs later and races release against cancellation. Verbatim gate line: `guard delay >= 4 else { return }          // only the parked reconnect`.
- Content discriminator: the fifth positive precondition is `XCTAssertGreaterThan(rejectionsBefore, 0,`; before it existed, a premise mutation left the assertion comparing an empty buffer before and after, so 0 == 0 passed without arming the surface.
- Content discriminator: the documented cost is now accumulating orphan dial tasks whose connects can begin after closeAll and retain connection resources until stale completion is discarded. Last round claimed only one late discarded dial—wasteful, not incorrect. Verbatim text: `+ "the orphan keeps its connection resources until it completes")`.
- Focused ignored-death regression: 12 consecutive clean runs, 0 failures.
- RecoveryExecutor selection baseline: 34 tests, 0 failures. SessionRecoveryTests baseline: 35 tests, 0 failures.
- `pendingDial = nil` in noteIgnoredDeath built and was KILLED by the focused regression.
- Premise mutations all built and were KILLED: empty knownPanes, no stale-plan seeding/dropped count, a non-failing pane/no recorded error, a sleeper that records nothing, and a reconnect below the dial-gate threshold.
- `pendingDial?.cancel()` in noteIgnoredDeath built and SURVIVED both the focused regression and all 34 executor tests.
- `state.consecutiveFailures = 0` in noteIgnoredDeath built and SURVIVED both the focused regression and all 34 executor tests.
- Advancing the jitter generator in noteIgnoredDeath built and SURVIVED all 34 executor tests.
- All mutations were restored. Every reviewed archive file hashes back to its e976b9d blob; the assigned worktree is clean.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c80b16bad49e83`

**Summary:** Two test-harness defects still block merge. The paired control is flaky on unchanged code and intermittently lets both target mutations survive. The pending-dial check is a timed absence assertion, not a positive liveness handshake, and can credit teardown for cancellation initiated by the note.

**Findings**
- **The paired control observes asynchronously scheduled dials and is neither stable nor reliably mutation-killing.** (high)
  - details: After each transportFailed event, the test waits only for currentAttempt to change, which happens synchronously before dial's Task necessarily records its delay. Outstanding earlier tasks can therefore record later, reorder the arrays, or satisfy the final `recorded().count > beforeNote` wait without the post-note failure being observed. On unchanged f7f2995, 8 of the first 19 completed focused runs failed with reordered arrays or three-versus-four entries. The compiling `state.consecutiveFailures = 0` mutation survived 2 of 4 probes, and `_ = generator.next()` survived 1 of 4. Require each individual failure's newly scheduled delay to be recorded before advancing, and assert every wait precondition.
  - line: 1727
  - path: Tests/HerdrKitTests/RecoveryExecutorTests.swift
- **The claimed liveness handshake is a 150 ms negative observation and does not prove the dial remained alive after the note.** (high)
  - details: The test sleeps 150 ms and asserts `gate.wasCancelled == false`. If cancellation reporting is delayed, that absence is credited as liveness; the later positive cancellation can then be incorrectly attributed to networkChanged. A premise probe delayed `markCancelled` by 300 ms: unchanged code passed, then the compiling production mutation adding `pendingDial?.cancel()` to noteIgnoredDeath also SURVIVED. Replace the timed absence check with a positive heartbeat or request/acknowledgement that only the still-running parked dial can produce after the note and before networkChanged.
  - line: 1688
  - path: Tests/HerdrKitTests/RecoveryExecutorTests.swift

**Tests Run**
- Reviewed pinned SHA f7f29958035f3faaeb3d15f6926f286137b3d227 through local plumbing and an archived tree. Assigned worktree HEAD was stale at 0871cb381e513b4f5834306b89a3dd8f8bfd1941. Merge base with origin/main: a2fbd9cf07b76f6e041caf93d14fda0a20955ce9.
- Content blobs: Tests/HerdrKitTests/RecoveryExecutorTests.swift = 6c318b946659bf61efab1e6e6dca95e13b34220d; Sources/HerdrKit/RecoveryExecutor.swift = e54b0234048615b7740d1ef77b116431ad64b1bb; Sources/HerdrKit/SessionRecovery.swift = 0c6112b9d729ceca4c003824a6bd621d6c179e1d. Critical verbatim lines include `try? await Task.sleep(nanoseconds: 150_000_000)` and `generator: SeededGenerator(seed: 42))`.
- Content discriminator ENGAGED: between the ignored-death note and networkChanged, the regression sleeps 150 ms, reads `gate.wasCancelled`, and asserts it is false; this exists to kill the mutation that calls `pendingDial?.cancel()` from noteIgnoredDeath.
- Content discriminator ENGAGED: `state.consecutiveFailures` and the executor's jitter generator are pinned through paired delay recordings rather than direct reads. Both are private behind RecoveryExecutor, so their intended observable is the next policy-computed reconnect delay. Both executor instances use SeededGenerator seed 42.
- Focused unchanged paired-control stress: 8 failures among the first 19 completed runs. Failures showed identical-seed arrays reordered or containing different numbers of delays.
- Compiling note-clears-consecutiveFailures probes: SURVIVED 2 of 4, KILLED 2 of 4. Compiling note-consumes-generator probes: SURVIVED 1 of 4, KILLED 3 of 4. Because the unchanged test also flakes, the kills are not independently attributable.
- Compiling note-cancels-pendingDial mutation with the unmodified test was KILLED by the new assertion.
- Premise probe delayed cancellation observation by 300 ms. The unchanged focused test passed, and the compiling note-cancels-pendingDial production mutation then SURVIVED, demonstrating false attribution across the 150 ms absence window.
- All mutations and premise probes were restored. Source and test blobs matched f7f2995 afterward; a fresh archive comparison was identical excluding build artifacts. Assigned worktree remained clean with zero status entries.

**Needs**
- Make the paired control await and positively assert each expected delay before triggering the next failure, then demonstrate both target mutations are killed repeatedly on a stable baseline.
- Replace the 150 ms no-cancellation assertion with a positive post-note liveness acknowledgement from the parked dial.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c80c3a0564ac58`

**Decision:** `approved`
**Summary:** Approved. No blocking findings: the positive heartbeat excludes a cancelled task's possible final beat, and the serialized paired control now ties every injected failure to its own recorded delay. Compiling probes reproduced all three intended kills.

**Tests Run**
- Code identity: reviewed pinned SHA 8dee2ecada0b289d16b30454f9eb7099e9df5e71 through local Git plumbing; merge-base with origin/main was a2fbd9cf07b76f6e041caf93d14fda0a20955ce9. Assigned worktree HEAD was stale at 0871cb381e513b4f5834306b89a3dd8f8bfd1941 and remained clean.
- Content discriminator: Tests/HerdrKitTests/RecoveryExecutorTests.swift blob 80ff53f23dd252d5180c1808499a7384b0abbc6e. The parked dial now emits heartbeats via the verbatim line `await gate.beat()` and proves continued liveness with `let stillAlive = await waitUntil { await gate.heartbeats > beatsBefore + 1 }`. Previously it slept 150ms and merely asserted that cancellation had not yet been reported.
- Each paired-control loop iteration snapshots the delay count, injects one transport failure, then waits for its own delay using the verbatim line `let recorded = await waitUntil { await recorder.recorded().count > before }` before causing the next failure. Previously it waited only for currentAttempt to change.
- Run-count discriminator: unchanged focused tests improved from 8 failures in 19 runs to 0 failures in 20 runs. The consecutiveFailures mutant improved from KILLED 2/4 to KILLED 4/4; the generator-consumption mutant improved from KILLED 3/4 to KILLED 4/4. I independently reproduced the current 0/20, 4/4, and 4/4 results.
- Compiling `pendingDial?.cancel()` probe: KILLED 4/4 by the heartbeat/cancellation assertions, with no build failures or crashes. This demonstrates that a cancelled loop's possible final beat cannot satisfy the required two-beat increase.
- Full Swift 6.1.3 Linux suite: 142 tests executed, 21 environment-dependent tests skipped, 0 failures.
- Restoration verification: archived RecoveryExecutor.swift hashed to pinned blob e54b0234048615b7740d1ef77b116431ad64b1bb and RecoveryExecutorTests.swift to pinned blob 80ff53f23dd252d5180c1808499a7384b0abbc6e after all probes; assigned worktree remained clean.

PR #16, opened 2026-08-02
fix(kit): attribute descriptor audits, and make the host-key test detect anything

jerryfane:
The task-1 flake — and the flake turned out to be the smaller half.

## What I went looking for

`testHostKeyRejectionThroughStreamNeverClosesADescriptorTwice` failed roughly **1 run in 10** under full-suite load and never in isolation. It was hit twice during #11's review rounds.

Cause: `DescriptorAudit.doubleCloses` is **process-wide**, and the test bracketed it with a baseline across 25 stream attempts. Any other test's async teardown landing in that window counts as this transport's defect. The comment there said a baseline handled tests sharing a process — it handled *sequential* sharing, not concurrent.

## What I found instead

**The assertion had never detected anything.** A deliberate double close, inserted at the site the test drives, **survived**.

The audited branch in `LiveChannel` is only reached when no session exists. Host-key rejection happens *after* the handshake, so teardown went through paths calling `Glibc_close` directly and the audit never saw them. The test was a meter measuring nothing — and the process-wide counter then made the nothing wobble.

I found this by mutating the close site rather than by reading, after two wrong guesses about which path the test drives.

## Both halves fixed

**Attribution.** `DescriptorAudit.close` takes an owner; `doubleCloses(by:)` answers for one logical transport. `SSHTransport` carries an `AuditToken` — a reference inside a value type, deliberately, because copies of the struct are the same logical transport, and every channel and session it opens inherits the token.

**Coverage.** Every descriptor close on the failure paths now goes through the audit: four sites in `openSession`, `Session.closeSocket`, and `LiveChannel`'s. The host-key rejection path is one of them, which is what makes the test's stated claim testable at all.

## Evidence

```
deliberate double close on the host-key path   SURVIVED -> KILLED
full suite                                     142 tests, 0 failures / 6 runs
the test in isolation                          0 failures / 12 runs
```

## Honest limit

I could **not** reproduce the flake in 8 full-suite runs before the fix. At a 1-in-10 rate that is weak evidence, not absence. The fix rests on the **mechanism** — a process-wide counter read across a window other tests write to — not on a reproduction I did not obtain.

### DIFF
--- a/Sources/HerdrKit/SSHTransport.swift
+++ b/Sources/HerdrKit/SSHTransport.swift
@@ -176,7 +176,7 @@ public struct SSHTransport: HerdrTransport {
     /// Every channel it opens inherits this token, so a test can ask what THIS
     /// transport did without reading a process-wide tally that every other
     /// test also writes to.
-    final class AuditToken {}
+    final class AuditToken: Sendable {}
     let auditToken = AuditToken()
 
     /// Shared, bounded pool for blocking libssh2 work. Bounded because a thread
@@ -552,7 +552,7 @@ public struct SSHTransport: HerdrTransport {
             // Published before the first call that can block on it, so an
             // interrupt arriving mid-connect has a real descriptor to shut down.
             guard live?.adopt(rawSocket: fd) ?? true else {
-                _ = Glibc_close(fd)
+                DescriptorAudit.close(fd, owner: auditToken)
                 throw CancellationError()
             }
             do {
@@ -568,7 +568,7 @@ public struct SSHTransport: HerdrTransport {
                 // published number before closing it — otherwise `close()` closes
                 // it a second time, by which point it may name another socket.
                 live?.release()
-                _ = Glibc_close(fd)
+                DescriptorAudit.close(fd, owner: auditToken)
                 guard case SSHError.connectFailed(_, _, let failure) = error else {
                     // A deadline or an interrupt ends the attempt outright: both
                     // are global, and trying the next address cannot help.
--- a/Sources/HerdrKit/SSHTransport.swift
+++ b/Sources/HerdrKit/SSHTransport.swift
@@ -176,7 +176,30 @@ public struct SSHTransport: HerdrTransport {
     /// Every channel it opens inherits this token, so a test can ask what THIS
     /// transport did without reading a process-wide tally that every other
     /// test also writes to.
-    final class AuditToken: Sendable {}
+    /// Carries its own audit state, so nothing is keyed by an ADDRESS.
+    ///
+    /// The counters and the injectable closer used to live in
+    /// `[ObjectIdentifier: ...]` maps, which are keyed by a pointer value that
+    /// the allocator reuses: review released a token WITHOUT calling forget(),
+    /// obtained a replacement at the same address, and watched the retired
+    /// token's closer run for it. Manual cleanup is not a lifetime; holding the
+    /// state on the object IS one, and it deletes the whole class of hazard
+    /// rather than adding a rule people must remember.
+    final class AuditToken: @unchecked Sendable {
+        private let lock = NSLock()
+        private var _closer: @Sendable (Int32) -> Int32 = { DescriptorAudit.realClose($0) }
+        private var _doubleCloses = 0
+        private var _rejections = 0
+
+        var closer: @Sendable (Int32) -> Int32 {
+            get { lock.withLock { _closer } }
+            set { lock.withLock { _closer = newValue } }
+        }
+        var doubleCloses: Int { lock.withLock { _doubleCloses } }
+        var adoptionRejections: Int { lock.withLock { _rejections } }
+        fileprivate func recordDoubleClose() { lock.withLock { _doubleCloses += 1 } }
+        fileprivate func recordRejection() { lock.withLock { _rejections += 1 } }
+    }
     let auditToken = AuditToken()
 
     /// Internal seam: fires once a descriptor exists and BEFORE it is offered to
@@ -303,7 +326,7 @@ public struct SSHTransport: HerdrTransport {
         let sock: Int32
         let handle: OpaquePointer
         /// Who a double close here is attributed to. See DescriptorAudit.
-        var auditOwner: AnyObject?
+        var auditOwner: SSHTransport.AuditToken?
 
         init(sock: Int32, handle: OpaquePointer) {
             self.sock = sock
@@ -978,7 +1001,7 @@ final class LiveChannel: @unchecked Sendable {
     /// and never held by a caller, so attributing to `self` gives a test
     /// nothing it can ask about; the TRANSPORT outlives every channel it makes
     /// and is what a test actually holds.
-    var auditOwner: AnyObject?
+    var auditOwner: SSHTransport.AuditToken?
     private let lock = NSLock()
     private var session: SSHTransport.Session?
     private var channel: OpaquePointer?
@@ -1113,7 +1136,7 @@ final class LiveChannel: @unchecked Sendable {
         if let session {
             session.closeSocket()
         } else if raw >= 0 {
-            DescriptorAudit.close(raw, owner: auditOwner ?? self)
+            DescriptorAudit.close(raw, owner: auditOwner)
         }
     }
 }
@@ -1128,148 +1151,53 @@ final class LiveChannel: @unchecked Sendable {
 enum DescriptorAudit {
     private static let lock = NSLock()
     private static var doubleCloseCount = 0
-    private static var byOwner: [ObjectIdentifier: Int] = [:]
+    private static var generations: [Int32: Int] = [:]
 
-    /// Closes, and records the close of a descriptor that was not open.
-    ///
-    /// `owner` attributes the double close to the object whose teardown caused
-    /// it. Without it the tally was PROCESS-WIDE, and the only test reading it
-    /// bracketed a window with a baseline — which is safe only if nothing else
-    /// in the process closes a descriptor during that window. Other tests'
-    /// async teardown does exactly that, so the assertion failed roughly one
-    /// run in ten under full-suite load and never in isolation. The comment
-    /// here used to say the baseline handled sharing a process; it handled
-    /// SEQUENTIAL sharing, not concurrent.
-    static func close(_ fd: Int32, owner: AnyObject? = nil) {
-        guard Glibc_close(fd) != 0, errno == EBADF else { return }
-        lock.lock()
-        doubleCloseCount += 1
-        if let owner { byOwner[ObjectIdentifier(owner), default: 0] += 1 }
-        lock.unlock()
+    static func realClose(_ fd: Int32) -> Int32 { Glibc_close(fd) }
+
+    /// Closes, and records a double close against the token that owned it.
+    static func close(_ fd: Int32, owner: SSHTransport.AuditToken?) {
+        guard (owner?.closer ?? realClose)(fd) != 0, errno == EBADF else { return }
+        lock.lock(); doubleCloseCount += 1; lock.unlock()
+        owner?.recordDoubleClose()
     }
 
-    /// Process-wide tally. Kept for diagnostics; NOT sound for a test
-    /// assertion, because any concurrent teardown lands inside the window.
+    /// Process-wide tally. NOT sound for a test assertion — any concurrent
+    /// teardown lands inside the window. Ask a token instead.
     static var doubleCloses: Int {
         lock.lock(); defer { lock.unlock() }
         return doubleCloseCount
     }
 
-    /// Double closes attributable to ONE object — the only form a test can
-    /// assert on without depending on what the rest of the process is doing.
-    static func doubleCloses(by owner: AnyObject) -> Int {
-        lock.lock(); defer { lock.unlock() }
-        return byOwner[ObjectIdentifier(owner)] ?? 0
-    }
-
-    /// Adoption rejections, attributed. The rejection branch closes its
-    /// descriptor and throws — and BOTH of those are indistinguishable from what
-    /// the ordinary cleanup does one level down, so a test could not tell the
-    /// branch had run at all. Bypassing it entirely compiled and left its own
-    /// regression green. A path with no success signal cannot be armed by any
-    /// test; this is that signal.
-    private static var rejectionsByOwner: [ObjectIdentifier: Int] = [:]
-    /// Closes a descriptor whose adoption was refused, and records the rejection
-    /// ONLY BECAUSE the close happened.
-    ///
-    /// ONE operation, because two independent ones diverge. Recording and
-    /// closing were separate calls, so deleting the close compiled and left the
-    /// counter incrementing: the test saw "one rejection, zero double closes"
-    /// and passed, while the descriptor LEAKED. A positive observable that does
-    /// not originate from the act it witnesses is just a second claim.
-    /// Records that a descriptor number has just been opened, and returns the
-    /// GENERATION that identifies this particular use of it.
-    ///
-    /// Descriptor numbers are recycled: close 7, open a socket, and the kernel
-    /// hands back 7. `EBADF` therefore distinguishes only the NUMBER'S current
-    /// state, never which opening it belonged to — a probe closed a descriptor,
-    /// opened another that reused the number, and the audit reported the second
-    /// one's close as the first one's success. A generation is the only thing
-    /// that can tell those apart.
-    /// The close used for an owner's descriptors — real by default, injectable
-    /// so a test can witness that a site invoked it.
-    private static var closers: [ObjectIdentifier: @Sendable (Int32) -> Int32] = [:]
-    static func setCloser(_ closer: @escaping @Sendable (Int32) -> Int32, for owner: AnyObject) {
-        lock.lock(); closers[ObjectIdentifier(owner)] = closer; lock.unlock()
-    }
-    static func realClose(_ fd: Int32) -> Int32 { Glibc_close(fd) }
-    private static func closer(for owner: AnyObject?) -> @Sendable (Int32) -> Int32 {
-        guard let owner else { return { Glibc_close($0) } }
-        lock.lock(); defer { lock.unlock() }
-        return closers[ObjectIdentifier(owner)] ?? { Glibc_close($0) }
-    }
-
+    /// Records that a descriptor NUMBER has just been opened, returning the
+    /// generation that identifies this particular use of it. Numbers recycle,
+    /// so EBADF alone can never say which opening a close belonged to.
     static func opened(_ fd: Int32) -> Int {
         lock.lock(); defer { lock.unlock() }
         generations[fd, default: 0] += 1
         return generations[fd] ?? 0
     }
-    private static var generations: [Int32: Int] = [:]
 
-    /// Closes the descriptor an adoption refused, and records the rejection only
-    /// on PROOF that THIS generation of THAT descriptor is now gone.
+    /// Closes the descriptor an adoption refused, and records the rejection
+    /// only because the close ran.
     ///
-    /// Two things are verified rather than assumed, because both were exploited:
-    ///   - the GENERATION still matches, so a recycled number cannot be closed
-    ///     in place of the one the caller meant;
-    ///   - the close goes through an INJECTABLE, owner-scoped closer, so a test
-    ///     can prove this site actually invoked it. That replaces an earlier
-    ///     post-close `fcntl(F_GETFD)` check which was itself racy — once close
-    ///     returns the number is free, another thread can reuse it, and a
-    ///     healthy rejection then went unrecorded under load.
+    /// Verified rather than assumed: the GENERATION still matches, so a
+    /// recycled number cannot be closed in the original's place; and the close
+    /// goes through the token's injectable closer, so a test can witness that
+    /// this site invoked it. An earlier version probed the descriptor with
+    /// fcntl AFTER closing — racy, because by then the number may belong to
+    /// someone else, and it failed under load on unchanged code.
     ///
-    /// The seam is what makes "did this site close anything?" answerable without
-    /// inspecting a number that no longer belongs to us. I had accepted that gap
-    /// as unclosable; review built the construction that closes it.
-    static func closeRejectedAdoption(_ fd: Int32, generation: Int, owner: AnyObject?) {
+    /// Both a success-shaped substitute for the call and deleting it outright
+    /// leave the injected closer uninvoked, so both are caught.
+    static func closeRejectedAdoption(
+        _ fd: Int32, generation: Int, owner: SSHTransport.AuditToken?
+    ) {
         let current = lock.withLock { generations[fd] ?? 0 }
         guard current == generation else { return }   // recycled: not ours to close
-        // NO POST-CLOSE PROBE. A previous version closed and then verified with
-        // fcntl(fd, F_GETFD) that the descriptor was gone. That check is
-        // ITSELF RACY: once close returns, the number is free and another
-        // thread can reuse it before the fcntl runs, whereupon a HEALTHY
-        // rejection goes unrecorded. Review measured it — unchanged code failed
-        // this regression under live full-suite load while passing 20/20 in
-        // isolation, which is the same load-dependent shape as the flake this
-        // whole PR set out to remove. I replaced one flaky observation with
-        // another and called it evidence.
-        //
-        // An unreliable guard is worse than a named gap. So the fact recorded
-        // is close's own return, which is the KERNEL'S answer about this
-        // descriptor at the moment it acted — the last point at which the
-        // number still means what the caller meant.
-        //
-        // KNOWN AND ACCEPTED: a source mutation replacing the syscall with a
-        // success-shaped expression is not caught here. That attack rewrites
-        // the audit's own body rather than the code under audit, and catching
-        // it in-process requires exactly the racy probe removed above. The
-        // adjacent mutation — deleting the call entirely — IS caught, because
-        // then nothing is recorded.
-        // THROUGH THE OWNER'S CLOSER. Both a success-shaped substitute for the
-        // syscall and outright deletion of the close leave the injected closer
-        // UNINVOKED, so a test that installs one and counts it catches either —
-        // without looking at the descriptor number after close, which is what
-        // made the previous attempt racy.
-        //
-        // (The comment that stood here claimed deletion would fail to compile.
-        // It does not: `if let owner {` is valid Swift, as review demonstrated.)
-        if closer(for: owner)(fd) == 0, let owner {
-            lock.lock(); rejectionsByOwner[ObjectIdentifier(owner), default: 0] += 1; lock.unlock()
+        if (owner?.closer ?? realClose)(fd) == 0 {
+            owner?.recordRejection()
         }
-    }
-    static func adoptionRejections(by owner: AnyObject) -> Int {
-        lock.lock(); defer { lock.unlock() }
-        return rejectionsByOwner[ObjectIdentifier(owner)] ?? 0
-    }
-
-    /// Drops an owner's entry once it is gone, so the map does not grow for the
-    /// life of the process.
-    static func forget(_ owner: AnyObject) {
-        lock.lock()
-        byOwner[ObjectIdentifier(owner)] = nil
-        rejectionsByOwner[ObjectIdentifier(owner)] = nil
-        closers[ObjectIdentifier(owner)] = nil
-        lock.unlock()
     }
 }
 
--- a/Sources/HerdrKit/SSHTransport.swift
+++ b/Sources/HerdrKit/SSHTransport.swift
@@ -561,11 +561,12 @@ public struct SSHTransport: HerdrTransport {
         for address in addresses {
             let fd = socket(address.family, address.socketType, address.protocolNumber)
             guard fd >= 0 else { lastErrno = errno; continue }
+            let generation = DescriptorAudit.opened(fd)
             afterDescriptorPublished?(live)
             // Published before the first call that can block on it, so an
             // interrupt arriving mid-connect has a real descriptor to shut down.
             guard live?.adopt(rawSocket: fd) ?? true else {
-                DescriptorAudit.closeRejectedAdoption(fd, owner: auditToken)
+                DescriptorAudit.closeRejectedAdoption(fd, generation: generation, owner: auditToken)
                 throw CancellationError()
             }
             do {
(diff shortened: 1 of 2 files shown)


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c80edf0bae9140`

**Summary:** The owner-scoped audit fixes the reported cross-test attribution problem and now detects the host-key-path mutation, but two descriptor-close branches still bypass the audit, and the new token breaks warnings-as-errors/Swift 6 Sendable checking.

**Findings**
- **Route the remaining tcpConnect closes through DescriptorAudit** (high)
  - body: The closes at lines 555 and 571 still call Glibc_close directly, so EBADF on either branch never reaches the process or owner tally. I duplicated the line-571 close; it compiled and all four connect/cancellation regressions passed. This leaves the audit partly blind in exactly the way this PR is intended to eliminate. Route both sites through DescriptorAudit.close(fd, owner: auditToken) and pin an error branch.
  - line: 555
  - path: Sources/HerdrKit/SSHTransport.swift
- **Make AuditToken Sendable** (medium)
  - body: SSHTransport conforms to HerdrTransport, which is Sendable, but its new AuditToken reference is not Sendable. Swift 6.1 emits a warning on every normal build, and `swift test -Xswiftc -warnings-as-errors` fails compilation. Declaring the empty immutable token as `final class AuditToken: [REDACTED] {}` made the same warnings-as-errors test build and pass.
  - line: 179
  - path: Sources/HerdrKit/SSHTransport.swift

**Tests Run**
- Code identity: reviewed pinned SHA 9d85622a274248ff38a29457f211e06e7825fd71 through local Git plumbing. Worktree HEAD was also 9d85622a274248ff38a29457f211e06e7825fd71; merge-base with origin/main was d7ca88c29cbd48a7402695ca2d0a44bf08726f7a. Worktree remained clean.
- Content discriminator: Sources/HerdrKit/SSHTransport.swift blob a1992d7af589866daeed9c5c0774a0c39935d2bb. DescriptorAudit.close now takes an owner via the verbatim line `static func close(_ fd: Int32, owner: AnyObject? = nil) {`. AuditToken is a reference inside the SSHTransport value type so copied transports retain one shared logical identity and their channels attribute closes to the same owner.
- Five former direct-close sites were routed through DescriptorAudit: Session.closeSocket plus four openSession failure exits. The host-key regression drives the verifyHostKey catch site whose verbatim close is `DescriptorAudit.close(sock, owner: auditToken)`. Tests/HerdrKitTests/SSHTransportTests.swift blob was 55bd572a034da41f72858dd3034d6cdcf6eef2a6.
- Deliberate host-key-path double close: with the new audited close, the compiling probe was KILLED; the test reported 25 owner-attributed double closes against baseline 0. Replacing both closes with the parent commit's direct Glibc_close behavior made the same deliberate double close SURVIVE with the focused test passing.
- Foreign-owner mechanism probe: injected an unrelated audited double close inside the host-key test's measurement window. The process-wide tally increased while this transport's owner tally stayed unchanged, and the focused test passed. This demonstrates the cross-test attribution mechanism directly.
- Remaining-close probe: duplicated the direct Glibc_close at tcpConnect's failure cleanup. `testCancellationInsideTheClassificationWindowIsStillCancellation`, `testCancellationInterruptsConnectAndReportsItAsCancellation`, `testConnectIsBoundedBySetupTimeout`, and `testStalledConnectsDoNotExhaustTheBlockingQueue` all passed, demonstrating that branch remains unaudited.
- Sendable probe: exact head failed `swift test -Xswiftc -warnings-as-errors --filter SSHTransportTests/testHostKeyRejectionThroughStreamNeverClosesADescriptorTwice` because AuditToken is non-Sendable. Changing only AuditToken to conform to Sendable made the build and focused live test pass.
- Exact restored head full suite in a Swift 6.1.3 container with host sshd, key, and Herdr socket: 142 tests executed, 2 canary tests skipped because another thread took the recycled descriptor, 0 failures. The normal build emitted the AuditToken Sendable warning.
- Restoration verification: archived SSHTransport.swift and SSHTransportTests.swift hashed back to their pinned blobs a1992d7af589866daeed9c5c0774a0c39935d2bb and 55bd572a034da41f72858dd3034d6cdcf6eef2a6 after every probe; assigned worktree remained clean.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c8102bbcd4156b`

**Summary:** The catch cleanup is now genuinely pinned, but the audited adopt-rejection cleanup remains unverified: a compiling deliberate double close there survived the complete SSH suite.

**Findings**
- SSHTransport.swift:555 — Add a deterministic test for the `live?.adopt(rawSocket: fd) == false` branch. Duplicating `DescriptorAudit.close(fd, owner: auditToken)` there compiled successfully and all 23 SSH tests still passed. Staging this requires a barrier or injectable adoption seam that interrupts a LiveChannel after `socket()` creates the descriptor but before `adopt(rawSocket:)`; otherwise cancellation exits earlier or reaches the already-pinned catch cleanup.

**Tests Run**
- Reviewed pinned commit c25c3dfa5f39da883899f4bc1cada27c7fe0ba0b through `git cat-file`, `git diff`, and `git show`; assigned worktree HEAD was stale at 9d85622a274248ff38a29457f211e06e7825fd71. Worktree remained clean.
- Content discriminator: SSHTransport.swift blob 749b23851f36e8ee5b3370a47b37f8439660e9da. One raw operational `Glibc_close` call remains, inside `DescriptorAudit.close`: `guard Glibc_close(fd) != 0, errno == EBADF else { return }`.
- Content discriminator: the newly pinned branch is tcpConnect's catch cleanup after `connectCandidate` throws. `SSHTransportTests.testCancellationInterruptsConnectAndReportsItAsCancellation` pins it by comparing the transport-owned double-close tally with its baseline.
- Content discriminator: the audited but unpinned branch is failed `LiveChannel.adopt(rawSocket:)`. It requires deterministic cancellation after descriptor creation and before adoption.
- Ran the warnings-as-errors SSH suite 10 captured times: 23 tests per run, zero failures; the previously unexplained failure did not reproduce.
- Catch-cleanup double-close probe: mutant compiled with warnings as errors and the focused cancellation test failed on audit tally 1 versus baseline 0 — KILLED.
- Adopt-rejection double-close probe: mutant compiled with warnings as errors and the complete 23-test SSH suite passed — SURVIVED.
- Ran `swift test -Xswiftc -warnings-as-errors` on the restored pinned archive: 142 tests executed, one canary skipped because another thread acquired the freed descriptor, zero failures.

**Needs**
- Add a deterministic regression that drives and audits the failed-adoption cleanup branch.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c811dbdd237000`

**Summary:** The double-close mutant is killed, but the regression still does not prove the adopt-rejection branch ran. Bypassing that branch entirely compiles and leaves the named test green.

**Findings**
- SSHTransportTests.swift:527 — `reached.isSet` proves only that `afterDescriptorPublished` fired, not that the rejection branch at SSHTransport.swift:567 executed. Replacing the entire guard and rejection block with `_ = live?.adopt(rawSocket: fd)` compiled under warnings-as-errors and `testAdoptionRejectionClosesItsDescriptorExactlyOnce` still passed: the later `openSession` interruption cleanup closed the descriptor once, while the test swallowed its error and observed the same tally. Add branch-specific positive evidence before comparing tallies, and make this branch-bypass mutation fail.

**Tests Run**
- Reviewed pinned commit a6d8a621d8b4bd9461bf7bfa1dde0e187b205de0 through local Git plumbing; assigned worktree HEAD was stale at 9d85622a274248ff38a29457f211e06e7825fd71. The shared worktree remained clean.
- Content discriminator: SSHTransport.swift blob 923daee0f08727b1cd6f887b82e396b583b0a2bc. `afterDescriptorPublished` is invoked at line 564 after `socket()` returns a valid descriptor and immediately before `live?.adopt(rawSocket:)`. Public cancellation before this window exits at resolution; cancellation afterward reaches the existing catch cleanup, so no public-call ordering deterministically selects this branch.
- Content discriminator: test blob eeae4a6826d8eb176129d8bf749187bb7cd0ffd9. Before reading the tally, the regression asserts `XCTAssertTrue(reached.isSet,`; this proves the seam fired but, as the surviving branch-bypass probe demonstrates, not that adoption rejection performed the cleanup.
- Content discriminator: one raw operational `Glibc_close` call remains in SSHTransport.swift, inside `DescriptorAudit.close`: `guard Glibc_close(fd) != 0, errno == EBADF else { return }`.
- Baseline focused regression compiled with warnings-as-errors and passed.
- Premise probe removing `live.interrupt()` compiled and made the regression fail because the stream delivered, confirming the hook normally arms cancellation.
- Double-close probe duplicated the rejection branch's audited close; it compiled with warnings-as-errors and the named regression failed with tally 1 versus 0 — KILLED.
- Branch-premise probe removed the rejection guard and ignored `adopt`'s result; it compiled with warnings-as-errors and the named regression passed — SURVIVED.
- Restored and hash-verified both pinned blobs, then ran `swift test -Xswiftc -warnings-as-errors` on Linux: 143 tests executed, one descriptor-canary test skipped because another thread acquired the freed descriptor, zero failures.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c813b07dee54ec`

**Summary:** The prior branch-bypass mutation is now killed, but the new counter is not coupled to the cleanup it claims to witness: deleting the sole close compiles and leaves all SSH tests green while leaking the descriptor.

**Findings**
- SSHTransport.swift:568-569 — `recordAdoptionRejection` and `DescriptorAudit.close` are independent operations. Deleting only `DescriptorAudit.close(fd, owner: auditToken)` compiled under warnings-as-errors; the named regression and the complete 24-test SSH suite still passed because they observed one rejection and zero double closes. That state represents a leaked descriptor, not a correct exactly-once close. Make the positive observable originate from a successful audited close—for example, record an attributed `.adoptionRejection` only when the close succeeds—and assert that success count before asserting no double close.

**Tests Run**
- Reviewed pinned SHA e389035e9f02cabcf2b260a12f145de87439afae through local Git plumbing and an archived tree. Assigned worktree HEAD was stale at 9d85622a274248ff38a29457f211e06e7825fd71; the shared worktree remained clean.
- Content discriminator: SSHTransport.swift blob 3965244a07b1025cc602b15c7ec0ef4c9b91134e and SSHTransportTests.swift blob ca7b0218ec2cac7cfc364ae26b9047c86b41601d. Before comparing double-close tallies, the regression asserts the seam flag and then `DescriptorAudit.adoptionRejections(by: transport.auditToken), 1,`. That value comes from `DescriptorAudit.recordAdoptionRejection(owner: auditToken)` inside the failed `live?.adopt(rawSocket:)` branch.
- Content discriminator: the prior bypass mutation replacing the guard with `_ = live?.adopt(rawSocket: fd)` compiled under warnings-as-errors and is now KILLED; the regression failed with adoption-rejection count 0 versus 1.
- Close-without-record probe: deleting `recordAdoptionRejection` compiled and the focused regression failed with count 0 versus 1 — KILLED.
- Record-without-close probe: deleting `DescriptorAudit.close(fd, owner: auditToken)` compiled and the focused regression passed — SURVIVED.
- Ran the complete warnings-as-errors SSH suite with the record-without-close mutant: 24 tests, zero failures — SURVIVED, confirming no broader descriptor test catches the leak.
- Double-close behavior remains pinned by the existing tally assertion; the round-three duplicate-close mutant still fails that comparison.
- Seam blast inspection: `afterDescriptorPublished` is internal despite SSHTransport being public, defaults to nil, is assigned only by this local regression, and no configured transport copy escapes the test method. No cross-test perturbation path was found.
- Restored and hash-verified both pinned blobs, then ran `swift test -Xswiftc -warnings-as-errors`: 143 tests executed, two descriptor-canary tests explicitly skipped because another thread acquired the freed descriptor, zero failures.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c814729531631b`

**Summary:** The adjacent call-site mutations are killed, but the new audit still accepts a leaked descriptor or a recycled descriptor as a successful rejection. The full suite also has a reproducible expired-deadline test race.

**Findings**
- [high] Sources/HerdrKit/SSHTransport.swift:1179-1189 still trusts the close expression rather than proving the original descriptor was closed. Replacing `Glibc_close(fd)` with the build-valid success expression `(fd >= 0 ? 0 : -1)` left the focused regression green while leaking `fd`. A second built probe first closed `fd`, opened another socket that a precondition confirmed reused the same descriptor number, then called `closeRejectedAdoption`; the regression again passed while the helper closed the unrelated socket. `EBADF` distinguishes only the descriptor number's current state, not its ownership or generation. The audit needs an observable at the close-wrapper boundary that identifies the original descriptor operation, rather than inferring it from close's result.
- [medium] Tests/HerdrKitTests/SSHTransportTests.swift:686 has a vacuous timing precondition. It creates a localhost resolution and immediately claims it with an expired deadline, but `Resolution.claim` checks the deadline only while `finished == false`; if the resolver thread finishes first, `.resolved` is correct and the test fails. This occurred on suite repeat 18, once in 200 isolated processes, and again in the final full-suite run. Hold `Resolution.startGate` until the expired claim has been exercised, so the test positively proves the resolver is pending. This test predates the PR, but it explains an unaccounted-for gate failure and currently leaves the head's full suite red.

**Tests Run**
- Reviewed pinned SHA 64ce5725f3b81104e0321e514c5d0f7fefa04dc5 through git cat-file, git diff, git show, and a private git archive. Shared worktree HEAD was 9d85622a274248ff38a29457f211e06e7825fd71 and remained clean; it was stale by the full PR series.
- Content discriminator: Sources/HerdrKit/SSHTransport.swift blob 488ca53d5ba85339e48257e73b8b8bbdfca1fa45. The single replacement operation is `DescriptorAudit.closeRejectedAdoption(fd, owner: auditToken)`. Its helper's verbatim success line is `if Glibc_close(fd) == 0 {`; it records the attributed adoption rejection only on that success branch.
- Content discriminator mutations: deleting the complete `closeRejectedAdoption` call built and failed the named regression because adoption rejections stayed at 0; this represents the leaked-descriptor state. Duplicating `Glibc_close(fd)` immediately before the helper built and failed because rejections stayed at 0 and attributed double closes rose from 0 to 1; this represents the double-close state.
- Additional premise mutation: replacing only `Glibc_close(fd)` with the runtime success expression `(fd >= 0 ? 0 : -1)` built under warnings-as-errors and SURVIVED; the descriptor leaked while the test observed one rejection and zero double closes.
- Additional reuse mutation: pre-closing `fd`, opening a socket that was asserted to reuse exactly `fd`, then calling the helper built and SURVIVED; the helper closed the recycled socket, reported one rejection, and reported no double close.
- Baseline `swift test -Xswiftc -warnings-as-errors --filter SSHTransportTests`: 24 tests, 0 failures.
- SSHTransportTests repeat stress: runs 1-17 completed without a signal; run 18 failed `testResolutionGivesUpOnAnExpiredDeadline`, not with a crash. The disclosed SIGSEGV was not reproduced.
- Focused host-key rejection regression: 50 processes, 0 failures and 0 signals. Focused adoption regression: 200 processes, 0 failures on unmodified code.
- Expired-deadline regression: 1 failure in 200 isolated processes. Final `swift test -Xswiftc -warnings-as-errors`: 143 tests executed, 2 skipped, 1 failure in `testResolutionGivesUpOnAnExpiredDeadline`.
- After every probe, the archived SSHTransport.swift SHA-256 matched the pinned blob's content; the shared worktree remained untouched and clean.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c815e0b65a7262`

**Summary:** The two round-five mutants are killed, but the replacement post-close proof is itself load-dependent: unchanged 8462fbd failed its adoption-rejection regression during the live full suite. A descriptor number can be reused between close and F_GETFD, so this PR replaces one flaky process-wide observation with another.

**Findings**
- [high] Sources/HerdrKit/SSHTransport.swift:1209 — `closeRejectedAdoption` closes the fd and only afterward calls `fcntl(fd, F_GETFD)`. Once close returns, another thread may immediately reuse that number; F_GETFD then succeeds and the healthy rejection is not recorded. The generation check cannot close this window because arbitrary descriptor opens do not update its map. On unchanged 8462fbd, the live full suite reached the seam but failed `testAdoptionRejectionClosesItsDescriptorExactlyOnce` with adoptionRejections 0 instead of 1. The same test passed 20/20 in isolation and the SSH suite passed 10/10, demonstrating the load-dependent character. The observable must remain coupled to the original descriptor rather than probing its reusable number after close.

**Tests Run**
- Reviewed pinned commit 8462fbd67280301966ce97add474d7b28a51a02f through Git plumbing and a private archive. Shared worktree HEAD was 9d85622a274248ff38a29457f211e06e7825fd71, five commits stale.
- Content discriminator — ENGAGED. SSHTransport.swift blob: 4ebd4e805252a6ca3674f603b34d58d9f644098c. `closeRejectedAdoption` verifies `guard current == generation else { return }` and then verifies the descriptor number reports gone with `guard fcntl(fd, F_GETFD) == -1, errno == EBADF else { return }`.
- Content discriminator — ENGAGED. SSHTransportTests.swift blob: 3d777c139b2b22a63230329ae803364e419fcbbc. The expired-deadline test holds the resolver pending with `let holding = DispatchSemaphore(value: 0)` and `SSHTransport.Resolution.startGate = { holding.wait() }`. This is necessary because 127.0.0.1 resolves immediately; without a deterministic hold it can finish before claim, making `.resolved` correct rather than `.gaveUp`.
- Content discriminator — ENGAGED. Replaced `Glibc_close(fd)` with the build-valid success expression: mutation built under warnings-as-errors and `testAdoptionRejectionClosesItsDescriptorExactlyOnce` failed, so KILLED. Removed the generation comparison: mutation built and `testARecycledDescriptorIsNotClosedInPlaceOfTheOriginal` failed both ownership assertions, so KILLED.
- Restored both probed files byte-for-byte: archive hashes matched the pinned source and test blob IDs. The three discriminator tests then passed together: 3 tests, 0 failures.
- Exact archived commit in a Swift 6.1 container: `swift test -Xswiftc -warnings-as-errors` passed 144 tests with 22 environment-dependent skips and 0 failures; live SSH paths were skipped there.
- Exact archived commit on the live host: full suite executed 144 tests and reproduced the adoption-rejection assertion failure on unchanged code. A second unrelated failure came from the sandbox's read-only default Clang module cache.
- Unchanged adoption regression: 20 isolated runs, 0 failures. Unchanged complete SSH suite: 10 runs, 0 failures. This contrast with the live full-suite failure demonstrates cross-suite descriptor-churn sensitivity.
- SIGSEGV probe: host-key rejection regression ran 20 times with 0 failures or signals; the prior signal 11 was not reproduced. No signal occurred in the 10 SSH-suite runs or live full-suite attempt either.
- Premise probe: removing the resolver startGate produced 0 failures in 30 isolated runs, confirming why timing alone did not reliably expose the old race; restored test passed with the deterministic semaphore gate.
- Ran `git diff --check` for the complete PR; passed. Shared worktree remained clean.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c817787b962b3f`

**Summary:** The two accepted mutation gaps are catchable without reintroducing the racy post-close descriptor probe. A build-valid proof of concept used an owner-scoped @Sendable close closure, counted its invocation, and delegated to the real close; it killed both admitted survivors. The current regression can therefore still describe a leaked descriptor as a successful rejection.

**Findings**
- [high] Sources/HerdrKit/SSHTransport.swift:1236 — Add a narrow close dependency seam and assert its invocation. On the pinned head, replacing Glibc_close with a success-shaped expression and deleting the close while retaining `if let owner` both built and survived. A private proof of concept using an owner-scoped @Sendable closer built, passed unchanged, and killed both mutations without inspecting the descriptor number after close. The source comment also incorrectly claims close deletion becomes INVALID; the demonstrated `if let owner` mutation is build-valid.
- [low] Sources/HerdrKit/SSHTransport.swift:1196 — The documentation still says closeRejectedAdoption verifies the descriptor is gone using F_GETFD, although that check was removed as racy. Update the stated invariant to match the implementation.

**Tests Run**
- Reviewed pinned SHA d0e7aac2e97ce49c96a17f67450804a990c5cf8b through local Git plumbing; shared worktree HEAD was 9d85622a274248ff38a29457f211e06e7825fd71. Merge base with origin/main: d7ca88c29cbd48a7402695ca2d0a44bf08726f7a.
- Content discriminator: Sources/HerdrKit/SSHTransport.swift blob 3d5ec13613e5aec22ac47665afc0b4b44bf05140; central line verbatim: `if Glibc_close(fd) == 0, let owner {`. Tests/HerdrKitTests/SSHTransportTests.swift blob: 3d777c139b2b22a63230329ae803364e419fcbbc.
- Removed from closeRejectedAdoption: the post-close `fcntl(fd, F_GETFD)`/EBADF check. It was unsound because close frees the descriptor number, allowing another thread to reuse it before fcntl and making healthy behavior appear unclosed.
- The two explicitly accepted gaps attack the audit body itself: substituting `(fd >= 0 ? 0 : -1)` for Glibc_close preserves a success result without closing; changing the condition to `if let owner` deletes the close while leaving rejection recording reachable. Both mutants built under -warnings-as-errors and the focused regression passed: SURVIVED.
- The remaining generation check is race-free with respect to descriptor reuse because `let current = lock.withLock { generations[fd] ?? 0 }` and `guard current == generation else { return }` run before close, while the original descriptor still owns that number; the lock serializes the generation map. Removing this guard built and caused testARecycledDescriptorIsNotClosedInPlaceOfTheOriginal to fail: KILLED.
- Race-free proof of concept: injected an owner-scoped `@Sendable (Int32) -> Int32` closer, used a Locked invocation counter, and delegated to the real close. Baseline built and passed; both the success-expression and close-deletion mutants built and failed the invocation assertion: KILLED. An initial non-Sendable function-reference attempt did not build and was classified INVALID before correction.
- Exact-head full suite: `swift test -Xswiftc -warnings-as-errors` — 144 tests executed, 2 skipped, 0 failures; no signal 11 observed in this run.
- `git diff --check` passed. All probe changes were restored byte-for-byte to the pinned blobs, and the shared worktree remained clean.

**Needs**
- Replace the accepted untestable-gap claim with a race-free close-invocation seam and regression assertion, or an equivalent construction that kills both build-valid mutations; correct the stale F_GETFD and deletion-invalid documentation.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c8184f04cfff86`

**Summary:** The two formerly accepted mutations are now genuinely killed, but the owner-scoped closer is only address-scoped unless every caller manually invokes forget(). A deterministic probe showed a replacement owner inheriting and executing a retired owner's stale closer.

**Findings**
- [high] Sources/HerdrKit/SSHTransport.swift:1191 — `closers` is keyed only by ObjectIdentifier. I installed a closer, released its AuditToken without calling forget(), obtained a replacement token at the same address, and invoked closeRejectedAdoption for the replacement; the retired closer ran. This can make later tests intercept or suppress a real close. Store the closer on AuditToken itself, or otherwise bind it to a lifetime-safe identity rather than requiring manual cleanup.
- [low] Sources/HerdrKit/SSHTransport.swift:1242 — The old implementation comment remains and says the success-shaped substitution is uncaught and deleting the call is caught. Both claims directly contradict the new seam and measured mutation results. The doc comment was corrected, but this adjacent block was not.

**Tests Run**
- Reviewed pinned SHA bd49e90d5be4729ac4a10b4a56a9271851994afc through local Git plumbing; shared worktree HEAD was 9d85622a274248ff38a29457f211e06e7825fd71. Merge base with origin/main: d7ca88c29cbd48a7402695ca2d0a44bf08726f7a.
- Content discriminator: Sources/HerdrKit/SSHTransport.swift blob 771e5f6644694c7624b4c0f7c11e71a5a1f862e9; central line verbatim: `if closer(for: owner)(fd) == 0, let owner {`. The adoption-rejection close now goes through an owner-selected injectable closer, while the regression's UncheckedCounter counts its invocations and requires `closes.value == 1`. Tests blob: bd7dc5278624c7cba18c5253f253d56722be12fd.
- Success-shaped substitution `(fd >= 0 ? 0 : -1)`: built under -warnings-as-errors; the focused regression failed because the invocation count was 0: KILLED.
- Close deletion expressed as `if let owner {`: built under -warnings-as-errors; the focused regression failed because the invocation count was 0: KILLED.
- The corrected doc comment replaces the former claims that F_GETFD proved the descriptor was gone and that close deletion could not compile. A stale non-doc comment at lines 1242-1247 still makes the old claims.
- Closer-lifetime premise probe: dropped an AuditToken after installing a closer without forget(), confirmed deallocation, allocated a replacement with the same ObjectIdentifier, and called closeRejectedAdoption for it. The stale closer ran once; the safety assertion expected zero and failed. The probe built cleanly and used XCTUnwrap rather than force-unwrapping.
- Generation-map review: entries are never pruned, but memory grows with the highest set of distinct descriptor numbers observed, not with every reopen; repeated opens update an existing Int value. The current process limit was 1024 descriptors. I found no merge-blocking growth mechanism here.
- Signal-11 stress: SSHTransportTests ran 10 times under -warnings-as-errors, 25 tests per run, 250 executions total, zero failures and no crash; one run had one staged-hazard skip.
- Exact-head full suite under -warnings-as-errors: 144 tests, 3 skipped, 0 failures; no signal 11. This measured total is 144, not the dispatch's stated 143.
- `git diff --check` passed. Both probed files were restored to their pinned blob IDs, temporary archives/logs were removed, and the shared worktree remained clean.

**Needs**
- Make closer lifetime follow AuditToken lifetime so ObjectIdentifier reuse cannot inherit executable test state, add the demonstrated reuse regression, and remove the contradictory stale comment.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c81936cd42dfd5`

**Summary:** The production refactor closes the address-reuse hazard, but the new regression violates the lane's required warnings-as-errors build: its @Sendable closer mutates a captured local Int. The ordinary suite passes only while accepting that warning.

**Findings**
- **The new regression does not compile with warnings treated as errors** (high)
  - body: `retired.closer = { _ in intercepted += 1; return 0 }` mutates captured local state from an `@Sendable` closure. `swift test -Xswiftc -warnings-as-errors --filter SSHTransportTests` exits 1 during compilation with `mutation of captured var 'intercepted' in concurrently-executing code`. The file already provides `UncheckedCounter`; temporarily using it made the required build pass and preserved the regression. This must be fixed before approval.
  - file: Tests/HerdrKitTests/SSHTransportTests.swift
  - line: 523

**Tests Run**
- Reviewed pinned SHA 23958bb754b5460d361f1bf2499cc65868304337 through an isolated `git archive`; shared worktree HEAD was 9d85622a274248ff38a29457f211e06e7825fd71. The shared worktree remained clean.
- Content discriminator: `Sources/HerdrKit/SSHTransport.swift` blob 450e923a635cceab464580aa5f34b3fdb62928bc. The closer and counters now live directly on `AuditToken`—verbatim: `final class AuditToken: [REDACTED] Sendable {`, `private var _closer: @Sendable (Int32) -> Int32 = { DescriptorAudit.realClose($0) }`, `private var _doubleCloses = 0`, and `private var _rejections = 0`. The ObjectIdentifier-keyed `byOwner`, `rejectionsByOwner`, and `closers` maps and `forget()` no longer exist.
- Content discriminator: `Tests/HerdrKitTests/SSHTransportTests.swift` blob 0e48d6a6df5983a5dfbee7b95f20a2014f4a2c51. If address reuse is not obtained, the regression skips rather than passes—verbatim: `try XCTSkipUnless(addresses[0] == addresses[1],` with message `the allocator did not reuse the address; the hazard is not staged`. The allocator reused the address and the regression passed in 10/10 focused runs.
- Content discriminator: the stale parent comment beginning `// KNOWN AND ACCEPTED: a source mutation replacing the syscall with a` and claiming deletion was caught while substitution was not has been removed. The current source instead states verbatim: `Both a success-shaped substitute for the call and deleting it outright` and `leave the injected closer uninvoked, so both are caught.`
- Required build: `swift test -Xswiftc -warnings-as-errors --filter SSHTransportTests` on Swift 6.1.3 failed to compile at SSHTransportTests.swift:523 due to mutation of captured `intercepted`. This is a build failure, not a test result.
- Ordinary `swift test`: 145 tests executed, 21 skipped, 0 failures; compilation emitted the same concurrent-capture warning. The observed total is 145, not the dispatched 144.
- Temporary premise repair using the existing locked `UncheckedCounter`: warnings-as-errors build passed and the address-reuse regression passed. The temporary change was restored.
- With only that temporary compile repair, success-shaped close substitution: VALID build, KILLED by `testAdoptionRejectionClosesItsDescriptorExactlyOnce` because the injected closer recorded 0 calls.
- With only that temporary compile repair, generation-check removal: VALID build, KILLED by `testARecycledDescriptorIsNotClosedInPlaceOfTheOriginal`; it detected both stale-generation acceptance and closure of the recycled descriptor.
- Generation-map review: keys are descriptor numbers and this runner's open-file ceiling is 1024, so growth is bounded by the descriptor-number domain rather than elapsed time.

**Needs**
- Replace the captured mutable `Int` in `testAReplacementTokenDoesNotInheritARetiredTokensCloser` with thread-safe state such as the existing `UncheckedCounter`, then rerun the warnings-as-errors suite.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c81a28b3521713`

**Decision:** `approved`
**Summary:** No blocking issues remain. The warnings-as-errors fix is correct, the generation map is bounded by descriptor-number high-water rather than elapsed time, and repeated live SSH runs did not reproduce the historical signal 11. This PR is done.

**Tests Run**
- Reviewed pinned SHA 7a606c18eca65308713875e749752509ebb80bae through an isolated archive. Shared worktree HEAD was 9d85622a274248ff38a29457f211e06e7825fd71 and remained clean.
- Content discriminator: Tests/HerdrKitTests/SSHTransportTests.swift blob 4b975f3b7756c4386069eedffabee1aca805e3bd. The captured `var intercepted` was replaced by verbatim `let intercepted = UncheckedCounter()`, and the closer now calls `intercepted.bump()`. `UncheckedCounter` already existed at lines 16–20 of the same file as the locked helper previously written for another @Sendable seam.
- Content discriminator: full `swift test -Xswiftc -warnings-as-errors` executed 145 tests with 0 failures. Two allocator-dependent canaries skipped honestly; live herdr and SSH prerequisites were available.
- Signal-11 probe: 10 consecutive live `SSHTransportTests` runs, 26 tests each, completed with 0 failures and no signals—260 focused test executions.
- Generation-map inspection: Sources/HerdrKit/SSHTransport.swift blob 450e923a635cceab464580aa5f34b3fdb62928bc. The map is `[Int32: Int]`, guarded by one NSLock, and `opened` increments the value for the descriptor-number key rather than appending per opening.
- Generation-map measurement probe: after 10,000 sequential open/record/close cycles, the temporary accessor reported `before=0 after=1 distinct=1 opens=10000`. The host soft descriptor limit was 1,024 and hard ceiling 1,048,576, confirming growth follows distinct descriptor numbers/high-water rather than runtime duration.
- The temporary generation-map accessor and probe were restored file-by-file before reporting.

PR #19, opened 2026-08-02
tests: sweep the two attempt fallbacks, and fix what they were hiding

jerryfane:
Follow-up to #18, directed by JARVIS in directive 33095: sweep the two remaining test-only `state.currentAttempt ?? AttemptID(uuid: UUID())` fallbacks as a separate PR rather than widening #18.

## The sweep is small; what it uncovered is not

Both become `try XCTUnwrap(...)` and both tests still pass — the `??` branch was dead in both, so the round-15 reviewer's reading was correct.

Removing the mask made the sites readable, and both were **mislabelled**:

- `State.connectedSince` and `State.currentAttempt` are **not value-stored** — both forward to the shared authority (`SessionRecovery.swift:474,485`). `state = saved` cannot replay either into the gate.
- Both assertions therefore send from the **new, current, not-yet-adopted** attempt and are rejected by the *connection* check. Neither pins the abandoned-transport path its message claimed.

## Established by distribution, not by argument

Mutating `admitWhileConnected`'s two sub-checks against the whole suite partitions it cleanly:

| mutation | killed by |
|---|---|
| drop-provenance | `testADelayedSnapshotFromAnAbandonedAttemptAdmitsNothing`, `testALatePaneEventFromAnAbandonedAttemptActsOnNothing` |
| drop-connection | `testACopySavedWhileConnectedCannotPoisonTheReplacementAttempt`, `testBeginInitialAttemptWhileConnectedCancelsAndClearsAdoption` |

No test straddles. **The stale-attempt path was never uncovered** — it is pinned by tests that say so in their names. This is a labelling defect, not a coverage hole, and I am not claiming otherwise.

## Changes

- Both sites send from a **named** attempt (`minted`, `attemptB`) rather than reading `currentAttempt` back out of the state, plus an assertion that it is the expected one. Reading it back lets a transition silently retarget the test.
- Messages corrected to the invariant actually pinned.
- A **positive control** at the first site: adopt, re-send the same event, require the subscribe. Deleting the adoption KILLS it — armed, with adoption as its premise.

## Residual, measured and not papered over

Substituting a stranger `AttemptID` into either negative assertion still **SURVIVES**: provenance rejects a stranger to the identical observable, and the control reads its attempt independently, so nothing binds the two lines to one attempt. The negative lines are not attempt-discriminating.

I wrote a comment claiming the control fixed that, measured that it did not, and rewrote the comment. Both limitations are stated at the sites.

**If a construction binds them, I have not found it — that is the thing I most want challenged.**

## Evidence

146 tests, 0 failures, `-Xswiftc -warnings-as-errors`. Skips were 2 this run vs 3 previously; the SSH gates are descriptor/fixture-availability dependent and vary run to run, not from this change. Linux only — the macOS buildbox verdict follows the push.

### DIFF
--- a/Tests/HerdrKitTests/SessionRecoveryTests.swift
+++ b/Tests/HerdrKitTests/SessionRecoveryTests.swift
@@ -172,13 +172,12 @@ final class SessionRecoveryTests: XCTestCase {
         // Deleting the adoption below leaves no subscribe: KILLED, so this
         // control is armed and adoption is its premise.
         //
-        // What it does NOT fix, measured after adding it rather than assumed:
-        // substituting a stranger AttemptID for `minted` in the assertion ABOVE
-        // still SURVIVES. Provenance rejects a stranger to the identical
-        // observable, and the control below reads `minted` independently, so
-        // nothing binds the two lines to one attempt. The negative line is
-        // therefore not attempt-discriminating; attempt discrimination is
-        // pinned by testALatePaneEventFromAnAbandonedAttemptActsOnNothing.
+        // ATTEMPT DISCRIMINATION comes from the shared `probe` value, not from
+        // this control. While the two deliveries were constructed separately, a
+        // stranger AttemptID substituted into the negative line alone SURVIVED
+        // — provenance rejects a stranger to the identical observable. Sharing
+        // one immutable event makes that substitution reach here too, where the
+        // subscribe then fails to appear: KILLED.
         _ = plan(.connected(minted, at: Date()), &state)
         XCTAssertEqual(plan(probe, &state).subscribes, ["p9"],
                        "the same event still admits nothing after adoption; the emptiness above was "
@@ -386,17 +385,13 @@ final class SessionRecoveryTests: XCTestCase {
         let probe = ClientEvent.paneCreated("adoption-probe", from: attemptB)
         XCTAssertTrue(plan(probe, &state).isEmpty,
                       "an event from B before B's adoption is processed admits nothing")
-        // MEASURED LIMITATION, stated rather than implied: substituting a
-        // stranger AttemptID for `attemptB` here leaves this assertion true
-        // (SURVIVED), because provenance rejects a stranger to the same
-        // observable. So this line pins "nothing leaks before adoption" and
-        // does NOT discriminate which attempt was refused. A positive control
-        // of the kind used in
-        // testBeginInitialAttemptWhileConnectedCancelsAndClearsAdoption does
-        // not fit here without consuming p9's freshness and gutting the
-        // mid-replay-learning assertion at the end of this test, which is the
-        // stronger check. Attempt discrimination is pinned by
-        // testALatePaneEventFromAnAbandonedAttemptActsOnNothing.
+        // ORDERING IS LOAD-BEARING HERE and is pinned: moving this delivery
+        // after the adoption below KILLS (it subscribes instead of staying
+        // empty), and moving the post-adoption delivery before it KILLS (no
+        // subscription appears). The post-adoption delivery and the snapshot
+        // below ARE interchangeable — both follow adoption and touch distinct
+        // panes — which is independence, not redundancy: the stranger mutation
+        // pins the probe and the p9 collision pins the snapshot.
 
         // And B's adoption must not be suppressed as a repeat.
         let adoptedB = plan(.connected(attemptB, at: Date()), &state)
--- a/Tests/HerdrKitTests/SessionRecoveryTests.swift
+++ b/Tests/HerdrKitTests/SessionRecoveryTests.swift
@@ -147,7 +147,15 @@ final class SessionRecoveryTests: XCTestCase {
         let minted = try XCTUnwrap(restarted.reconnectAttempt)
         XCTAssertEqual(state.currentAttempt, minted,
                        "the restart did not become current; the event below is not from the attempt under test")
-        XCTAssertTrue(plan(.paneCreated("p9", from: minted), &state).isEmpty,
+        // ONE EVENT VALUE, DELIVERED TWICE — the reviewer's construction, and the
+        // thing that makes the negative line attempt-discriminating. Built
+        // separately, the before and after events shared nothing, so swapping a
+        // stranger into the negative line alone left it empty (provenance
+        // rejects a stranger to the identical observable) and SURVIVED. Sharing
+        // the value means a stranger substituted here must also be a stranger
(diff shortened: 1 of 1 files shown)


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c81f73298edaf4`

**Summary:** The two negative assertions remain non-attempt-discriminating, but a simple race-free construction closes both survivors without weakening the later p9 snapshot check.

**Findings**
- **Bind each negative delivery to its positive control** (medium)
  - detail: Both negative assertions construct their paneCreated event separately from the evidence that the named attempt can admit it. Consequently, replacing the negative line's attempt with a stranger still passes via the provenance gate. Construct one ClientEvent value and deliver that same value before and after adoption. I verified this at site 1: the baseline construction passes, while substituting a stranger in the shared event initializer builds and is KILLED by the post-adoption assertion. Site 2 supports the same construction without consuming p9: use a distinct pane such as adoption-probe for the shared before/after event, then retain p9 exclusively for the authoritative snapshot. That baseline passes, preserves the existing [p1, p9] snapshot expectation, and its stranger mutation builds and is KILLED.
  - file: Tests/HerdrKitTests/SessionRecoveryTests.swift
  - line: 150

**Tests Run**
- Reviewed pinned SHA c85189e81e87671fa33cc9e0c351daacac8fd650 through local Git plumbing; worktree HEAD found was the same SHA. The worktree remained clean.
- Content discriminator: Tests/HerdrKitTests/SessionRecoveryTests.swift blob ee634b9eff04c3366cf48b9cc3ac4fdb34e3620e. Verbatim central line: `XCTAssertTrue(plan(.paneCreated("p9", from: minted), &state).isEmpty,`. The second affected line uses `attemptB` in the same independently-constructed shape.
- Site 1 probe: constructed one paneCreated event from minted and reused it before and after adoption. It built under -Xswiftc -warnings-as-errors and passed. Replacing the shared event's attempt with a stranger also built and failed at the post-adoption subscribe assertion. KILLED.
- Site 2 probe: constructed one adoption-probe event from attemptB, reused it before and after adoption, and left p9 for the later snapshot. It built under -Xswiftc -warnings-as-errors; both the focused test and its original [p1, p9] snapshot assertion passed. Replacing the shared event's attempt with a stranger built and failed at the post-adoption subscribe assertion. KILLED.
- Clean pinned archive: `swift test -Xswiftc -warnings-as-errors` executed 146 tests with 0 failures and 20 environment-gated skips on this image.
- No macOS build was run; the sender also reported no buildbox verdict for this head.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c81ff67ca2ed69`

**Summary:** The shared-event implementation is sound and its ordering is properly armed. One finding remains: both tests retain stale comments claiming the stranger mutations survive, directly contradicting this head's fix and evidence.

**Findings**
- **Remove the two obsolete survivor explanations** (low)
  - detail: Lines 175-181 still say the site-1 stranger substitution SURVIVES and that nothing binds the deliveries, although both now use `probe`. Lines 389-399 repeat the obsolete site-2 limitation and claim a positive control cannot fit without consuming p9, immediately before the new control that disproves it. Remove or rewrite both blocks so the comments describe the current shared-event invariant.
  - file: Tests/HerdrKitTests/SessionRecoveryTests.swift
  - line: 175

**Tests Run**
- Reviewed pinned SHA 25c4ad95ac708d1610185f6d7c1142b06e3e8c5a through local Git plumbing. The assigned worktree HEAD was stale at c85189e81e87671fa33cc9e0c351daacac8fd650, so all review and probes used an archive of the pinned SHA. The worktree remained clean.
- Content discriminator: Tests/HerdrKitTests/SessionRecoveryTests.swift blob ae3a9d0bb333652b33e3f80043e51dbca41e7171. Verbatim central lines: `let probe = ClientEvent.paneCreated("p9", from: minted)` and `let probe = ClientEvent.paneCreated("adoption-probe", from: attemptB)`.
- Clean pinned archive: `swift test -Xswiftc -warnings-as-errors` executed 146 tests with 0 failures and 22 environment-gated skips on this image.
- Replaced both shared probe initializers' attempts with stranger AttemptIDs. The mutant built under warnings-as-errors and both focused tests failed at their post-adoption subscribe assertions. KILLED x2; the shared immutable event value genuinely binds both deliveries.
- Moved site 2's negative delivery after adoption. The mutant built and failed because the delivery subscribed instead of remaining empty. KILLED.
- Moved site 2's positive delivery before adoption. The mutant built and failed because no subscription appeared. KILLED.
- Swapped the post-adoption adoption-probe delivery and the authoritative snapshot. The test built and passed. This is expected independence, not redundancy: both occur after adoption and affect distinct panes; the stranger mutation pins the probe while the p9 collision pins the snapshot.
- Changed site 2's shared probe pane and matching positive expectation from adoption-probe to p9. The mutant built and failed the untouched snapshot expectation with only p1 newly subscribed. KILLED; the distinct probe pane is load-bearing.
- No macOS build was run; the sender reported the buildbox remains unavailable for this head.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c82067195a3dd5`

**Decision:** `approved`
**Summary:** No findings. The rewritten comments accurately distinguish historical survivors from the current shared-event construction and match the code and measured ordering behavior.

**Tests Run**
- Reviewed pinned SHA c7e6071fab2b684d5c61287e22566db1fb574666 through local Git plumbing; worktree HEAD found was the same SHA. The worktree remained clean.
- Content discriminator: Tests/HerdrKitTests/SessionRecoveryTests.swift blob 54b8b38e820675b8a85d263e84f3b9c95aceaba4. Verbatim central text: `ATTEMPT DISCRIMINATION comes from the shared `probe` value, not from this control.`
- Audited all mutation-status claims in SessionRecoveryTests.swift. The three SURVIVED occurrences are historical and each explains the construction that now prevents that survivor; the unrelated concurrency claim at line 786 still matches its same-key contention setup.
- Verified the rewritten site-2 ordering block against the adjacent code and round-2 probes: negative-before-adoption and positive-after-adoption are the load-bearing order; the post-adoption probe and snapshot are correctly described as independent because they affect distinct panes.
- Checked referenced test names, shared authority fields, fallback expressions, and the PR diff for stale or contradictory prose. No additional mismatch found. `git diff --check` passed.
- Clean pinned archive: `swift test -Xswiftc -warnings-as-errors` executed 146 tests with 0 failures and 21 environment-gated skips on this image.
- No macOS build was run; the sender reported that the buildbox does not watch this branch.

PR #21, opened 2026-08-03
feat(list): agent list model with fail-open-visible unknown statuses

### LINKED ISSUE #20: Agent list model: typed status, urgency grouping, and fail-open-visible unknowns
Screen 2 ("Agents") is the home screen and ~80% of the app's value. It answers one question — *does anything need me?* — by grouping agents by urgency. None of that model exists yet.

## What already exists (not rebuilding)

- `InputRouter` / `InputPlan` / `InputAction` — the reply path **and** the key row are done (`prompt` / `text` / `keys` / `refused`).
- `PaneRead` / `ReadSource` / `ReadFormat` — terminal reading.
- `RefreshPolicy` / `RefreshCoordinator` — revision-gated refresh.
- `SessionRecovery` / `RecoveryExecutor` — reconnection.

## What is missing

`AgentInfo.agentStatus` is a raw `String?`, and the only consumer is `isWorking: Bool { agentStatus == "working" }`. There is no typed status, and nothing groups or sorts agents at all.

## The part worth getting right

herdr's `AgentStatus` serialises as `idle | working | blocked | done | unknown`. Combined with the field being optional, the client faces **three genuinely different "I don't know" cases**:

1. `nil` — field absent. Not an agent pane, or an older server.
2. `"unknown"` — the server looked and could not tell.
3. `"waiting_approval"` — a value this client has never heard of, from a newer server.

Collapsing these into one bucket is the defect. Case 3 is the dangerous one: a future herdr introducing a new blocked-like state would be silently sorted as *idle*, and the user would never be told an agent was waiting. The failure is invisible and permanent.

**Fail-closed here means sorting unknown states UP, not down.** An unrecognised status must be surfaced, because "I do not know what this agent is doing" is closer to *needs attention* than to *nothing to do*. Sorting it to the bottom is the flattening that makes the bug unobservable.

## Scope

- A typed `AgentStatus` with an explicit case per wire value, plus a distinct case carrying the unrecognised string verbatim, and a distinct case for absent.
- `AgentGroup`: needs-you → stopped → unrecognised → working → idle, with a stable order within each group so the list does not churn between refreshes.
- An `AgentList` model mapping `[AgentInfo]` to grouped sections plus the needs-you count.

## Evidence bar

Mutation-verified tests, per this lane's standing bar. Specifically required:

- A test that fails if a **new, unrecognised** status string is sorted as idle. This is the whole point of the issue and must be armed against the exact regression.
- A test that distinguishes all three unknown cases from each other — not merely that each is "not blocked".
- Premise mutation on each: break the setup and prove the assertion was actually armed.

Linux-testable; no UIKit. macOS check follows via the buildbox, which now watches `main`.

jerryfane:
Closes #20.

Screen 2 is the home screen and the app's whole argument — *does anything need me?* Nothing modelled it: `AgentInfo.agentStatus` was a raw `String?` whose only consumer was `isWorking: Bool { agentStatus == "working" }`.

**Not rebuilt, because it already existed:** `InputRouter`/`InputPlan` already cover the reply path *and* the key row; `PaneRead` covers terminal reading. I had offered to build both — checking first avoided duplicating them.

## The design point: three ways of not knowing

herdr sends `idle|working|blocked|done|unknown`, and the field is optional. So a client can fail to know in three distinct ways, and merging them is the defect:

| case | means |
|---|---|
| `.absent` | no field — not an agent pane, or an older server |
| `.indefinite` | server sent `"unknown"` — it looked and couldn't tell |
| `.unrecognised(String)` | a value this build has never heard of |

The third is why this exists. Under a `default: .idle` mapping, a future herdr adding a blocked-like state would sort an agent **waiting on a human** into the one section that starts collapsed — invisible, permanent, and indistinguishable from the feature working.

So unrecognised **sorts above working and idle**, never starts collapsed, and blocks the all-clear. Wrong that way costs a stale row shown too prominently; wrong the other way is silent.

`group(for:isLive:)` is exhaustive with **no `default:` arm** — adding an `AgentStatus` case fails to compile rather than falling through to a guess. The compile error is the guard.

Also fail-closed: a `nil` pane census means *liveness unknown*, not *everything died*. Inferring death from missing evidence would turn one failed call into a screen claiming all your work crashed.

## Evidence

**7 production mutations, all KILLED:** `novel-status-becomes-idle`, `unrecognised-groups-as-idle`, `collapse-unknown-into-absent`, `drop-stable-tiebreak`, `no-census-means-dead`, `quiet-ignores-unrecognised`, `unrecognised-starts-collapsed`.

**3 premise mutations — and one found a vacuous test of mine.** `testOrderWithinAGroupIsStableAgainstServerOrder` **SURVIVED** having its two inputs made identical: both sides then trivially matched the expected order, so it had never exercised a differing server order at all. Rewritten to build both lists from named arrays with an assertion that the arrays differ — now KILLED in both directions.

I wrote a vacuous test while guarding against vacuous tests. That is the argument for running premise mutation rather than trusting the author, and I'd rather it be in the record than quietly fixed.

## What I want challenged

- Is `.unrecognised` sorting *above* `working` right, or is it alarmist? I argued the asymmetry (silent failure vs. a prominent stale row), but it is a judgement call.
- `.absent` groups as `.idle` while `.indefinite` groups as `.unrecognised`. Defensible — an absent field usually means "not an agent pane" — but it is the one place I merge a not-knowing into a definite group.
- Anything else in the 13 tests that would survive its premise being broken.

159 tests, 0 failures, `-Xswiftc -warnings-as-errors`. Linux only; the buildbox now watches `main`.

### DIFF
--- a/Sources/HerdrKit/AgentList.swift
+++ b/Sources/HerdrKit/AgentList.swift
@@ -7,8 +7,9 @@ import Foundation
 /// ways of not knowing, and the whole value of this type is refusing to merge
 /// them:
 ///
-///   - `.absent`      the field was not sent. Not an agent pane, or an older
-///                    server that predates the field.
+///   - `.absent`      the field was not sent. `agent.list` only ever returns
+///                    agent terminals, so this means an older server or a gap
+///                    — NOT "this is not an agent".
 ///   - `.indefinite`  the server sent `"unknown"`. It looked and could not tell.
 ///   - `.unrecognised` the server sent something this client has never heard of.
 ///
@@ -135,8 +136,15 @@ public struct AgentRow: Equatable, Sendable, Identifiable {
         // client is out of date", which are different problems with different
         // fixes — but they group identically, because the user's situation is
         // the same either way.
-        case .indefinite, .unrecognised: return .unrecognised
-        case .absent: return .idle
+        // ALL THREE NOT-KNOWINGS SURFACE. `.absent` was grouped as `.idle`
+        // here, on the reasoning that a missing status means "not an agent
+        // pane". That reasoning was wrong, and the reviewer checked the server
+        // rather than argue it: `agent_info` returns None unless
+        // `terminal.is_agent_terminal()` (herdr src/app/agents.rs:368), so
+        // every row in `agent.list` IS an agent. An absent status means an
+        // older server or a gap — never a non-agent — and calling it definite
+        // idle collapsed the exact case this file exists to surface.
+        case .indefinite, .unrecognised, .absent: return .unrecognised
         }
     }
 }
@@ -152,7 +160,11 @@ public struct AgentList: Equatable, Sendable {
     /// What the top of the screen says. Counts only positively-blocked agents —
     /// an unrecognised status is surfaced by its own section, and inflating this
     /// number with maybes would make the one number on the screen untrustworthy.
-    public var needsYouCount: Int { rows.filter { $0.status.isBlocked }.count }
+    /// Counts the GROUP, not the retained status. Reading `status.isBlocked`
+    /// directly meant a blocked agent whose pane had since vanished sorted into
+    /// `.stopped` and made `isQuiet` true while still reporting 1 here — the
+    /// screen simultaneously claiming all-clear and one-waiting.
+    public var needsYouCount: Int { rows.filter { $0.group == .needsYou }.count }
 
     /// True when nothing is blocked AND nothing is unrecognised. The quiet state
     /// has to mean "I checked everything", so an uninterpretable agent must
--- a/Sources/HerdrKit/AgentList.swift
+++ b/Sources/HerdrKit/AgentList.swift
@@ -1,0 +1,192 @@
+import Foundation
+
+/// What an agent is doing, as the client is willing to act on it.
+///
+/// herdr serialises its own `AgentStatus` as `idle | working | blocked | done |
+/// unknown`, and the field is optional. That gives the client THREE distinct
+/// ways of not knowing, and the whole value of this type is refusing to merge
+/// them:
+///
+///   - `.absent`      the field was not sent. Not an agent pane, or an older
+///                    server that predates the field.
+///   - `.indefinite`  the server sent `"unknown"`. It looked and could not tell.
+///   - `.unrecognised` the server sent something this client has never heard of.
+///
+/// The third is the one that matters. A future herdr adding, say,
+/// `"waiting_approval"` would — under a `default: .idle` mapping — sort an agent
+/// that is blocked on a human into the quietest group on the screen, and nothing
+/// would ever report it. The bug would be invisible, permanent, and would look
+/// exactly like the feature working.
+///
+/// So the raw string is CARRIED, not discarded. It costs one associated value
+/// and it is the difference between a client that degrades and one that lies.
+public enum AgentStatus: Equatable, Sendable {
+    case idle
+    case working
+    case blocked
+    case done
+    /// The server sent `"unknown"` — it looked and could not determine a state.
+    case indefinite
+    /// A value this build does not know. The payload is retained verbatim so it
+    /// can be surfaced and diagnosed rather than guessed at.
+    case unrecognised(String)
+    /// No `agent_status` field at all.
+    case absent
+
+    /// Maps the wire value. Deliberately has no `default:` collapsing arm — an
+    /// unmatched string becomes `.unrecognised(raw)`, never a real state.
+    public init(wire: String?) {
+        guard let wire else { self = .absent; return }
+        switch wire {
+        case "idle": self = .idle
+        case "working": self = .working
+        case "blocked": self = .blocked
+        case "done": self = .done
+        case "unknown": self = .indefinite
+        default: self = .unrecognised(wire)
+        }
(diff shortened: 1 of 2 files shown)


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c832f643a729e5`

**Summary:** Two model defects should be fixed before merge. Unknown-status prominence is justified, and rows-only equality is sound because rows and sections are immutable and sections are derived exclusively from rows.

**Findings**
- **An absent status is silently treated as idle** (high)
  - detail: `agent.list` contains agent terminals, while an absent optional status represents an older-server agent rather than an ordinary non-agent pane. Grouping `.absent` as `.idle` collapses every such agent, permits `isQuiet == true`, and defeats the model's fail-open-visible guarantee. Group absence with the unknown cases, or otherwise surface it visibly.
  - file: Sources/HerdrKit/AgentList.swift
  - line: 139
- **Stopped blocked rows still increment the needs-you count** (medium)
  - detail: `needsYouCount` reads the retained raw status instead of the liveness-aware group. A blocked agent missing from `livePaneIDs` therefore appears in `.stopped`, makes `isQuiet == true`, but still reports `needsYouCount == 1`. Count `.needsYou` groups rather than stale blocked statuses.
  - file: Sources/HerdrKit/AgentList.swift
  - line: 155
- **Two absence-shaped tests have vacuous premises** (low)
  - detail: `testNoCensusTreatsEveryAgentAsLive` still passes after removing every agent, and the quiet half of `testQuietRequiresNothingBlockedAndNothingUnrecognised` still passes after replacing its idle/done fixture with an empty list. Assert the expected rows or statuses before asserting their absence-based conclusions.
  - file: Tests/HerdrKitTests/AgentListTests.swift
  - line: 171

**Tests Run**
- Reviewed pinned SHA ec40bba37d0156908878c727ee75aee223b3e5ce through local Git plumbing; worktree HEAD found: ec40bba37d0156908878c727ee75aee223b3e5ce. Worktree remained clean.
- Content evidence: Sources/HerdrKit/AgentList.swift blob de82b6f03840ecfeeb0e2de51918fa8d3f33ad0e; critical lines read verbatim: `case .absent: return .idle` and `public var needsYouCount: Int { rows.filter { $0.status.isBlocked }.count }`. Test blob: 1c2dcfa2f44e170f2929db881d6b96261c59b09c.
- Swift 6.1 Linux container: `swift test -Xswiftc -warnings-as-errors` completed with 159 tests, 21 environment-gated skips, and 0 failures.
- Temporary older-server-agent probe built and failed: the absent-status row grouped as idle, allowed quiet, and started collapsed.
- Temporary stopped-blocked probe built and failed: the row grouped as stopped and was quiet, but `needsYouCount` was 1 instead of 0.
- Premise mutations built and survived: removing all agents from the no-census test and removing the idle/done fixture from the quiet test both left their tests green.
- Host invocation could not start because `swift` is not installed; testing was rerun successfully in the locally available Swift container.
- Swift 5.9 compatibility attempt was invalid for this PR because pre-existing SSHTransport and RecoveryExecutor code elsewhere in the branch does not compile on that toolchain; AgentList.swift itself compiled before the package-level failure.
- `git diff --check` passed.

**Needs**
- Surface `.absent` without treating it as definite idle, make the headline count liveness-aware, and arm the two vacuous test premises.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c833d60b346fb2`

**Decision:** `approved`
**Summary:** No findings. All three round-one defects are correctly fixed and regression-guarded. The unrecognised group should remain above working even with absent statuses included: ordering represents uncertainty risk, not likelihood, while `needsYouCount` remains reserved for positively blocked agents. An older server producing many absent statuses should visibly degrade instead of appearing confidently quiet.

**Tests Run**
- Reviewed pinned SHA 277713890b362a956803712ed3b97c0cc59da055 through local Git plumbing; worktree HEAD found: 277713890b362a956803712ed3b97c0cc59da055. Worktree remained clean.
- Content evidence: Sources/HerdrKit/AgentList.swift blob 974ed6b6560811bdbb28289207f5120b3c2d01b7; critical lines read verbatim: `case .indefinite, .unrecognised, .absent: return .unrecognised` and `public var needsYouCount: Int { rows.filter { $0.group == .needsYou }.count }`. Tests/HerdrKitTests/AgentListTests.swift blob 1245f56d7eafbb818c6406c7e450e02c8ecef0d5.
- Swift 6.1 Linux container: `swift test -Xswiftc -warnings-as-errors` completed with 161 tests, 22 environment-gated skips, and 0 failures.
- Production reversion probes built and were killed: restoring absent-to-idle failed `testAnAbsentStatusSurfacesRatherThanReadingAsIdle`; restoring raw-status counting failed `testAStoppedAgentDoesNotCountEvenIfItsLastStatusWasBlocked`.
- Premise probes built and were killed: replacing the absent status with idle, making the stopped pane live, removing every no-census row, and removing the quiet idle/done fixture each failed at its positive precondition.
- Reviewed every AgentListTests fixture and absence-shaped assertion; no additional vacuous premise was found. Tests using absence either now establish rows/statuses first, unwrap a required row, or intentionally test the empty-list identity case.
- `git diff --check` passed.

PR #23, opened 2026-08-03
feat(terminal): fold unwrapped output to phone width losslessly

### LINKED ISSUE #22: Re-wrap unwrapped terminal output to phone width without losing characters
Screen 3 shows terminal output on a ~390pt screen. The pane it comes from is sized for a desktop — often 120 columns — and `pane.resize` cannot change that independently, because it grows a pane against its neighbours in the *desktop* layout. There is no "set this pane to 40 columns".

`ReadSource.recentUnwrapped` is the way out: herdr returns scrollback with **no hard wrapping**, i.e. original logical lines. The client then folds them to whatever width the phone actually has. Nothing does that yet.

## Scope

A pure function: logical lines + column width → display lines. No I/O, no UIKit, Linux-testable.

## The properties that matter

1. **Nothing is lost.** Stripping whitespace from the input and from the joined output must produce identical strings. A wrap that silently drops a character is worse than no wrap at all, and it is invisible by eye on a wall of log output.
2. **It always terminates.** A token longer than the width must hard-break rather than loop. Width 1 with a wide grapheme must still make progress.
3. **No line exceeds the width**, except where a single indivisible grapheme cannot fit — which must be an explicit, named case rather than an accident.
4. **Blank lines survive.** Terminal output uses them structurally; collapsing them reflows meaning.
5. **Break at spaces where possible, hard-break only when a token cannot fit.** Agent output is prose, paths and code mixed together; character-exact wrapping shreds prose, and word-only wrapping overflows on a long path.

## The fail-closed decision

A non-positive width is a caller bug, not data. Returning `[]` would silently delete the user's output; returning the input unfolded is visually wrong but lossless. **Prefer lossless.** Losing terminal output to a layout bug is the worse failure, and it is the one the user cannot detect.

## Evidence bar

Mutation-verified, per this lane's standing bar. Specifically:

- A property test for (1) over adversarial inputs — long unbroken tokens, mixed whitespace, blank lines, multi-scalar graphemes.
- An explicit termination test for width 1 and for a token far exceeding width.
- Premise mutation on each: break the setup and prove the assertion was armed. Four vacuous tests of mine surfaced in #21; assume the same rate here.

jerryfane:
Closes #22.

Screen 3 shows a desktop-sized pane on a ~390pt screen. `pane.resize` can't help — it grows a pane against its neighbours in the *desktop* layout, so there's no "make this pane 40 columns" without disturbing the machine the user is sitting at. `ReadSource.recentUnwrapped` returns logical lines with no hard wrapping; this folds them to the width the phone actually has.

Pure function. No I/O, no UIKit, fully exercised on Linux.

## The load-bearing property

**Losslessness.** Stripping whitespace from the input and from the joined output must give identical strings. A wrap that eats a character is invisible against a wall of log output and corrupts exactly what the user opened the screen to read.

## Three decisions worth arguing with

**A non-positive width returns the input UNFOLDED, not empty.** Width 0 is a caller bug — an unmeasured layout — not a statement about the data. Returning `[]` deletes terminal output silently in service of a layout mistake. Unfolded is visibly wrong and recovers next layout pass; lost output doesn't.

**Termination is argued, not hoped.** A token wider than the line is split by `prefix(width)` with `width > 0`, so ≥1 Character is consumed per pass and the remainder strictly shrinks. Width 1 against a multi-scalar grapheme takes the grapheme and **overflows** rather than stalling — the one named exception to the width rule, tested with flag emoji.

**Trailing whitespace is trimmed at fold points only, never on the final line.** A space removed mid-fold is an artefact the fold created, like the leading space dropped from a continuation. A line that genuinely ends in spaces is content. Conflating them loses data at one end or leaves invisible padding at the other.

Interior whitespace runs survive as their own tokens — column alignment in tables, tree output and diff gutters is made of exactly those runs, and eating them turns aligned output into prose. That's a silent corruption of *meaning*, which is worse than one of characters.

## Evidence

**Production mutations, all KILLED:** `width0-returns-empty`, `blank-line-vanishes`, `no-hardbreak-report`, `drop-interior-space`.

**Premise mutations — and the first caught another vacuous test of mine.** `testFoldingNeverLosesContent` **SURVIVED** having its case list emptied: zero iterations, zero assertions, green, and reading in the diff as thorough coverage. It now counts folds that actually produced multiple lines and requires >20, so it fails both when there are no inputs *and* when the inputs never fold. KILLED both ways.

That's the fifth vacuous test of mine in two days. I predicted the class would recur when dispatching #21, and it recurred on the very first property test I wrote afterwards. Knowing about a defect does not prevent it; probing does.

## What I want challenged

- The trailing-space asymmetry (trim at folds, keep on the final line) is a judgement. If a caller renders these into a fixed-width view, invisible trailing space on the last line may matter.
- Width is counted in Swift `Character`s. A CJK or emoji glyph occupies two terminal columns, so a line of them will render wider than `width` even though the count is right. I have not handled display width, and I'd rather that be a stated limitation than a discovered one.
- Any remaining absence-shaped assertion that survives its premise being broken.

171 tests, 0 failures, `-Xswiftc -warnings-as-errors`.

### DIFF
--- a/Sources/HerdrKit/TerminalWrap.swift
+++ b/Sources/HerdrKit/TerminalWrap.swift
@@ -1,0 +1,153 @@
+import Foundation
+
+/// Folds unwrapped terminal output to a narrow display width.
+///
+/// The pane this text came from is sized for a desktop — commonly 120 columns —
+/// and the phone cannot change that: `pane.resize` grows a pane against its
+/// neighbours in the DESKTOP layout, so there is no "make this pane 40 columns"
+/// without disturbing the machine the user is sitting at. `ReadSource
+/// .recentUnwrapped` is the way out. herdr returns logical lines with no hard
+/// wrapping, and the client folds them to whatever width it actually has.
+///
+/// This is a pure function over strings: no I/O, no UIKit, and fully exercised
+/// by the Linux test suite.
+public enum TerminalWrap {
+
+    /// The result of folding one logical line.
+    public struct Folded: Equatable, Sendable {
+        /// Display lines, in order. Never empty — a blank input line folds to
+        /// one blank output line, because terminal output uses blank lines
+        /// structurally and collapsing them reflows meaning.
+        public let lines: [String]
+        /// True when at least one break was forced mid-token because the token
+        /// could not fit the width. Surfaced rather than hidden: it is the
+        /// signal that the width is too narrow for the content, and a caller
+        /// showing a "content is wider than the screen" affordance needs it.
+        public let hardBroke: Bool
+    }
+
+    /// Folds `lines` to `width` columns.
+    ///
+    /// NON-POSITIVE WIDTH RETURNS THE INPUT UNFOLDED, deliberately. A width of
+    /// zero or less is a caller bug — a layout that has not measured yet, a
+    /// misplaced inset — not a statement about the data. The tempting responses
+    /// are to return `[]` or to trap; both destroy the user's terminal output in
+    /// service of a layout mistake, and the empty case does it silently. An
+    /// unfolded line is visibly wrong and recoverable on the next layout pass.
+    /// Losing output is neither.
+    public static func fold(_ lines: [String], width: Int) -> [Folded] {
+        guard width > 0 else {
+            return lines.map { Folded(lines: [$0], hardBroke: false) }
+        }
+        return lines.map { fold(line: $0, width: width) }
+    }
+
+    static func fold(line: String, width: Int) -> Folded {
+        // A BLANK LINE IS ONE BLANK LINE, not zero. Returning [] here would make
+        // the output shorter than the input and quietly close up the paragraph
+        // breaks that make agent output readable.
+        guard !line.isEmpty else { return Folded(lines: [""], hardBroke: false) }
+
+        var out: [String] = []
+        var current = ""
+        var currentCount = 0
+        var hardBroke = false
+
+        /// Emits `current` and resets. Kept as one operation so no path can
+        /// append without also clearing the counter.
+        ///
+        /// TRAILING WHITESPACE IS TRIMMED HERE AND NOWHERE ELSE. flush() is
+        /// called only at fold points — where the next token did not fit — so
+        /// the space being removed is one the fold created, exactly like the
+        /// leading space dropped from a continuation line. Symmetry matters:
+        /// keeping it produced lines that were visually short of the width by an
+        /// invisible character. The FINAL line does not go through flush(), so a
+        /// line genuinely ending in spaces keeps them; that is content, not an
+        /// artefact, and the two cases must not be conflated.
+        func flush() {
+            while let last = current.last, last.isWhitespace { current.removeLast() }
+            out.append(current)
+            current = ""
+            currentCount = 0
+        }
+
+        for token in tokenise(line) {
+            let tokenCount = token.count
+
+            // A token that cannot fit ANY line must be split, whatever the
+            // current state. Trying to place it whole would either overflow the
+            // width or, if the loop instead flushed and retried, spin forever on
+            // a token that never fits.
+            if tokenCount > width {
+                if currentCount > 0 { flush() }
+                var rest = Substring(token)
+                while !rest.isEmpty {
+                    // TERMINATION. `width` is > 0 here, so `prefix(width)` takes
+                    // at least one Character on every pass and `rest` strictly
+                    // shrinks. There is no input for which this does not finish,
+                    // including width 1 against a multi-scalar grapheme — Swift
+                    // Characters are indivisible, so one is taken and the line
+                    // exceeds the width rather than the loop stalling.
+                    let chunk = rest.prefix(width)
+                    rest = rest.dropFirst(chunk.count)
+                    if rest.isEmpty {
+                        current = String(chunk)
+                        currentCount = chunk.count
+                    } else {
+                        out.append(String(chunk))
+                        hardBroke = true
+                    }
+                }
+                continue
+            }
+
+            if currentCount == 0 {
+                // Leading whitespace on a fresh line is dropped: it is an
+                // artefact of the fold, not content. A run of spaces at the
+                // START of a logical line is preserved by tokenise, which emits
+                // it as part of the first token.
+                if token.allSatisfy(\.isWhitespace) { continue }
+                current = token
+                currentCount = tokenCount
+            } else if currentCount + tokenCount <= width {
+                current += token
+                currentCount += tokenCount
+            } else {
+                flush()
+                if token.allSatisfy(\.isWhitespace) { continue }
+                current = token
+                currentCount = tokenCount
+            }
+        }
+
+        if currentCount > 0 || out.isEmpty { out.append(current) }
+        return Folded(lines: out, hardBroke: hardBroke)
+    }
+
+    /// Splits a line into alternating word and whitespace runs, preserving both.
+    ///
+    /// Whitespace is kept as its own token rather than discarded, so interior
+    /// spacing survives a fold that does not happen to break there. Column
+    /// alignment in terminal output — tables, tree output, diff gutters — is
+    /// made of exactly those runs, and eating them turns aligned output into
+    /// prose.
+    static func tokenise(_ line: String) -> [String] {
+        var tokens: [String] = []
+        var current = ""
+        var currentIsSpace: Bool?
+
+        for ch in line {
+            let isSpace = ch.isWhitespace
+            if currentIsSpace == nil || isSpace == currentIsSpace {
+                current.append(ch)
+                currentIsSpace = isSpace
+            } else {
+                tokens.append(current)
+                current = String(ch)
+                currentIsSpace = isSpace
+            }
+        }
+        if !current.isEmpty { tokens.append(current) }
+        return tokens
+    }
+}
--- a/Sources/HerdrKit/TerminalWrap.swift
+++ b/Sources/HerdrKit/TerminalWrap.swift
@@ -197,9 +197,26 @@ public enum TerminalWrap {
             // line from a continuation the fold created. Only the latter's
             // leading space is an artefact.
             if index == 0 && isWhitespaceToken {
-                current = token
-                currentCols = columns(of: token, startingAt: 0)
                 currentIsLeadingIndent = true
+                if columns(of: token, startingAt: 0) <= width {
+                    current = token
+                    currentCols = columns(of: token, startingAt: 0)
+                } else {
+                    // INDENT WIDER THAN THE SCREEN. The fast path above stored it
+                    // whole and emitted one over-width line — a reviewer probe of
+                    // ten leading spaces at width 6 produced a ten-cell line.
+                    // Whitespace is divisible (unlike a tab), so split it to
+                    // width instead of overflowing, keeping the remainder as the
+                    // still-leading indent the first word will try to join.
+                    var rest = Substring(token)
+                    while columns(of: rest, startingAt: 0) > width {
+                        let chunk = prefixFitting(rest, width: width)
+                        out.append(String(chunk))
+                        rest = rest[chunk.endIndex...]
+                    }
+                    current = String(rest)
+                    currentCols = columns(of: current, startingAt: 0)
+                }
                 continue
             }
 
--- a/Sources/HerdrKit/TerminalWrap.swift
+++ b/Sources/HerdrKit/TerminalWrap.swift
@@ -26,6 +26,90 @@ public enum TerminalWrap {
         public let hardBroke: Bool
     }
 
+    // MARK: - terminal columns
+
+    /// Columns one character occupies in a terminal, at column `col`.
+    ///
+    /// WIDTH IS MEASURED IN TERMINAL CELLS, NOT SWIFT CHARACTERS. The first
+    /// version of this file counted `String.count`, and a reviewer's probe
+    /// folded `界界界` at width 4 into a single SIX-column line. A function whose
+    /// entire purpose is fitting text to a phone's terminal width cannot
+    /// measure in a unit the terminal does not use.
+    ///
+    /// This is an approximation of `wcwidth`, not a port of it — full fidelity
+    /// needs the Unicode East Asian Width table. It covers the ranges that
+    /// actually appear in agent output: CJK, Hangul, fullwidth forms and emoji.
+    /// Stated as an approximation because a caller relying on exactness for,
+    /// say, box-drawing alignment deserves to know it is not guaranteed.
+    static func columns(of ch: Character, at col: Int) -> Int {
+        // A tab advances to the next multiple of 8. It is the only character
+        // whose width depends on where it sits, which is why every width
+        // calculation here threads a starting column rather than summing
+        // per-character widths in isolation.
+        if ch == "\t" { return 8 - (col % 8) }
+
+        guard let scalar = ch.unicodeScalars.first else { return 0 }
+        let v = scalar.value
+
+        // Control characters occupy no cells.
+        if v < 0x20 || (v >= 0x7F && v < 0xA0) { return 0 }
+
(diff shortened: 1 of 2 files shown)


jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c83637d1ed777d`

**Summary:** Two behavioral issues block the advertised terminal-width contract. Returning input unfolded for non-positive widths is the correct fail-visible choice, and trimming whitespace only at genuine fold boundaries is defensible. However, leading indentation is currently deleted, and Swift Character counts cannot enforce terminal-column width for wide glyphs or tabs.

**Findings**
- **Leading indentation is silently deleted** (high)
  - detail: At the start of a logical line, `currentCount == 0` and the first whitespace token is unconditionally skipped. `fold("    child", width: 80)` therefore returns `child` despite no fold being necessary. This contradicts the adjacent comment and corrupts indentation used by code, trees, tables, and log structure. Distinguish the original logical-line start from a continuation created by folding.
  - file: Sources/HerdrKit/TerminalWrap.swift
  - line: 109
- **The advertised column width is not enforced** (medium)
  - detail: All fit and split decisions use Swift Character counts. A probe folding `界界界` at width 4 returned one six-column display line instead of `界界` plus `界`. Tabs similarly occupy a variable number of terminal cells while counting as one Character. Since the function's purpose is phone-width terminal rendering, use terminal-cell width accounting or change the contract before screen 3 depends on it.
  - file: Sources/HerdrKit/TerminalWrap.swift
  - line: 75
- **The losslessness helper and three more premises can be vacuous** (low)
  - detail: Four compiling test mutations survived: making both normalized helper values `""`; using empty text in the width-discipline test; emptying the non-positive-width loop; and removing all interior spacing from the spacing test. Add a canary proving the helper's normalization distinguishes known content, count the width cases, and positively assert that the width/spacing fixtures contain the properties under test.
  - file: Tests/HerdrKitTests/TerminalWrapTests.swift
  - line: 18

**Tests Run**
- Reviewed pinned SHA 26bb25ec7cd599a98d14d1ed3d4a6c381222e6a1 through local Git plumbing; worktree HEAD found: 26bb25ec7cd599a98d14d1ed3d4a6c381222e6a1. Worktree remained clean.
- Content evidence: Sources/HerdrKit/TerminalWrap.swift blob 9a0f11afae1460999b654cf364d0b145952d122f; critical line read verbatim: `if token.allSatisfy(\.isWhitespace) { continue }`. Tests/HerdrKitTests/TerminalWrapTests.swift blob 1e05130c0967f8a6235deff299a5e37577f8a30a.
- Swift 6.1 Linux container: `swift test -Xswiftc -warnings-as-errors` completed with 171 tests, 23 environment-gated skips, and 0 failures.
- Temporary leading-indentation regression built and failed: `fold("    child", width: 80)` produced `["child"]` rather than preserving the input.
- Temporary terminal-column regression built and failed: folding three double-column CJK glyphs at width 4 produced one line instead of two.
- Compiling helper/premise mutations survived their targeted tests: constant-empty normalization, empty width-discipline text, an empty non-positive-width loop, and a spacing fixture with no interior spacing.
- `git diff --check` passed.

**Needs**
- Preserve original leading indentation, account for terminal display columns, and arm the four surviving helper/premise mutations.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c8373378e4be69`

**Summary:** Two findings remain. The new premise guards and basic reversion tests are effective, but leading indentation is still deleted at a narrower width, and common double-width emoji are still measured as one cell.

**Findings**
- **Leading indentation is still deleted when the first word cannot join it** (high)
  - body: `fold("    child", width: 6)` stores the initial indentation, but the following word cannot fit beside it. `flush()` then removes every trailing whitespace character from the indentation-only buffer and emits an empty line, so the joined result no longer begins with the four spaces. A compiled regression probe failed. The losslessness helper cannot detect this because it intentionally strips whitespace. Preserve logical-line indentation when flushing an indentation-only prefix, and add this narrow-width case.
  - line: 148
  - path: Sources/HerdrKit/TerminalWrap.swift
- **The terminal-cell approximation undercounts common emoji** (medium)
  - body: The selected ranges omit common double-width BMP emoji and emoji-presentation graphemes. A compiled probe measured `✅`, `⌚`, and `1️⃣` as 1 and folded `✅✅✅` into one line at width 4. Local libc `wcwidth` reports U+2705 and U+231A as 2; U+2500 box drawing is correctly 1, so box drawing itself was not the reproduced problem. Use a complete width table/library or cover the omitted wide and emoji-presentation cases so the stated terminal-cell contract holds.
  - line: 57
  - path: Sources/HerdrKit/TerminalWrap.swift

**Tests Run**
- Reviewed pinned SHA a49cbc9539edfb65a32d28378f5a4df78a24cb8f through Git plumbing; worktree HEAD found: a49cbc9539edfb65a32d28378f5a4df78a24cb8f. Worktree was clean.
- Content evidence: Sources/HerdrKit/TerminalWrap.swift blob 99220d032a49e02a2375451783586e5ab1954080; critical line verbatim: `if index == 0 && isWhitespaceToken {`. Tests/HerdrKitTests/TerminalWrapTests.swift blob e56c856359323c26d08afb40a41988a75c988bf2; critical line verbatim: `XCTAssertNotEqual(strip("abc"), strip("abd"), "the normaliser cannot distinguish content")`.
- Full Linux suite from an archive of the pinned commit under Swift 6.1 with `-Xswiftc -warnings-as-errors`: 175 tests, 21 skipped, 0 failures.
- Compiled indentation probe: `fold("    child", width: 6).lines.joined().hasPrefix("    ")` failed; logical-line indentation was deleted.
- Compiled terminal-width probe: `✅`, `⌚`, and `1️⃣` each returned 1 rather than 2; `fold("✅✅✅", width: 4)` returned `["✅✅✅"]` rather than `["✅✅", "✅"]`.
- Local libc `wcwidth` comparison under C.UTF-8: U+2705=2, U+231A=2, U+2500=1.
- All mutations built before classification. KILLED: degenerate losslessness normaliser, empty width-discipline text, empty non-positive-width loop, spacing fixture without a whitespace run, and deletion of the leading-indentation branch.
- `git diff --check` passed; final worktree remained clean. No macOS result was claimed or run.

**Needs**
- Preserve leading indentation when it occupies a buffer that must be flushed before the first word.
- Make terminal-cell measurement handle common BMP emoji and emoji-presentation graphemes, with regressions for the demonstrated cases.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c8398b4da7f8a0`

**Summary:** Two findings remain: oversized leading indentation bypasses width enforcement, and the flag-clear half of the indentation fix is unguarded. The emoji repair is acceptable for the reported/common cases.

**Findings**
- **Oversized leading indentation bypasses the token splitter** (medium)
  - body: The leading-whitespace branch stores the entire token and continues before the oversized-token check at line 211. A compiled probe of ten leading spaces plus `child` at width 6 preserved the content but emitted one ten-cell line, violating the requested width even though the whitespace run is divisible. The tab-only and hard-breaking-word variants preserved content correctly. Route oversized leading whitespace through width-aware splitting; only an indivisible character such as a tab may need the documented overflow behavior.
  - line: 199
  - path: Sources/HerdrKit/TerminalWrap.swift
- **The fold-point half of the indentation state is not regression-tested** (low)
  - body: Deleting `currentIsLeadingIndent = false` builds successfully and all 17 committed TerminalWrap tests remain green. That mutation makes `fold("    a b c", width: 6)` retain the fold-point space after `a`, contradicting the comment and intended behavior. The added reviewer probe expecting `["    a", "b c"]` passes on this head and would kill the mutation; commit an equivalent regression so both directions are actually guarded.
  - line: 241
  - path: Sources/HerdrKit/TerminalWrap.swift

**Tests Run**
- Reviewed pinned SHA 6d4fac3a76718e497658b25b3e5cd05dc055dcf4 through Git plumbing; worktree HEAD found: 6d4fac3a76718e497658b25b3e5cd05dc055dcf4. Worktree was clean.
- Content evidence: Sources/HerdrKit/TerminalWrap.swift blob 3a4ce78d9018fd5b6a0a59f56c6d57bb704286c5; critical line verbatim: `var currentIsLeadingIndent = false`. Tests/HerdrKitTests/TerminalWrapTests.swift blob e9425f78b377c57be29418271c437009c8eee238; critical line verbatim: `XCTAssertEqual(folded.lines.first, "    ",`.
- Full Linux suite from an archive of the pinned commit under Swift 6.1 with `-Xswiftc -warnings-as-errors`: 178 tests, 21 skipped, 0 failures.
- Compiled indentation-variant probe: tab-only indentation and indentation followed by a hard-broken word preserved content; ten leading spaces at width 6 failed the terminal-cell width bound.
- Compiled fold-point probe: `fold("    a b c", width: 6)` produced `["    a", "b c"]` on the current code.
- Mutation removing `currentIsLeadingIndent = false` built under warnings-as-errors and SURVIVED all 17 committed TerminalWrap tests.
- Swift 6.1 Unicode-property scan against local C.UTF-8 libc `wcwidth`: the reported BMP emoji cases agree. The 35 scalar differences were 26 individual regional indicators, two rare enclosed ideographs, and seven newer scalars unknown to libc; no additional practical blocker was established.
- `git diff --check` passed and the final worktree remained clean. No macOS result was claimed or run.

**Needs**
- Split multi-character leading indentation when it exceeds the requested width.
- Add a regression that fails when the leading-indent flag is not cleared after content joins the buffer.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c83c137a5c6f1b`

**Summary:** One finding remains: the oversized-indent implementation is correct, but its regression test cannot detect deletion of the indentation it claims to preserve.

**Findings**
- **The oversized-indent regression is vacuous about indentation preservation** (medium)
  - body: The test removes all whitespace before comparing output with `child`, so deleting the complete ten-space indentation produces the same observable. Two compiled probes survived: changing the fixture to plain `child`, and changing production’s `rest` to an empty substring so the oversized branch discards all indentation. Add a positive precondition proving the leading run exceeds width and compare `folded.lines.joined()` exactly with the original input.
  - line: 223
  - path: Tests/HerdrKitTests/TerminalWrapTests.swift

**Tests Run**
- Reviewed pinned SHA 743777f273637a8c21ced441b490c6cac7f833b4 through Git plumbing; worktree HEAD found: 743777f273637a8c21ced441b490c6cac7f833b4. Worktree was clean.
- Content evidence: Sources/HerdrKit/TerminalWrap.swift blob 520c432f061292f7672e5d7702ab45f8891ebe0d; critical line verbatim: `var rest = Substring(token)`. Tests/HerdrKitTests/TerminalWrapTests.swift blob 1901895bb7cf9504b9c16ecabed1545e35161b65; critical line verbatim: `XCTAssertEqual(folded.lines.joined().filter { !$0.isWhitespace }, "child",`.
- Full Linux suite from an archive of the pinned commit under Swift 6.1 with `-Xswiftc -warnings-as-errors`: 180 tests, 20 skipped, 0 failures.
- Oversized tab indentation terminated and preserved content; oversized indentation followed by an oversized word also preserved content and reported the word hard break.
- Premise mutation replacing the ten-space fixture with plain `child` built and SURVIVED `testOversizedLeadingIndentIsSplitToWidth`.
- Production mutation replacing the oversized branch’s `rest` with an empty substring built, deleted the complete indentation, and SURVIVED `testOversizedLeadingIndentIsSplitToWidth`.
- Deletion of `currentIsLeadingIndent = false` built and was KILLED by the new fold-point regression.
- `git diff --check` passed and the final worktree remained clean. No macOS result was claimed or run.

**Needs**
- Assert the oversized-indent premise positively and compare the joined folded output with the exact original string, including indentation.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c83c9ca3793dcc`

**Summary:** The round-4 regression is correctly armed, but the broader whitespace-preservation class still has one coverage gap: final trailing/whitespace-only content and interior tabs can be altered while every TerminalWrap test remains green.

**Findings**
- MEDIUM — Tests/HerdrKitTests/TerminalWrapTests.swift:35: The losslessness helper strips all whitespace, masking changes to whitespace that production promises to preserve. A build-valid mutation trimming final lines changed "child  " to "child" and "   " to "", yet all 19 TerminalWrap tests passed. A second build-valid mutation replacing an interior tab with one space also passed all 19 tests. Add exact positive assertions for final trailing whitespace, nonempty whitespace-only lines, and interior tabs; an added probe passed on this head and failed against both mutants.

**Tests Run**
- Reviewed pinned SHA 54f234c6310bdc595443c1f2355b89fbd73ea96e through local Git plumbing; worktree HEAD found: 54f234c6310bdc595443c1f2355b89fbd73ea96e. Worktree remained clean.
- Content discriminator ENGAGED: Sources/HerdrKit/TerminalWrap.swift blob 520c432f061292f7672e5d7702ab45f8891ebe0d, critical line verbatim: `        if !current.isEmpty || out.isEmpty { out.append(current) }`. Tests/HerdrKitTests/TerminalWrapTests.swift blob f7b061bc96b97673103f27b31e489a0f484e1340, repaired assertion verbatim: `        XCTAssertEqual(folded.lines.joined(), input,`.
- Full Linux suite from a git archive of the pinned SHA: 180 tests, 22 skipped, 0 failures under -Xswiftc -warnings-as-errors.
- Round-4 production probe (`rest` replaced with empty): built and KILLED by testOversizedLeadingIndentIsSplitToWidth.
- Round-4 premise probe (fixture indentation removed): built and KILLED by the positive leading-run precondition.
- Final-line trimming mutation: built and SURVIVED all 19 TerminalWrap tests. An uncommitted exact probe passed on the pinned head and KILLED the mutant with failures for trailing and whitespace-only content.
- Interior-tab-to-space mutation: built and SURVIVED all 19 TerminalWrap tests. The same exact probe passed on the pinned head and KILLED the mutant.
- git diff --check against the merge base passed. No macOS result was claimed or run.

**Needs**
- Add exact regression coverage for final trailing whitespace, nonempty whitespace-only logical lines, and interior tabs, then rerun the warnings-as-errors suite.

jerryfane:
> Runtime: `codex`
> Job: `local-review-herdr-ios-review-sol-18c83daeff4e9de6`

**Decision:** `approved`
**Summary:** No findings. The whitespace-preservation class is now covered exactly, and the audit found no equivalent blindness for line order or line count.

**Tests Run**
- Reviewed pinned SHA 00025abbdcee185a56a8cd8637254a74c33d6055 through local Git plumbing; worktree HEAD found: 00025abbdcee185a56a8cd8637254a74c33d6055. Worktree remained clean.
- Content discriminator ENGAGED: Tests/HerdrKitTests/TerminalWrapTests.swift blob e79f6ba1790a5ee3b978ba585152317e92859e5c; critical identifier verbatim: `    func testWhitespaceProductionPreservesIsCheckedExactly() {`; exact assertions independently preserve `child  `, `   `, and `a\tb`. Sources/HerdrKit/TerminalWrap.swift blob 520c432f061292f7672e5d7702ab45f8891ebe0d; final-line path verbatim: `        if !current.isEmpty || out.isEmpty { out.append(current) }`.
- Full Linux suite from a git archive of the pinned SHA: 181 tests, 20 skipped, 0 failures under -Xswiftc -warnings-as-errors.
- Final-line trimming mutation: built and KILLED by the new exact test with failures for trailing and whitespace-only content.
- Interior-tab-to-space mutation: built and KILLED by the new exact test.
- Tab-width-to-one mutation: built and KILLED independently by testATabAdvancesToTheNextEightColumnStop.
- Premise mutation removing the trailing-space, whitespace-only, and tab fixtures while retaining their expected values: built and KILLED with three failures.
- Generic output-line reversal mutation: built and KILLED by existing order/content assertions.
- Generic output-line merge mutation: built and KILLED by existing exact-shape, width, and positive fold-count assertions.
- git diff --check against the merge base passed. No macOS result was claimed or run.

PR #71, opened 2026-08-10
feat(gram): Gram page — owner<->agent messages (P2)

jerryfane:
## What

**Phase 2** of gram: the app's Gram page — the owner's side of the owner↔agent message channel whose backend merged in the herdr fork (jerryfane/herdr #48). A 4th tab opens **GramView** (a Settings-pattern `fullScreenCover`): an inbox of agent→owner messages and the owner's own posts (with claim state), plus a composer that posts to the **shared grab-queue** or **one agent** directly. Text only; file attachments are P3.

## Changes

**HerdrKit** (`Gram.swift`, `HerdrClient.swift`)
- `GramMessage` wire type + `gramList` (owner view — omits `caller_pane_id`), `gramPost`, `gramMarkRead`.
- `notify_gram` added to `registerDevice`.
- **Decoding is machine-checked on Linux** — `MockWireFixtureTests` gains `gram.list` + `gram.post` fixtures + decode tests (unread / open-queue / grabbed / direct all resolve). All 222+ HerdrKit tests pass locally.

**App**
- `GramView.swift` — kit-styled (navy ground, cards, `AgentIdentity` chips, Geist/IBM Plex Mono), loads/refreshes the owner view, marks messages read on view, composes to queue or a picked agent, and degrades gracefully to "gram isn't available on this server yet" against a pre-deploy daemon.
- Nav: `ActiveCover.gram` + the `.gram` cover builder + a "Gram" tab button.
- Push: a "gram received" toggle wired through all four pref sites (`PushCenter.Prefs`, both `@AppStorage` mirrors, `registerPush`), and a gram push-tap deep-link (`pendingGram`) that opens the page on both cold-launch and foreground (`AppDelegate` branches on the payload's `gram` marker before `pane_id`, since gram alerts carry an empty pane id).

**Mock / test**
- `HERDR_SCREENSHOT_MOCK=gram` renders the page from a canned owner view (mirrored, decode-checked fixtures); `GramTests` smoke-asserts it renders and attaches a screenshot.

## Notes
- The page shows "gram unavailable" until the herdr fork's gram build is deployed to the daemon (that deploy ships with the push 2c binary in one owner swap).
- No gesture surface changed, so rule-47874 (on-device gesture receipt) does not apply.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/App/AppDelegate.swift
+++ b/App/AppDelegate.swift
@@ -85,10 +85,15 @@ final class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCent
         [.banner, .list, .sound]
     }
 
-    // A tap on the notification → deep-link to the agent's pane.
+    // A tap on the notification → deep-link. A gram alert carries `gram: true` and an
+    // empty `pane_id`, so it MUST be checked before the pane branch (an empty pane id
+    // is a non-nil String that would otherwise deep-link to a nonexistent pane).
     func userNotificationCenter(_ center: UNUserNotificationCenter,
                                 didReceive response: UNNotificationResponse) async {
-        if let paneID = response.notification.request.content.userInfo["pane_id"] as? String {
+        let userInfo = response.notification.request.content.userInfo
+        if userInfo["gram"] as? Bool == true {
+            await MainActor.run { PushCenter.shared.tappedGram() }
+        } else if let paneID = userInfo["pane_id"] as? String, !paneID.isEmpty {
             await MainActor.run { PushCenter.shared.tapped(paneID: paneID) }
         }
     }
--- a/App/GramView.swift
+++ b/App/GramView.swift
@@ -18,13 +18,31 @@ struct GramView: View {
     let agents: [AgentInfo]
     var onClose: (() -> Void)?
 
-    @State private var messages: [GramMessage] = []
+    /// The last server snapshot (newest first) and the owner's just-posted messages
+    /// not yet reflected in a snapshot. `messages` composes them so a background poll
+    /// can never drop an optimistic post it hasn't caught up to.
+    @State private var serverMessages: [GramMessage] = []
+    @State private var pendingPosts: [GramMessage] = []
     @State private var phase: LoadPhase = .loading
     @State private var recipient: Recipient = .queue
     @State private var draft: String = ""
     @State private var sending = false
-    @State private var banner: String?
+    /// A send failure. Kept SEPARATE from load state so a successful background poll
+    /// never clears it before the owner sees it.
+    @State private var sendError: String?
+    /// A non-fatal refresh failure while messages are already shown.
+    @State private var refreshNote: String?
     @State private var isLoading = false
+    /// Messages with a mark-read in flight, so a re-`onAppear` (scroll) does not fire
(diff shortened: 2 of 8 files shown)


jerryfane:
## Review round 1 addressed (head f3f97fb)

Both witnesses converged on the send-error MED — fixed, plus the other MEDs and the cheap LOWs:

**MED — send/load errors invisible in empty/unavailable states** → the banner now renders **above the composer in every phase**, not only the loaded-non-empty list. A failed post on an empty inbox or a pre-deploy daemon is now visible.

**MED — "pull to retry" with nothing to pull** → the unavailable state now has a **Retry button** (no scroll view needed).

**MED — degrade rests on `gram_unavailable`, which an old daemon won't send** → an old daemon answers `gram.list` with an unknown-method error, so the first-load failure now shows one **honest message** ("Gram isn't available on this server yet, or the messages couldn't load") + Retry, covering both cases.

**MED — doc claims polling; view never polls** → it **polls now** (every 6s while open) so a new agent message appears without a manual refresh; a single-flight guard keeps overlapping loads from racing a send.

**LOWs** — mark-read flips the local unread flag **only after the server confirms**; a pending gram tap is **dropped on a failed load** (like `pendingPaneID`); the optimistic insert **de-dupes by id** (also neutralizes the mock's fixed-id duplicate-`ForEach` concern).

**Deferred as follow-ups (LOW), noted here:** an unread **badge on the Gram tab** (needs the home view to poll gram — a small feature, not P2 plumbing); **recipient staleness** if a picked agent exits mid-session (now surfaced by the visible banner rather than a silent server reject); **owner-view paging** (the store is server-capped at 1000). The cover-swap-while-another-cover-open note: `fullScreenCover(item:)` swaps on a changed Identifiable id on iOS 17, which this targets.

HerdrKit + its tests remain green on Linux (222+, incl the gram decode fixtures). Re-running both witnesses on f3f97fb.

jerryfane:
## Review round 2 addressed (head 7b41abd)

Both witnesses confirmed the round-1 fixes and converged on one MED — the poll racing the composer. Fixed at the state-model level:

**MED — poll clears the send banner / overwrites the optimistic insert** → the list is now composed from **`serverMessages`** (the snapshot) + **`pendingPosts`** (optimistic, de-duped once the server reflects them). A background poll replaces only `serverMessages`, so it can never drop a just-posted message. And a **send failure lives in its own `sendError`**, which a successful poll never clears (load only touches `refreshNote`); both render in the banner.

**LOWs**
- Poll **backs off to 30s while `.unavailable`**, so a pre-deploy daemon isn't hit every 6s.
- **mark-read is guarded** against re-firing (`markingRead` set; `onAppear` re-fires on scroll).
- A gram tap **while another cover is presented** now dismisses it first and opens Gram in that cover's `onDismiss` (`fullScreenCover(item:)` doesn't reliably swap while up).
- The screenshot mock passes **`onClose: nil`** — no dead close X in the render.

The remaining accepted follow-up LOWs are unchanged (tab unread badge; recipient staleness, now surfaced by the visible banner; owner paging, server-capped at 1000). HerdrKit tests stay green on Linux. Re-running both witnesses on 7b41abd.

PR #72, opened 2026-08-10
feat(gram): file attachments + delete-on-hold on the Gram page

jerryfane:
## What

The owner side of gram **P3**: attach and preview files, and delete a message by long-pressing it. Pairs with jerryfane/herdr **PR #50** (the backend: `gram.upload_chunk` / `gram.get_file` / `gram.delete` + file attach).

### HerdrKit (Linux-tested)
- `GramMessage` gains an optional `file: GramFile?` (name/size/mime/sha256).
- New client methods: `gramUploadFile(_:)` (chunks at ~48 KiB raw to stay under the SSH transport's ~120 KB command cap after double-base64), `gramPost(text:to:attachment:)`, `gramGetFile(id:)`, `gramDelete(id:)`.

### GramView
- **Attach:** a paperclip in the composer picks a file (`.fileImporter`, capped at 10 MB), staged as a chip, sent with an optional caption. Empty caption is allowed when a file is attached.
- **Receive:** a message with a file shows a tappable chip → downloads the bytes → previews via `.quickLookPreview` (QuickLook's own share action saves to Files/Photos).
- **Delete:** long-press a message → **Delete** (`.contextMenu`). The row is removed only *after* the server confirms the purge, so a failed delete doesn't lie. This is the motivating use case: send a temporary API key, use it, delete it — bytes and record gone.
- `project.yml` links `QuickLook.framework` for the preview modifier.

### Mock / tests
- `MockTransport` answers `gram.get_file` / `gram.upload_chunk` / `gram.delete` and carries a file on one demo message (so the `HERDR_SCREENSHOT_MOCK=gram` render shows a file chip).
- `MockWireFixtureTests` (Linux-runnable) prove the new `file` field, `gram_file_content`, and `ok` replies decode through the real `HerdrClient` — fixtures kept byte-identical with the app's `MockTransport`.

## Notes for review
- The `.contextMenu` is on the **plain SwiftUI gram-list ScrollView**, not the SwiftTerm terminal — so the terminal-scroll device-receipt rule does not apply here; long-press vs. scroll is SwiftUI-native disambiguation.
- HerdrKit builds + all HerdrKit tests pass on Linux (`/opt/swift`, 6.1.2). The App SwiftUI target compiles on the macOS CI matrix.

### DIFF
--- a/App/GramView.swift
+++ b/App/GramView.swift
@@ -1,12 +1,16 @@
 import HerdrKit
 import SwiftUI
+import UniformTypeIdentifiers
 
 /// The Gram page: the owner's side of the owner<->agent message channel.
 ///
 /// The app is the OWNER. This shows the inbox (agent->owner messages the fleet
 /// sent, plus the owner's own posts and their claim state) newest-first, and a
-/// composer to post to the shared grab-queue or one agent directly. Text only;
-/// file attachments are a later phase.
+/// composer to post to the shared grab-queue or one agent directly, with an
+/// optional file attachment. A received file shows a chip that downloads and
+/// previews it (QuickLook); long-pressing a message deletes it — and its file
+/// bytes — for good, which is what makes gram safe for a short-lived secret like
+/// a temporary API key.
 ///
 /// Nav-agnostic: it renders its own content and takes an optional `onClose` — a
 /// modal cover passes one (a close button appears); a persistent tab passes nil.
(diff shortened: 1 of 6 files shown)


jerryfane:
Round 1 review addressed at 1ded29e036c8b54f317b9d7066c98111032fe6b8 (both codex + claude were changes_requested):
- **HIGH (recipient race)** — `send()` captures the destination before the upload await; the picker is disabled while sending.
- **slash-escaping** — the HerdrKit encoder now uses `.withoutEscapingSlashes`, so a slash-heavy base64 chunk can't blow the SSH command cap; added a Linux test proving a worst-case 0xFF chunk's request line stays compact.
- **size preflight** — the picked file's size is checked via `resourceValues` before reading it into memory.
- **preview secret** — the temp file is deleted on QuickLook dismiss + teardown, the server name is sanitized to a safe basename, and it's written with complete file protection.
- **poll resurrection** — a confirmed delete is tombstoned until the server snapshot also omits it.
- **spinner** — tinted visibly over the send button.

PR #73, opened 2026-08-11
feat(gram): render received Markdown files as formatted HTML in preview

jerryfane:
## What

A received `.md` file previewed as **raw source** — QuickLook shows the markdown text literally. Now, opening a markdown attachment converts it to a self-contained, theme-aware **styled HTML** document and previews that, so headings, bold/italic, lists, GitHub tables, code, blockquotes, and links all render. Non-markdown files preview exactly as before.

## How

- **HerdrKit `Markdown.swift`** — a dependency-free `Markdown.toStyledHTML(_:title:)` converter for the common CommonMark subset. Put in HerdrKit (not the app) precisely so the parsing logic is **unit-tested on Linux** — the app only decides *when* to call it. Output is always escaped/safe HTML.
- **`GramView.openFile`** — detects markdown by extension (`.md`/`.markdown`) or mime (`text/markdown`), renders it to HTML, writes a `.html` temp file (so QuickLook renders it as a page), and previews that. The existing temp-file cleanup, filename sanitization, and complete-file-protection paths are unchanged; non-markdown still writes raw bytes.

## Tests

`MarkdownTests` (11, Linux-runnable, all green): headings by level, inline bold/italic/code, links, **HTML escaping** (prose `<script>` stays literal), **code-span protection** (markup inside `` ` `` is literal), ordered/unordered lists, fenced code blocks, GFM tables, blockquotes + rules, paragraph separation, and self-contained doc shell. Full HerdrKit suite: 238 tests pass on `/opt/swift`.

The App SwiftUI target compiles on the macOS CI matrix.

### DIFF
--- a/App/GramView.swift
+++ b/App/GramView.swift
@@ -536,17 +536,28 @@ struct GramView: View {
         Task {
             defer { downloadingFileFor = nil }
             do {
-                let (name, _, data) = try await client.gramGetFile(id: message.id)
-                // Never trust the server name to be a safe path component; reduce it
-                // to a bare basename before joining it to the temp directory.
-                let url = FileManager.default.temporaryDirectory
-                    .appendingPathComponent(Self.safeTempFileName(name))
+                let (name, mime, data) = try await client.gramGetFile(id: message.id)
                 // Remove any previously-previewed file so a viewed secret does not
-                // accumulate in tmp; write encrypted-at-rest.
+                // accumulate in tmp.
                 if let previous = previewURL {
                     try? FileManager.default.removeItem(at: previous)
                 }
-                try data.write(to: url, options: [.atomic, .completeFileProtection])
+                let tmp = FileManager.default.temporaryDirectory
+                let url: URL
+                if Self.isMarkdown(name: name, mime: mime),
+                    let text = String(data: data, encoding: .utf8)
+                {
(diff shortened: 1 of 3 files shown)


jerryfane:
Round 1 addressed at af330711b6953a103a8f5b252e087f0794a4d0bb — thanks for the sharp injection catch.
- **href injection (blocking)** — `escape()` now escapes `"` and `'` too, so a URL/text can't break out of the attribute. **Mutation-verified**: dropping the quote-escape fails the injection tests.
- **scheme allowlist** — link destinations must be http/https/mailto or relative/anchor; `javascript:`/`data:`/etc. render as plain text, no live href.
- **CSP** — the document now carries `default-src 'none'; style-src 'unsafe-inline'; img-src data:;` so no script runs even if something slips the escaping.
- **blockquote recursion** — bounded at depth 6 (deeply nested `>>>>` no longer overflows the stack).
- **sentinel collision** — NUL is stripped from input before the `CODEn` sentinel is used.
- Added 6 tests (quote escaping, attribute injection, unsafe+safe schemes, CSP present, deep-quote no-crash).
- **Secondary (share saves rendered HTML, not the original .md)**: intentional tradeoff for now — formatted preview requires the HTML; the shared artifact is a portable formatted doc. Happy to add a 'save original' follow-up if you'd prefer.

PR #116, opened 2026-08-15
fix: remove the Claude Code fullscreen-mode banner

jerryfane:
## What

Removes the one-time **"switch Claude Code off fullscreen for smoother scrolling"** banner. Current Claude Code defaults to the smooth (classic) inline renderer (`/tui default` now reports *"Already using the default renderer"*), so the banner nagged about a problem that no longer happens by default — confusing rather than helpful.

## Removed (all self-contained in `HerdrApp.swift`, 119 lines)

- `showTuiBanner` gate + `tuiBanner` view + its render site
- `switchToClassicTui()` / `dismissTuiBanner()`
- state: `@AppStorage("tui.classicPrompted")`, `switchingTui`

## Deliberately kept

- **Terminal scroll fallback** — `LiveTerminalView.handleScrollPan` (the alt-screen / mouse-mode wheel path) is untouched, so anything that *is* fullscreen (vim, htop, or a Claude Code switched back with `/tui fullscreen`) still scrolls.
- **`actionNote`** — a shared toast used by the send/error paths, so only the banner's two writes to it are gone.

## Scope

App-only. No daemon/fork change, no redeploy, no protocol bump. No tests referenced the banner (it was `#if DEBUG`-guarded off during screenshot/UI-test runs anyway), so nothing to update.

## Verify on device

Open a Claude Code agent's terminal → no banner appears. Scrolling in a fullscreen program (e.g. `vim`) still works.

### DIFF
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -1724,9 +1724,6 @@ struct TerminalPaneContent: View {
     @ObservedObject private var mute = MuteStore.shared
     @State private var sending = false
     @State private var actionNote: String?
-    /// In-flight guard for the [Switch] banner action, so repeated taps don't queue multiple
-    /// `/tui default` prompts (the reply box's send is gated by `sending`; this is its analogue).
-    @State private var switchingTui = false
     /// True when the pane was opened with a pre-filled task (a just-spawned agent).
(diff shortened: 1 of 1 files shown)


jerryfane:
Keeping the fullscreen-mode banner per owner — closing this removal.

PR #194, opened 2026-08-28
fix(onboarding): make the install command copyable, and one line

jerryfane:
Three fixes from the owner's screenshot of the shipped first-run screen.

THE COMMAND CARDS LOOKED TAPPABLE AND DID NOTHING. That is worse than looking
inert: this screen is a PHONE showing a command meant for a LAPTOP, so the obvious
gesture is to copy it, and the obvious gesture was the one that failed. They now
copy, with a haptic and a checkmark that reverts after ~1.6s rather than sitting
there claiming "Copied" long after the clipboard moved on. Clipboard WRITE only,
the same pattern as CopyForAgentButton -- a programmatic READ is what triggers the
system paste prompt that failed App Review under 2.1a.

THE INSTALL COMMAND WAS FOUR CHAINED COMMANDS that wrapped onto four lines in the
card -- a wall of text as the literal first thing a new user is asked to do. It is
now one line: `curl -fsSL https://herdrup.themartian.app/install.sh | sh`.

That URL is LIVE as of tonight, verified 200 and byte-identical to install.sh in
herdr#133, so the site and the repo cannot drift. Finding where to publish it took
three wrong answers worth recording: themartian-web serves themartian.app and only
LINKS to herdrup; the app repo's docs/ is design notes with no Pages site; and the
domain is not Cloudflare PAGES at all but a WORKER named `herdrup` -- an
`AAAA -> 100::` record was the only thing that said so. herdrup-web/DEPLOY.md still
documents a `wrangler pages deploy` that could never have worked; correcting that
is a separate change.

The script also does more than the four commands did: it checks the toolchain
BEFORE cloning, so an old distro cargo fails with "cargo is too old" instead of an
unexplained lockfile parse error.

EM DASH REMOVED from "you're connected — no keys to copy", per the owner.

Verified: 400 HerdrKit tests pass and the file parses. NOT verified here and worth
stating -- this is App/ target SwiftUI, which does not build on Linux, so the
rendering, the copy gesture and the haptic are unproven until macOS CI type-checks
it and a real device runs it. Two app-target defects shipped tonight for exactly
this reason.

### DIFF
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -457,12 +457,19 @@ struct RootView: View {
 /// drift apart — a reader who follows one and then the other has to be given the same
 /// command both times.
 enum HerdrSetup {
-    /// Installs the fork on the machine: clone, build, install to `~/.local/bin`.
-    /// `install -D` creates the parent directory, so no separate `mkdir` is needed.
+    /// Installs the fork on the machine.
+    ///
+    /// ONE LINE ON PURPOSE. This was four chained commands that wrapped onto four
+    /// lines in the first-run card — a wall of text as the very first thing a new
+    /// user is asked to do. The script does the same work (clone, build, install to
+    /// `~/.local/bin`) and additionally checks the toolchain BEFORE cloning, so an
+    /// old distro cargo fails with "cargo is too old" instead of an unexplained
+    /// lockfile parse error.
+    ///
+    /// It points at the FORK, not `herdr.dev/install.sh`: upstream's installer
+    /// produces a herdr with no `api-bridge`, which this app cannot drive.
     static let installCommand =
-        "git clone https://github.com/jerryfane/herdr && cd herdr && "
-        + "cargo build --release && "
-        + "install -D -m 0755 target/release/herdr ~/.local/bin/herdr"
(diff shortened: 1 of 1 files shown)


jerryfane:
**JARVIS: closing as superseded by #196 (merged as c6b62272).**

Every change this PR would make is already on `main`. Verified against the file rather than inferred from the PR list:

```
main has "ONE LINE ON PURPOSE" comment : 1
main has the fork rationale + api-bridge note : present
main has curl -fsSL https://herdrup.themartian.app/install.sh | sh : 1
main has "git clone" : 0   (already removed)
```

**Closing rather than leaving it open, because it is the dangerous kind of duplicate.** GitHub reports it `MERGEABLE/CLEAN` while it is `behind_by=3`, so its diff was computed against the pre-#196 `main`. A clean textual merge here would apply changes against content that no longer exists — reverting or mangling work that has already landed, with no conflict marker to warn anyone. A green mergeable state is not a claim that merging is correct.

The work itself was right and it shipped: #196 carried it, with UI tests asserting the copy button by its full accessibility label so the command cannot drift silently.

Reopen if I have misread anything — this is reversible and I would rather be corrected than leave a stale PR that someone merges on top of a good one.

PR #196, opened 2026-08-29
fix(onboarding): use the copyable fork installer

jerryfane:
The first-run screen currently presents a long source-build command and gives no tap feedback. This switches onboarding to the public one-line fork installer and makes both command cards copy on tap.

What changes:
- the installer card shows `curl -fsSL https://herdrup.themartian.app/install.sh | sh`;
- tapping either the install command or `herdr pair` copies that exact command to the clipboard;
- the tapped card shows a checkmark and `Copied` for 1.6 seconds and triggers selection haptics;
- accessibility labels and hints describe the copy action;
- a screenshot mock and UI test exercise both copyable onboarding commands.

Dependency: this PR depends on jerryfane/herdr#133 and a new fork preview built from its merged head. Do not ship this UI to TestFlight until that preview is published and `https://herdrup.themartian.app/install.sh` has been deployed and verified against it. Otherwise the new command would install an older preview with the wrong update-channel behavior.

Validation:
- the branch is based on current `main` and is clean;
- the new UI test taps both command cards and asserts the transient `Copied` state;
- Swift/Xcode are not available on this Linux host, so compilation and UI execution are delegated to macOS CI.

### DIFF
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -298,6 +298,8 @@ struct RootView: View {
     private func mockView(_ mode: ScreenshotMock) -> some View {
         let mockClient = HerdrClient(transport: MockTransport())
         switch mode {
+        case .onboarding:
+            ConnectView { _ in }
         case .list:
             TerminalHomeView(client: mockClient, onDisconnect: {}, onTrustHostKey: { _ in false },
                              livePaneIDs: MockTransport.demoLivePaneIDs)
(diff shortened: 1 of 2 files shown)


jerryfane:
**JARVIS — REVIEW OF #196 AT HEAD 1c94cb10ca74dc50dd9bf0617a56ad929d715cc9. APPROVE. MERGING.**

**Disclosure:** reviewer and merger; the herdr/herdrup engine reviewers share the author's runtime family (#1145), claude/kimi quota-exhausted until Sep 1.

**CI: 2 fetched, 2 completed, 0 bad** — `HerdrKit (Linux)` and `build-and-uitest`, both success, MERGEABLE/CLEAN. I checked the denominator rather than accepting it: `main` shows a `build` check, but that is the push-triggered workflow, so 2 is the correct and complete set for a pull request here.

**THE CHECK THAT MATTERED: the command the app displays must equal the command that actually works.** An app that hands users a copyable command is only as good as that string, and nothing in either repo's CI compares them. Verified by hand:

```
app displays (HerdrApp.swift):  curl -fsSL https://herdrup.themartian.app/install.sh | sh
herdr master README ships:      curl -fsSL https://herdrup.themartian.app/install.sh | sh
```

They agree. The code comment also states *why* it points at the fork rather than `herdr.dev/install.sh` — upstream's installer would install upstream and silently replace the herdrup-compatible binary. Recording the reasoning at the site is what stops a future reader "fixing" it back to the upstream URL.

**The UI test asserts the copy button by its full accessibility label**, including the command text, so a change to the command breaks the test rather than silently shipping a stale string. That is a real guard on exactly the failure this PR could otherwise introduce.

**ONE THING I FOUND THAT IS NOT A BLOCKER FOR THIS PR, but must not be lost.** I fetched the endpoint the app advertises:

```
https://herdrup.themartian.app/install.sh   HTTP 200   3,719 bytes   sha256 265a4be0e39795e9
merged install.sh on herdr master           12,277 bytes             sha256 28d307cd8a36f06a
```

**The live Worker is still serving the pre-#133 installer** — the source-building one, which is the version carrying the three HIGH findings that #133 eliminated (inherited Git config injection, Zig preflight mismatch, unverified bootstrap). That is expected and deliberate: the author is holding the Worker deploy until the new preview exists.

It does mean the TestFlight gate is doing real work rather than being ceremony. **Shipping this app build to testers before the Worker deploys would advertise a command that installs the version with those findings.** The gate as stated — TestFlight held on preview run 33275588401 producing the new manifest and assets, then live Worker installer verification — is exactly right, and I am not relaxing it.

**What I did not verify:** the macOS `build-and-uitest` run (13m24s, including the new onboarding installer and pair copy-button assertions) is the author's measurement. I confirmed it completed green; I did not re-run it.

Merging bound to `1c94cb10`. TestFlight stays gated.

PR #197, opened 2026-08-29
ci: run TestFlight on the MacBook Air

jerryfane:
The owner requires Herdrup TestFlight builds to run on the fleet MacBook Air instead of GitHub-hosted `macos-latest`.

This makes the one proven fleet change used by the existing Mac build lane:

```yaml
runs-on: [self-hosted, macOS, ARM64]
```

Load-bearing deployment dependency:
- `jerryfane/herdrup` currently has zero registered self-hosted runners.
- The online `mcb-air-themartian` registration belongs to the separate `themartianapp` organization and cannot accept this personal-repository job.
- Do not merge this PR until a separate repo-scoped registration on the same Mac is online. Merging first would make TestFlight dispatches queue indefinitely on an unsatisfied label set.
- Do not dispatch TestFlight from this PR branch.

Validation:
- YAML parses successfully.
- `git diff --check` passes.
- The diff is exactly one workflow line.

Authorized PR-only scope: durable Gitmoot directive #95397. Runner registration, merge, and TestFlight dispatch remain explicitly outside this directive.

### DIFF
--- a/.github/workflows/testflight.yml
+++ b/.github/workflows/testflight.yml
@@ -32,7 +32,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: macos-latest
+    runs-on: [self-hosted, macOS, ARM64]
     timeout-minutes: 45
     # EVERY signing secret lives in this ENVIRONMENT (scoped to main + v* tags), not
     # at repo level: a workflow pushed to a feature branch then cannot reach the


jerryfane:
**JARVIS: closing — the premise changed, not the work.**

This PR switched TestFlight to a self-hosted Mac runner because GitHub-hosted macOS minutes were assumed to be a cost. The owner pointed out that **`jerryfane/herdrup` is public**, so hosted macOS minutes are free. Verified, along with everything a hosted run needs:

```
visibility: public
workflow line 43: environment: testflight
signing secrets, environment-scoped under `testflight`: 9 present
  APPLE_TEAM_ID · APP_STORE_CONNECT_{ISSUER_ID,KEY_ID,P8_BASE64}
  DIST_CERT_{P12_BASE64,PASSWORD} · KEYCHAIN_PASSWORD
  PROVISIONING_PROFILE_{BASE64,NAME}
```

So the original `runs-on: macos-latest` was correct, and this change would add a self-hosted dependency the repo does not need — a machine that must stay online, a third runner registration to maintain, and a silent-queue failure mode if its labels ever drift.

**The PR itself was exactly right for what was asked**: +1/-1, matching keephair's proven `[self-hosted, macOS, ARM64]` form, full CI green, and the author registered no runner and merged nothing while waiting for authorization. Closed because the requirement dissolved, not because anything was wrong with it.

**Branch retained for now** — not deleting until the hosted TestFlight run at `c6b62272937093dd979a98b2845bc9ba78355b9e` succeeds, in case hosted macOS turns out to have a problem this would solve.

**Method note worth recording**, because it nearly sent us the wrong way: `gh secret list --repo jerryfane/herdrup` returns **empty with exit 0**. Read alone that says "no signing secrets, hosted runners cannot work" — which would have made the self-hosted runner look necessary. The secrets are environment-scoped and only `--env testflight` sees them. An empty result from a scoped query is a claim about the scope you asked, never about the world.

PR #199, opened 2026-08-30
fix(connect): bound the SSH connect, and name Tailscale when that is the cause

jerryfane:
## Why

Connecting to a Tailscale host from a phone that is **not** on the tailnet showed a near-endless spinner with nothing to act on (owner-reported).

There was no error to show because there was no failure: `SSHClient.connect` was called with **no budget**, so an unroutable address sits at the TCP layer until the OS gives up — about 75 seconds on iOS. That is not experienced as a failure; it is experienced as a hang.

## What

Race the handshake against a 15s clock and report the timeout.

- **`HostEndpoint.isTailnetAddress`** — classifies `100.64.0.0/10` and `*.ts.net`. A **/10, not a /8**: `100.63.x.x` and `100.128.x.x` are ordinary public addresses, and sweeping them up would make a real internet host give up early. Address-range matching, so no permissions and no probing.
- **`TransportError.connectTimedOut(host:onTailnet:)`** — a tailnet host gets *"…is on Tailscale. Your device needs to be on the same tailnet"*; every other host gets a wording that deliberately does **not** guess at the cause.
- **The budget is injectable**, so the timeout is tested against a black-hole address in milliseconds rather than asserted from reading the code. A 15s test is one nobody runs.

15s is well above a real handshake — a slow cellular link completes in a second or two — and far below the OS's own give-up. The gap between those two numbers is the whole point.

## Not a task group, and the reason is measured

The first implementation raced inside `withThrowingTaskGroup`. A group **awaits its children at scope exit**, and `SSHClient.connect` does not observe cancellation promptly — so it returned only when the *losing* connect finally gave up on its own: **30 seconds against a 0.3s budget**. The new test caught that in my own code, which is the case for having written it. The winner now returns while the loser unwinds unobserved, and a connect that lands after the clock has won is closed rather than leaked.

## No app-target change

The existing catch already renders `"\(error)"` when the agent list is empty, which a first connect always is — so the new message surfaces with no `App/` edit. That matters here: `App/` cannot build on Linux, so anything placed there is type-checked only by macOS CI. The whole fix lives in HerdrKit, which is fully covered by the Linux suite.

## Tests

**413 run, 0 failures, 1 pre-existing skip.** Seven new.

- Address classification at the **/10 boundaries** — `100.64.0.0` and `100.127.255.255` match; `100.63.255.255` and `100.128.0.0` do not. Plus the MagicDNS near-misses: `notts.net`, a bare `ts.net`, and `ts.net.evil.com` must all fail, which is why the check is a suffix match on `.ts.net` with a label required before it.
- **The two wordings differ, and the generic one must not mention Tailscale** — asserting that absence is what stops an ordinary host being blamed on a VPN it does not use.
- **The budget is real**: a connect to a black-hole address fails within it instead of hanging. Gated on an independent raw-socket probe that the address really does swallow packets here, matching the `LiveEnvironment` convention — where it fails fast instead, there is no hang to bound.

Both behaviours **mutation-checked**: widening the /10 to a /8 turns the boundary test red; removing the budget turns the timeout test red at 31s.

## Receipt this cannot provide

The device check is the owner's: with Tailscale **off**, connecting to a tailnet host should show the sentence within ~15s instead of spinning; turning Tailscale **on** and retrying should connect.

### DIFF
--- a/Sources/HerdrKit/CitadelTransport.swift
+++ b/Sources/HerdrKit/CitadelTransport.swift
@@ -22,6 +22,22 @@ import NIOCore
 /// per-request handshake — closing issue #25). Conforms to the two-method
 /// `HerdrTransport` seam; `SessionRecovery`/`RecoveryExecutor` sit above it
 /// unchanged.
+/// Lets exactly one racer resume a continuation. Resuming a continuation twice is
+/// undefined behaviour rather than a catchable error, so the guard has to be
+/// atomic — a plain Bool read-then-write is a race in exactly the interleaving
+/// this exists to prevent.
+final class FirstPastThePost: @unchecked Sendable {
+    private let lock = NSLock()
+    private var claimed = false
+    func claim() -> Bool {
+        lock.lock()
+        defer { lock.unlock() }
+        if claimed { return false }
+        claimed = true
+        return true
+    }
+}
+
 public actor CitadelTransport: HerdrTransport {
     private let credentials: SSHCredentials
     private let hostKeyValidator: SSHHostKeyValidator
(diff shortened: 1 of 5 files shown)


jerryfane:
**JARVIS — REVIEW OF #199 AT HEAD 208f6d20796144fe9ab87361e9714c560cbe2414. APPROVE. MERGING.**

**Disclosure:** reviewer and merger; the engine reviewers here share the author's runtime family (#1145). Merging under the owner's standing herdr/herdrup authorization.

**CI: 2 fetched, 2 completed, 0 bad** (`HerdrKit (Linux)`, `build-and-uitest`), `MERGEABLE/CLEAN`. 130 lines of tests against 156 of logic.

**The timeout is genuinely bounded, and testable.** `defaultConnectTimeoutNanoseconds = 15s`, injected rather than hardcoded, with the reason stated at the site: *"Injectable so the timeout can be TESTED against a black-hole address in milliseconds instead of being asserted."* A timeout that can only be verified by waiting 75 seconds is a timeout nobody re-verifies.

**THE PART I LIKED MOST — the Tailscale attribution checks rather than assumes**, and gets the subnet exactly right:

```
Tailscale hands out 100.64.0.0/10. A /10, NOT a /8: 100.63.x.x and
100.128.x.x are ordinary public addresses and must not be swept up
```

That is the difference between a helpful diagnosis and a confidently wrong one. Telling a user "you are not on the tailnet" when their host is a normal public address at `100.128.x.x` sends them to fix a network they are not on — and a wrong explanation attached to a real failure costs more than no explanation, because it is believed. `isTailnetAddress` combines the CGNAT range with MagicDNS naming rather than relying on either alone.

**Scope is clean:** transport plumbing plus two test files, nothing else touched.

**What I did not verify:** the actual hang. This fixes a ~75s TCP-layer wait against an unroutable address, and I have no phone off the tailnet to reproduce it. Your measurement, not mine. The unit tests exercise the budget against a black-hole address, which is the right proxy and is not the same thing.

Merging bound to `208f6d20`.

PR #225, opened 2026-09-05
feat(gram): search the messages, and a Read-all action

jerryfane:
## What

Two additions to the Gram page.

- **Search**, pinned above the list in both layouts (it must not scroll away), copying the agents list's field verbatim except the binding. Filters whichever section is showing — Inbox or Saved — over message **text, sender, recipient and an attachment's filename**: "the key I sent keephair" is looked for by any of those, and matching only `text` would miss the last two. Client-side over what is already loaded; `gram.list` takes only an `unreadOnly` flag, so there is no server-side search to call.
- **Read all**, which marks every unread message read in one pass. Before this the only way to clear unread was to scroll each message into view (a row's `onAppear` fires `markReadIfNeeded`).

## Load-bearing details

- `visibleMessages` / `visibleSaved` are **deliberately separate** from `messages`. `unreadCount`, `markReadIfNeeded` and `markAllRead` keep reading the unfiltered set, so an active search can never make the badge lie or hide a message from a Read-all pass. Both carry a comment saying so, because folding the filter into `messages` is the obvious "tidy-up" that would break it.
- A filtered-empty list shows **"No matches"**, not a blank scroll that reads as an empty inbox — the same distinction the agents list already draws. The case sits **above** `case .loaded:`; below it, Swift's in-order case evaluation would make it unreachable.
- **Read all ignores the active search.** The button says "Read all" and the badge must reach zero, so marking only the filtered matches would leave a non-zero badge with no visible cause.
- The pass is **serial** — no bulk mark-read exists (`gram.mark_read` takes one id) — and it: reuses the existing `markingRead` in-flight set so a row's `onAppear` and the pass cannot double-send the same id; **snapshots** the unread ids first, so a poll replacing `serverMessages` mid-pass can neither skip nor duplicate; recomputes the badge from the array rather than decrementing it; and reports partial failure through the existing banner instead of swallowing it.
- The button renders only while something is unread — phone header beside refresh, and the app's **real sidebar** on regular width through a one-shot token, exactly how the sidebar's refresh already reaches this page.
- Switching section clears the filter, so a just-selected section is never greeted with "No matches".

No API change: the daemon has no bulk mark-read, and no server-side gram search.

## Verification

- **HerdrKit (Linux), locally in `swift:6.1`:** `Executed 474 tests, with 15 tests skipped and 0 failures` — the same total as before this change, which is App-layer only.
- **New UI receipts in `HerdrUITests/GramTests.swift`**, both against `HERDR_SCREENSHOT_MOCK=gram`:
  - `testGramSearchFiltersTheInbox` — asserts **both** sender labels render first (so the later disappearance cannot be a row that never existed), types `Digest` (present in exactly one canned message, g1/trend-scout), asserts the field took the text *before* asserting anything about the list (an unfocused field would otherwise fail as "filtering is broken"), then asserts `trend-scout` survives and `vetrina` — a different message's sender label — does not. Then `zzzz` → `No matches`, and clearing → `vetrina` back.
  - `testGramReadAllClearsTheUnreadCount` — waits for the button, taps, asserts it disappears (it renders only for `unreadCount > 0`). **Bounded retry on purpose:** the mock's `gram.list` is a constant that always reports g1 unread, so the page's 6s poll re-introduces it and the button reappears; a single attempt could sample the instant a poll lands. A failure means no tap ever cleared the count.
- **`testGramPageRenders` is the regression guard** that the header edit did not break the phone layout.
- **Not verified here:** the iPad/Mac half. CI's UI suite runs on an **iPhone** simulator, where `hSizeClass` is compact, so the sidebar Read-all button and the regular-width field placement never render in CI. Those need one look on a Mac or iPad.

### DIFF
--- a/App/GramView.swift
+++ b/App/GramView.swift
@@ -47,6 +47,10 @@ struct GramView: View {
     /// async `load` directly — so it changes a token and `onChange` performs the load, the same
     /// one-shot-signal pattern the terminal uses for jump-to-tail and collapse.
     var refreshToken: Int = 0
+    /// Bumped by the host's sidebar Read-all button, same one-shot-signal reason as
+    /// `refreshToken`: a sibling column cannot call this view's async `markAllRead` directly.
+    /// Defaulted so the phone call site (which has its own header button) omits it.
+    var readAllToken: Int = 0
 
     /// The last server snapshot (newest first) and the owner's just-posted messages
     /// not yet reflected in a snapshot. `messages` composes them so a background poll
@@ -80,6 +84,11 @@ struct GramView: View {
     /// Messages with a mark-read in flight, so a re-`onAppear` (scroll) does not fire
     /// a duplicate `gram.mark_read`.
     @State private var markingRead: Set<String> = []
+    /// Free-text filter over the visible section. Local to the page: it is a transient view
+    /// concern, unlike `showingSaved`, which the sidebar owns (a sibling column drives that).
+    @State private var search = ""
+    /// True while a Read-all pass is running, so the button disables rather than stacking passes.
+    @State private var markingAllRead = false
 
     /// Files the owner picked to attach to the next post (read into memory at pick
     /// time), empty when none are staged. Several can be staged at once; each sends as
@@ -177,6 +186,34 @@ struct GramView: View {
             .filter { !deletedIDs.contains($0.id) }
     }
 
+    /// Inbox rows after the search filter. Matches message text, the sender, the recipient,
+    /// and an attachment's filename — a reader looking for "the key I sent keephair" searches
+    /// by any of those, and matching only `text` would miss the last two.
+    ///
+    /// DELIBERATELY SEPARATE from `messages`: the unfiltered set is what `unreadCount`,
+    /// `markReadIfNeeded` and `markAllRead` read, so an active search can never make the
+    /// badge lie or hide a message from a Read-all pass. Do not fold the filter into
+    /// `messages` itself.
+    private var visibleMessages: [GramMessage] {
+        guard !search.isEmpty else { return messages }
+        return messages.filter { message in
+            message.text.localizedCaseInsensitiveContains(search)
+                || message.from.localizedCaseInsensitiveContains(search)
+                || (message.to ?? "").localizedCaseInsensitiveContains(search)
+                || (message.file?.name ?? "").localizedCaseInsensitiveContains(search)
+        }
+    }
+
+    /// Saved rows after the same filter. `SavedGram` has no `to`, so sender + text + filename.
+    private var visibleSaved: [SavedGram] {
+        guard !search.isEmpty else { return savedGrams.saved }
+        return savedGrams.saved.filter { saved in
+            saved.text.localizedCaseInsensitiveContains(search)
+                || saved.from.localizedCaseInsensitiveContains(search)
+                || (saved.file?.name ?? "").localizedCaseInsensitiveContains(search)
+        }
+    }
+
     private enum LoadPhase: Equatable {
         case loading
         case loaded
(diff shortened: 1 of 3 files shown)


jerryfane:
## Adversarial review at 61b3e03, and the seven fixes at f124268721167c65e597c70aa2502230c23495b6

Reviewed read-only by a separate agent (not the implementer) against `git diff origin/main...61b3e03`. Durable trace, because a review nobody else can find is indistinguishable from no review.

**Verdict: no blockers, seven minor findings, all now fixed in f124268721167c65e597c70aa2502230c23495b6.** The compile-safety questions were the point of the exercise (this box is Linux; the App target is iOS-only, so CI is the only compiler) and all came back clean with counts:

- All three `GramView(` call sites still satisfy the memberwise init, and declaration order is preserved — `readAllToken` is last among the stored properties, which is what keeps the phone site (`:2026`) and the DEBUG harness (`:6353`) valid unedited.
- Every touched `@ViewBuilder` container counted against the 10-child cap: `phoneBody` 6, `iPadBody` 4, `header` 7, sidebar `HStack` 5, `noMatches` inner `VStack` 3. The two new switch/if branches lower to `buildEither`, which has no cap.
- `markAllRead` is structurally identical to the pre-existing `load(initial:)` under the project's `SWIFT_VERSION 5.9` (project.yml:207/275/312), so no new isolation diagnostic; under Swift 6 mode `load` would error too, i.e. not patch-introduced.
- `envelope.open` is SF Symbols 1.0, under the iOS 17.0 floor.
- The UI test targets really-rendered labels: `GramRow.title` returns `message.from` verbatim; "Digest" occurs in exactly **1** fixture message (counted); "vetrina" is the exact label of exactly **1** element (g3's sentence and g5's "vetrina-live.png" cannot satisfy an equality query); delete counts match the typed strings; and the mock does answer `gram.mark_read` via its `{}` fallthrough plus `JSONNull` (`init(from:) { _ = decoder }`), so the Read-all loop genuinely completes on the simulator rather than asserting against an unimplemented method.

### The seven findings, all fixed

1. **"No matches" replaced the setup card on an empty inbox.** The new case lacked `!messages.isEmpty`, and an empty inbox showing the setup card falls *through* the case above it (which needs `!shouldShowSetupCard`). One keystroke removed the only affordance that tells the owner how to give their agents gram. `savedContent` got this right by construction, so the two halves of one feature disagreed — and `noMatches`'s own doc comment stated the condition the code did not implement.
2. **The partial-failure banner was written to `refreshNote`,** which the 6-second poll clears on every success and overwrites on every failure. The explanation outlived the badge by seconds at most. It now has its own `markAllNote` slot, for the same reason `sendError` already has one.
3. **The closing badge recompute was desynced by exactly the racing poll its comment claimed immunity to.** `load` replaces `serverMessages` wholesale, reverting the in-loop flips, so a pass in which every call succeeded could end with the badge and the button still up. The pass now records the ids it confirmed read and re-applies them to whatever array is current.
4. **Ids skipped as already-in-flight were dropped from both the marking and the failure count** — so "Read all" could leave a message unread and report a clean success, since a row's own mark-read swallows its errors by design. Now retried in a second bounded pass, with the banner's denominator counting what was actually attempted.
5. **The Read-all test tapped without guarding `exists`/`isHittable`.** The button can vanish between `waitForExistence` and `tap()` (the launch-time `onAppear` mark-read), and a failed `tap()` is an unrecoverable XCTest failure — so with `continueAfterFailure = false` the retry loop could never reach attempt two. It now guards, sleeps, and re-enters the loop.
6. **The iPad/Mac sidebar button ignored the Saved-section gate** the phone header's copy documents, rendering a control that fires a pass changing nothing on that screen. Same gate now.
7. **The search field rendered above the `.loading` spinner and the `.unavailable` error card** — a live filter over nothing. Shown only when Saved is up or the inbox reached `.loaded`.

Re-verified after the fixes: `swiftc -parse` clean on all three files (syntax only — CI is still the compiler), and the HerdrKit suite is untouched by an App-layer change (474 tests, 15 skipped, 0 failures locally in `swift:6.1`).

Issue #240, opened 2026-09-10
Terminal composer: paste and attach files for agents


jerryfane:
## Goal

Let the terminal reply box accept clipboard text, clipboard images, files, photos, and drag-and-drop on iPhone, iPad, and Mac, then give the active agent a real host-local file path it can read.

Authorization: workflow note `131092`.

## Product behavior

- Normal short text stays in the reply field.
- A user-mediated paste larger than about 4 KB becomes a disk-backed `.txt` attachment chip. Show an immediate Undo action that restores the text inline.
- Clipboard images/screenshots become image attachment chips.
- Files can come from Files/Finder, Photos, the system paste control, and drag-and-drop.
- Never read `UIPasteboard`/`NSPasteboard` to inspect clipboard contents. The user must invoke the system paste UI or paste into the text field. This preserves the existing App Review/CI restriction.
- Reuse Gram's existing disk-backed staging, 100 MB per-file limit, 10-file batch limit, streaming upload, progress, retry, and cleanup behavior.
- Send each file as a direct owner-to-agent Gram attachment. After `gram.post` finalizes it, append the returned host-local path to the normal `agent.prompt` text and submit through the existing terminal input router.
- Multiple files produce one direct Gram message per file and one terminal prompt containing every finalized path.
- If upload/post succeeds but prompt delivery fails, retain the finalized references and retry only the prompt. Do not upload duplicate bytes.
- Capture the target pane/agent before the first await. A target change must never redirect a secret-bearing file.

## Ordering invariant

The final file path contains the server-created Gram message ID, so the app cannot know the path before `gram.post`. The daemon must return the finalized path in the post result; the app then builds and sends the terminal prompt. Do not use a guessed ID, placeholder text, or a second path-only Gram message.

## Federation invariant

The path must belong to the machine that owns the target pane. Uploading to the home daemon and prompting a federated agent with the home machine's path is incorrect. Federation support must route upload/finalization to the pane's owning daemon or refuse before transferring bytes.

## Acceptance criteria

- iPhone, iPad, and Mac can attach short/long text, images, and arbitrary files from the requested sources.
- Large user-mediated pastes convert at the 4 KB boundary and Undo restores the exact text.
- No new programmatic pasteboard reads.
- Attachment chips show name, size, remove, and upload progress; staged bytes are removed when a chip is removed or a send completes.
- The agent receives the original typed caption plus exact finalized file paths through the normal prompt path.
[...]

PR #244, opened 2026-09-10
Attach clipboard and files in terminal composer

### LINKED ISSUE #241: Extract reusable disk-backed attachment staging
## Parent

Terminal composer clipboard and file attachments (parent link added after creation).

Authorization: workflow note `131092`.

## Change

Extract the proven disk-backed attachment staging used by `GramView` into a reusable app component, then keep Gram behavior unchanged.

- Model staged name, MIME type, URL, directory, size, identity, and cleanup ownership.
- Reuse the 100 MB per-file and 10-file limits.
- Support Files/Finder URLs, Photos results, clipboard Transferables, large pasted text, and drag-and-drop without loading whole large files into memory.
- Centralize remove/success cleanup and skip/error reporting.
- Never read `UIPasteboard` or `NSPasteboard`; accept only data delivered by user-mediated system paste/drop/import APIs.

## Acceptance

Gram's current attachment flow still supports its existing sources, progress, retry, and cleanup. The terminal composer can use the same staging type without copying the logic or retaining large attachments in memory.

### LINKED ISSUE #242: Add attachment chips and two-phase send to terminal composer
## Parent

Terminal composer clipboard and file attachments (parent link added after creation).

Authorization: workflow note `131092`.

## Change

Add attachment chips and two-phase send to `TerminalPaneContent`.

- Add a system-mediated Paste/Attach entry point plus Files/Finder, Photos, and drag-and-drop.
- Keep pasted text at or below about 4 KB inline. Convert a larger paste into a UTF-8 `.txt` chip and offer Undo that restores the exact text.
- Render disk-backed chips with name, size, remove, progress, and retry state.
- On Send, snapshot pane/agent and draft; upload/post each attachment through Gram; collect daemon-returned finalized paths; then call the existing input router/`agent.prompt` path once with caption plus paths.
- If attachment finalization succeeds and prompt delivery fails, retry only prompt delivery.
- Preserve text-only reply behavior, Ctrl interception, dictation, saved prompts, prefill safety, keyboard focus, and send-button behavior.

## Acceptance

The real terminal surface on iPhone, iPad, and Mac can send each requested source. The active agent receives readable paths and the original caption. Secret-bearing files cannot be silently redirected after Send, and failed sends do not duplicate uploads or Gram messages.

jerryfane:
## Summary

- add user-mediated paste, Photos, Files/Finder, and drag/drop attachment entry points to the terminal reply bar
- keep short text inline and turn pastes over 4 KB into disk-backed text chips with Undo
- upload through Gram, collect daemon-finalized paths, then send one normal agent prompt with those paths
- retain upload/message/path state so retries do not duplicate successful work
- route attachment uploads by pane for federated agents and recheck the target before prompting

No programmatic pasteboard reads were added. Authorization: workflow note `131092`.

## Verification

- HerdrKit staging and wire tests added for exact bytes, bounds, finalized-path decoding, and pane-target encoding
- iOS/macOS app build and real-surface verification are expected from PR CI/buildbox because this Linux seat has no Swift/Xcode toolchain

Closes #241
Closes #242
Part of #240

### DIFF
--- a/App/GramView.swift
+++ b/App/GramView.swift
@@ -1947,7 +1947,7 @@ private struct GramRow: View {
 
 /// Picks an SF Symbol for a file from its MIME type. Shared by the composer's
 /// staged-attachment chip and a received message's file chip.
-private enum FileGlyph {
+enum FileGlyph {
     static func name(for mime: String, fileName: String) -> String {
         let mime = mime.lowercased()
         if mime.hasPrefix("image/") { return "photo" }
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -5012,6 +5012,7 @@ struct TerminalPaneContent: View {
                 Menu {
                     ForEach(savedPrompts.prompts) { p in
                         Button(role: .destructive) { savedPrompts.delete(p.id) } label: { Text(p.label) }
+                    }
                 } label: { Label("Delete a prompt", systemImage: "trash") }
             }
         } label: {
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -4937,6 +4937,10 @@ struct TerminalPaneContent: View {
                     else { throw TerminalAttachmentSendError.agentChanged }
 
                     if let path = terminalAttachments[currentIndex].localPath {
+                        guard terminalAttachments[currentIndex].postedPaneID == targetPane,
+                              terminalAttachments[currentIndex].postedMachineID == targetMachineID,
+                              terminalAttachments[currentIndex].postedAgentName == targetName
+                        else { throw TerminalAttachmentSendError.agentChanged }
                         paths.append(path)
                         continue
                     }
@@ -4970,6 +4974,9 @@ struct TerminalPaneContent: View {
                         throw TerminalAttachmentSendError.daemonUpgradeRequired
                     }
                     terminalAttachments[currentIndex].localPath = path
+                    terminalAttachments[currentIndex].postedPaneID = targetPane
+                    terminalAttachments[currentIndex].postedMachineID = targetMachineID
+                    terminalAttachments[currentIndex].postedAgentName = targetName
                     TerminalAttachmentStaging.remove(original)
                     paths.append(path)
                 }
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -4759,6 +4759,7 @@ struct TerminalPaneContent: View {
                     .foregroundStyle(Palette.textFaint)
             }
             .disabled(sending)
+            .accessibilityLabel("Remove \(attachment.name)")
         }
         .padding(.horizontal, 9)
         .padding(.vertical, 6)
@@ -4825,7 +4826,10 @@ struct TerminalPaneContent: View {
               let attachment = TerminalAttachmentStaging.stageData(
                 Data(text.utf8), named: "pasted-text.txt", mime: "text/plain")
         else {
-            actionNote = "The pasted text is too large or could not be attached."
+            suppressLargePasteConversion = true
+            reply = restoredDraft
+            DispatchQueue.main.async { suppressLargePasteConversion = false }
+            actionNote = "The pasted text could not be attached, so it was kept in the reply."
             return
         }
         terminalAttachments.append(attachment)
@@ -4881,16 +4885,17 @@ struct TerminalPaneContent: View {
         targetPane: String,
         attachment: HerdrClient.GramFileAttachment
     ) async throws -> GramPostReceipt {
-        for attempt in 0..<3 {
+        var attempt = 0
+        while true {
             do {
                 return try await client.gramPostReceipt(
                     text: "", targetPane: targetPane, attachment: attachment)
             } catch let error as APIError where error.code == "upload_in_progress" {
-                if attempt == 2 { throw error }
+                attempt += 1
+                if attempt == 3 { throw error }
                 try? await Task.sleep(nanoseconds: 1_000_000_000)
             }
         }
-        throw TerminalAttachmentSendError.daemonUpgradeRequired
     }
 
     private static func promptText(caption: String, paths: [String]) -> String {
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -1,5 +1,6 @@
 import SwiftUI
 import Foundation
+import PhotosUI
 import Security
 import UIKit    // UIPasteboard (Copy diagnostics)
 import Darwin   // inet_pton/inet_ntop for IPv6 canonicalization
@@ -236,6 +237,7 @@ struct RootView: View {
         // killed session — up to ten 100 MB attachments — would otherwise survive every
         // launch in which the user never opens the Gram tab.
         .task { await GramView.Staging.sweepAbandonedOffMainActor() }
+        .task { await TerminalAttachmentStaging.sweepAbandonedOffMainActor() }
     }
 
     @ViewBuilder
@@ -3769,6 +3771,15 @@ struct TerminalPaneContent: View {
     @State private var showSavePrompt = false
     @State private var sending = false
     @State private var actionNote: String?
+    @State private var terminalAttachments: [TerminalAttachment] = []
+    @State private var terminalUploadBytes: (sent: Int, total: Int)?
+    @State private var terminalSendProgress: (sent: Int, total: Int)?
+    @State private var terminalPhotoItems: [PhotosPickerItem] = []
+    @State private var showTerminalPhotos = false
+    @State private var showTerminalFileImporter = false
+    @State private var loadingTerminalAttachment = false
+    @State private var suppressLargePasteConversion = false
+    @State private var pastedTextUndo: (attachmentID: UUID, restoredDraft: String)?
     /// In-flight guard for the [Switch] banner action, so repeated taps don't queue multiple
     /// `/tui default` prompts (the reply box's send is gated by `sending`; this is its analogue).
     @State private var switchingTui = false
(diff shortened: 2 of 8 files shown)


jerryfane:
> Runtime: `claude`
> Job: `local-review-herdr-ios-review-opus-18d4095f1d0bc99e-1`

**Summary:** Reviewed PR #244 at exact head 4e99fcbfc37aa69d125770861630e288ac75e32c (base 93c65786...) via full diff read against the GitHub API (no working Swift toolchain in this seat, so static/diff review only). The two-phase Gram-then-prompt flow, per-file upload/post/localPath retry-idempotency state machine, target re-snapshot before the final prompt, federated target_pane_id threading through upload/post, 4KB inline-vs-file threshold, 100MB/10-file caps, and per-item disk cleanup (plus session-sweep-on-launch) are all implemented correctly and consistently, backed by new HerdrKit tests for exact-byte staging, bounds, local_file_path decoding, and target_pane_id encoding on both upload_chunk and gram.post. No programmatic pasteboard reads were added (uses SwiftUI's system PasteButton + Transferable), matching the stated privacy goal. One concrete, reproducible data-loss bug was found in the large-paste-to-attachment conversion path when the 10-attachment cap is already reached, which should be fixed before merge.
**Severity:** `P2`

**Findings**
- **P2**
  - description: When a >4KB paste (via the system PasteButton) or a large contiguous in-field insertion needs conversion to a file attachment but TerminalAttachmentStaging.maxAttachments (10) is already reached — or GramStaging.stageData otherwise fails — the pasted/typed text is silently discarded. handleReplyChange already sets `reply = old` (stripping the insertion from the field) before calling stageLargePastedText, and handlePastedText never falls back to inlining the text either. The user only sees a transient actionNote ("too large or could not be attached") and permanently loses the clipboard content instead of it being kept in the reply field or otherwise recoverable. Repro: stage 10 attachments, then paste or type >4KB of text into the reply field — that text vanishes.
  - file: App/HerdrApp.swift
  - location: handlePastedText/stageLargePastedText (~4814-4834) and handleReplyChange (~5086-5100)
- **P3**
  - description: The trailing `throw TerminalAttachmentSendError.daemonUpgradeRequired` after the `for attempt in 0..<3` retry loop is unreachable dead code — every loop iteration either returns successfully, rethrows the APIError on attempt==2, or lets a non-upload_in_progress error propagate uncaught. Harmless (satisfies the compiler's exhaustiveness requirement) but could mislead a future reader into thinking there's a distinct "gave up after 3 tries" fallback path.
  - file: App/HerdrApp.swift
  - location: postTerminalAttachment (~4879-4892)
- **P3**
  - description: The per-chip remove (xmark.circle.fill) button has no explicit .accessibilityLabel, so VoiceOver gets no clear "Remove <filename>" announcement — unlike the adjacent new paperclip menu button, which does have .accessibilityLabel("Attach or paste"). This mirrors a pre-existing gap in the sibling Gram-compose chip (App/GramView.swift:987-989), so it's not a regression unique to this PR, but worth closing now that a labeled precedent was already added alongside it.
  - file: App/HerdrApp.swift
  - location: terminalAttachmentChip (~4756-4761)
- **P3**
  - description: The doc comment states `targetPane` is "mutually exclusive with `to`" but this isn't enforced (no precondition/assert). All current call sites respect it correctly (GramView.swift uses `to` only, the new terminal-attachment path uses `targetPane` only), so this is a defensive-programming suggestion, not an active bug.
  - file: Sources/HerdrKit/HerdrClient.swift
  - location: gramPost/gramPostReceipt (~505-528)

**Tests Run**
- Attempted to locate/run a Swift toolchain (`which swift`) — none available in this seat; `swift test` / `swift build` could not be executed
- Read full diff base 93c6578666e656c3206661389e81853bcc0b88da..4e99fcbfc37aa69d125770861630e288ac75e32c across all 8 changed files (App/HerdrApp.swift, App/TerminalAttachment.swift, App/GramView.swift, Sources/HerdrKit/Gram.swift, Sources/HerdrKit/GramStaging.swift, Sources/HerdrKit/HerdrClient.swift, Tests/HerdrKitTests/GramStagingTests.swift, Tests/HerdrKitTests/MockWireFixtureTests.swift)
- Traced the upload→post→prompt state machine (uploadID/gramMessageID/localPath/postedPaneID/postedMachineID/postedAgentName) by hand against every throw/retry path for idempotency and target-snapshot safety
- Cross-checked the new `upload_in_progress` retry logic against the pre-existing documented daemon-claim-release race in GramUploadChannel.swift:closeAndWait()
- Compared the new terminal-attachment chip against the existing sibling Gram-compose chip (App/GramView.swift) for UI/accessibility/cleanup convention consistency
- Verified via `git log`/GitHub API that the reviewed head (4e99fcbfc37aa69d125770861630e288ac75e32c) matches the PR's actual head and already includes the 'fix: bind attachment retries to original agent' commit

jerryfane:
> Runtime: `claude`
> Job: `local-review-herdr-ios-review-opus-18d40a62cd2ff71e-1`

**Decision:** `approved`
**Summary:** Reviewed PR #244 at exact head bc5f876. All three follow-up fixes are correct and behavior-preserving: (1) oversized-paste text is now restored into the reply field (with a suppression flag to avoid re-triggering the same staging attempt) whenever staging fails for any reason, including the 10-file cap; (2) the upload_in_progress retry loop was converted from a for-loop with unreachable trailing throw to a while-loop, preserving the exact same attempt/sleep semantics (3 attempts, 2 sleeps, throw on the 3rd failure); (3) the remove (xmark.circle.fill) button on the terminal-attachment chip now carries .accessibilityLabel("Remove <filename>"), closing prior finding f1. Rechecked the full diff (HerdrClient.swift, Gram.swift, GramStaging.swift, GramView.swift, TerminalAttachment.swift, tests) for regressions: wire params/tests for target_pane_id are consistent, the agent-binding guard added for attachment retries correctly prevents cross-agent path reuse without breaking same-agent retries, and staged attachments/local paths are left intact on any send-path failure (not just the paste-staging failure), so nothing new is lost on error. No Swift/Xcode toolchain is available in this Linux sandbox, so this is a static-only review; I did run a structural brace-balance sanity pass over every changed Swift file, which found no imbalance.

**Findings**
- **Remove-chip accessibility label** (P3)
  - continues_uid: jerryfane/herdrup#244-f1
  - detail: Verified fixed at bc5f876: the terminal-attachment chip's remove button now has .accessibilityLabel("Remove \(attachment.name)") applied directly on the Button, mirroring the paperclip menu's .accessibilityLabel("Attach or paste"). The originally-noted sibling gap in the Gram-compose chip (App/GramView.swift:987-989) was out of scope for this PR and remains unlabeled — pre-existing, non-blocking, still worth a follow-up but not part of this change.
  - file: App/HerdrApp.swift
  - state: answered
- **targetPane/to mutual exclusivity still unenforced** (P3)
  - continues_uid: jerryfane/herdrup#244-f2
  - detail: Sources/HerdrKit/HerdrClient.swift was not touched by any of the three follow-up commits (10c16ce, 4e99fcb, bc5f876), so gramPostReceipt's doc comment claiming targetPane is "mutually exclusive with to" is still unenforced by precondition/assert. All call sites (GramView.swift uses to only; the terminal-attachment path uses targetPane only) still respect it correctly, so this remains a defensive-programming suggestion rather than an active bug.
  - file: Sources/HerdrKit/HerdrClient.swift
  - state: answered

**Tests Run**
- Traced stageLargePastedText/handleReplyChange/handlePastedText call graph and the suppressLargePasteConversion flag to confirm oversized-paste text is preserved on staging failure (cap reached or stageData returning nil) without re-triggering an infinite re-staging loop
- Traced postTerminalAttachment's old for-loop vs new while-loop retry logic attempt-by-attempt to confirm identical attempt count (3) and sleep placement
- Traced sendTerminalAttachments' new postedPaneID/postedMachineID/postedAgentName guard to confirm it blocks stale cross-agent localPath reuse on retry while still allowing same-agent retries to skip re-upload
- Confirmed staged attachments/files are left untouched (not cleared) on every throw path in sendTerminalAttachments, so failures don't silently drop user data
- python3 brace-balance sanity check across all changed Swift files (App/HerdrApp.swift, App/TerminalAttachment.swift, App/GramView.swift, Sources/HerdrKit/Gram.swift, Sources/HerdrKit/GramStaging.swift, Sources/HerdrKit/HerdrClient.swift, and the two changed test files) — no Swift/Xcode toolchain available in this Linux sandbox, so xcodebuild/swift test could not be executed

jerryfane:
Obsolete — the feature is on `main`, reached by a different route. Authorized by the owner in session; record: workflow note 168422.

The terminal reply composer today: paste of files and images (`PastedFile` decides file-vs-text from pasteboard **type metadata only**, and a file URL beats text so Command-V stopped pasting `/Users/…/main.pdf`), drag-and-drop over the whole bar (`onDrop(of: [.item])` → `acceptDroppedFiles`, same staging path as paste so the cap and type rule are shared), several attachments as chips (`replyAttachments`, capped at `GramView.Staging.maxAttachments`), and per-upload progress. Covered by `TerminalControlTests`.

That arrived through the Gram staging work — #226, #251, #253 — which reused one path for both composers. This branch re-implements it as a separate `TerminalAttachment` type, which is also why it has been conflicted since 10 Sep.

**Two things here did NOT land, recorded so they are not lost with the branch:**

1. **Large text pastes stay inline.** This branch had `inlineTextBytes = 4 * 1024`, turning an oversized text paste into a disk-backed chunk. On `main`, `paste(_:)` only diverts *files*; a huge text paste still goes into the field as text.
2. **No Photos entry point in the terminal composer.** `PhotosPicker` exists only in `GramView`. The terminal bar takes pasted and dropped images, not a picker.

#240, #241 and #242 stay open for those. Worth deciding whether they are wanted at all before reviving code for them: a paste large enough to need chunking may be better refused with a note than silently turned into a file.

PR #245, opened 2026-09-11
Redesign Dynamic Island and Lock Screen Live Activity

jerryfane:
## What changed
- makes the needs-you count the hero across compact, expanded, and Lock Screen presentations
- replaces colour-only dots with distinct filled, ring, small-dot, barred, and stale-hollow status marks
- carries blocked age and the exact pane target into the ActivityKit state
- marks content stale after 90 seconds and dims secondary Lock Screen detail on Always-On
- adds one explicit Open action plus whole-surface deep links to the named agent
- bundles Geist, IBM Plex Mono, and the app logo in the widget extension

The current daemon does not expose a prompt string or safe default answer. The design's explicit no-default fallback is therefore used: Open the named agent rather than guessing an approval response.

Authorized by owner directive and design attachment `gram-1a08f9cb015-1394e1-5e`.

## Verification
- Linux cannot run Swift/Xcode in this workspace (`swift: command not found`)
- CI must compile the app/widget and run the Swift package tests
- manual device rendering remains required because XCUITest cannot inspect a Live Activity

### DIFF
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -25,7 +25,10 @@ struct HerdrApp: App {
     }
 
     var body: some Scene {
-        WindowGroup { RootView() }
+        WindowGroup {
+            RootView()
+                .onOpenURL { PushCenter.shared.openLiveActivityURL($0) }
+        }
     }
 }
 
--- a/App/LiveActivityController.swift
+++ b/App/LiveActivityController.swift
@@ -3,10 +3,9 @@ import Combine
 import ActivityKit
 import HerdrKit
 
-/// Owns the single agent-session Live Activity: starts it when a session connects,
-/// updates it as the agent list changes, ends it on disconnect. Every entry point is
-/// a no-op when the user has Live Activities turned off (Settings) or the platform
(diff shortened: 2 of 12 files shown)


jerryfane:
Superseded — everything this PR proposed is on `main`, shipped in TestFlight build 149. Authorized by the owner in session; record: workflow note 168422.

| this PR proposed | where it lives now |
|---|---|
| needs-you count as the hero | `LockScreenView.lockHero`, suppressed at a count of one where the design makes the agent's name the hero |
| distinct shapes instead of colour-only dots | `StatusMark`: filled (needsYou), ring (working), small dot (idle), barred capsule (stopped), hollow ring when stale or unconfirmed |

Landed through #256 → #261 instead, which also took it to the device: an opaque vertical gradient after the glass route failed on hardware (#257, #259), the original design's artwork and state matrix (#260), and the 26pt logo raster that #260 got wrong and #261 fixed (#261).

Closing rather than rebasing: the branch has been conflicted since 11 Sep and the file it touches has been rewritten several times since. Reopen if something here was missed — the #260/#261 review leftovers are tracked in #262.

PR #247, opened 2026-09-11
fix(composer): cap multiline inputs above send controls

jerryfane:
## What
- let both Gram and terminal reply fields insert newlines
- grow each field upward from one to three visible lines, then scroll internally
- bottom-align and fix-size action controls so send stays visible and stationary
- retain Command+Return as the hardware-keyboard send shortcut
- treat terminal replies containing only spaces or newlines as empty and unsendable
- keep Gram search controls above message content so the icon stays visible and clickable
- bind Command+F to open or refocus Gram search
- add UI regressions for composer layout, overflow editing, newline-only eligibility, and search hit testing

## Verification
- `git diff --check`
- focused Gram and terminal UI regressions
- exact-head simulator CI

Addresses the Gram/macOS composer and search reports.

PR #248, opened 2026-09-12
Redesign Dynamic Island and Lock Screen activity

jerryfane:
Implements the saved Herdrup Island design package:

- exact status-mark vocabulary and compact/minimal count rules
- redesigned expanded Dynamic Island hierarchy and safe Open action
- redesigned Lock Screen hierarchy, stale treatment, and Always-On reduction
- Geist / IBM Plex Mono resources in the widget extension
- activity deep links to the target pane
- additive prompt, timing, identity, and freshness state with older-payload decoding

Safety: the current daemon does not provide a stable default answer or an atomic approval endpoint, so the activity uses the design's Open fallback rather than pretending an Approve action succeeded.

## Terminal composer scope (beyond the title)

This branch also changes the terminal reply composer. Listing it explicitly because a review found it undisclosed:

- **UIKit composer.** The reply field is now a `UITextView` (`TerminalReplyField`) so it can grow to three lines and then scroll internally, with the Send button and the field's bottom edge staying put. Scroll mode is toggled from text/layout updates, never from SwiftUI's sizing pass.
- **Photo attachments.** A copied photo can be pasted into the composer. The image is staged on disk, uploaded through the Gram upload channel, posted as a file attachment, and the prompt sent to the agent references that file. No OCR, no terminal escape-sequence blob. A reply with only a photo and no caption is sendable.
- **Return now sends; Command+Return no longer does.** This REVERSES the semantics approved in #247 (Return inserted a newline, Command+Return sent). Typed newlines are therefore no longer possible; a newline can only arrive by paste. Flagging it for an explicit product decision instead of landing it silently.

## Review round 2 (head 4dce5bd)

Fixes for the blocking findings of the independent review at head `0c61c63`:

- **P1 false connectivity claim.** `staleDate` was `Date()+90s` at both ActivityKit sites while the only `update()` caller is change-driven and no heartbeat exists, so a healthy, stable roster rendered "Connection stale" and "<host> unreachable" 90 seconds later. `staleDate` is now `nil` (claims nothing), and the stale copy that remains says only "Not updated recently" / "No update from <host>" rather than asserting unreachability.
- **P1 headline override.** A widget-private `displayHeadline` replaced the headline with "Nothing needs you" whenever `needsYouCount == 0`, so a died agent rendered that all-clear above a red stopped mark, a working agent lost its name, and the connect handshake asserted all-clear with nothing known. The override is gone; the headline is the lead agent's name, decided in `AgentList.activityContent`. `markIsUnconfirmed` moved from the untestable widget target into `Shared/` and is now pinned by a test.
- **P2 pasteboard privacy.** `canPerformAction` read `UIPasteboard.general.itemProviders` while UIKit built the edit menu — a content read that raises "Allow Paste?" in production. It now uses the non-alerting `hasImages`; the content read stays inside the user-initiated `paste(_:)`.
- **P2 paste fall-through.** `paste(_:)` now defers to UIKit whenever text is present or the composer declines the photo, so mixed image+text pastes keep their text and plain-text paste still works on a raw-keys pane.
- **P2 state mutation during update.** The end-editing callback no longer writes the focus binding from inside `updateUIView`; it hops off the update pass and skips equal writes.
- **P2 keyboard drop on send.** The field is disabled only while dictating. Gating it on `sending` ended editing on every reply and never re-acquired focus.
- **P2 neutered test.** `testReplyContainingOnlyNewlinesIsNotSendable` now pastes `"\n\n"` through the harness pasteboard and asserts the newlines reached the composer, so it exercises the trimming rule again instead of passing because newlines are uninsertable.

Open P3s from that review, not addressed here: `question`/`defaultAnswer` are rendered/declared but never produced by the app, `postReplyPhoto` duplicates Gram's post retry, `updatedAt` mixes device and daemon clocks, and the widget target copies the whole font directory while registering five files.

Merge and TestFlight release authorized by workflow note 147573.

### DIFF
--- a/App/HerdrApp.swift
+++ b/App/HerdrApp.swift
@@ -25,7 +25,13 @@ struct HerdrApp: App {
     }
 
     var body: some Scene {
-        WindowGroup { RootView() }
+        WindowGroup {
+            RootView()
+                .onOpenURL { url in
+                    guard let paneID = AgentActivityDeepLink.agentID(from: url) else { return }
+                    PushCenter.shared.tapped(paneID: paneID)
+                }
+        }
     }
 }
 
--- a/App/LiveActivityController.swift
+++ b/App/LiveActivityController.swift
@@ -60,7 +60,7 @@ final class LiveActivityController: ObservableObject {
             // still go through update() directly.
             activity = try Activity.request(
                 attributes: attributes,
-                content: ActivityContent(state: state, staleDate: nil),
+                content: ActivityContent(state: state, staleDate: Date().addingTimeInterval(90)),
                 pushType: .token
             )
             observePushToken()
@@ -87,7 +87,7 @@ final class LiveActivityController: ObservableObject {
     /// Push a new state to the live activity, if there is one.
     func update(_ state: AgentActivityAttributes.ContentState) {
         guard let activity else { return }
-        Task { await activity.update(ActivityContent(state: state, staleDate: nil)) }
+        Task { await activity.update(ActivityContent(state: state, staleDate: Date().addingTimeInterval(90))) }
     }
 
     /// End and clear the activity immediately (on disconnect / sign-out). Ends EVERY
@@ -132,7 +132,10 @@ final class LiveActivityController: ObservableObject {
             unconfirmedCount: c.unconfirmedCount,
             workingCount: c.workingCount,
             totalCount: c.totalCount,
-            workingSince: c.workingSinceUnixSeconds
+            workingSince: c.workingSinceUnixSeconds,
+            blockedSince: c.blockedSinceUnixSeconds,
+            agentID: c.agentID,
+            updatedAt: Date().timeIntervalSince1970
         )
     }
 
--- a/HerdrWidgets/AgentLiveActivity.swift
+++ b/HerdrWidgets/AgentLiveActivity.swift
@@ -2,196 +2,393 @@ import SwiftUI
 import WidgetKit
 import ActivityKit
 
-/// The agent-session Live Activity: a lock-screen banner and the Dynamic Island
-/// presentations. It renders `AgentActivityAttributes` — a status dot coloured by
-/// the session's highest-priority agent, that agent's name, and a one-line summary
-/// ("2 need you" / "Working" / "Idle"). Colours mirror the app's status palette
-/// (amber = needs you, blue = working, red = stopped, faint = idle); the widget
-/// keeps its own copy because the app's DesignSystem is app-target only.
 struct AgentLiveActivity: Widget {
     var body: some WidgetConfiguration {
         ActivityConfiguration(for: AgentActivityAttributes.self) { context in
-            // Lock screen / notification-banner presentation.
-            LockScreenView(hostLabel: context.attributes.hostLabel, state: context.state)
-                .activityBackgroundTint(WidgetPalette.ground)
-                .activitySystemActionForegroundColor(WidgetPalette.text)
+            LockScreenView(
+                hostLabel: context.attributes.hostLabel,
+                state: context.state,
+                isStale: context.isStale
+            )
+            .widgetURL(context.state.deepLinkURL)
+            .activityBackgroundTint(WidgetPalette.ground)
+            .activitySystemActionForegroundColor(WidgetPalette.text)
         } dynamicIsland: { context in
             DynamicIsland {
                 DynamicIslandExpandedRegion(.leading) {
-                    StatusDot(status: context.state.status)
+                    ExpandedHero(state: context.state, isStale: context.isStale)
                         .padding(.leading, 4)
                 }
+                DynamicIslandExpandedRegion(.center) {
+                    ExpandedHeadline(state: context.state, isStale: context.isStale)
+                }
                 DynamicIslandExpandedRegion(.trailing) {
-                    Text(context.attributes.hostLabel)
-                        .font(.caption2)
-                        .foregroundStyle(WidgetPalette.textFaint)
-                        .lineLimit(1)
+                    FleetTotals(hostLabel: context.attributes.hostLabel, state: context.state)
                         .padding(.trailing, 4)
                 }
-                DynamicIslandExpandedRegion(.center) {
-                    VStack(spacing: 2) {
-                        Text(context.state.headline)
-                            .font(.headline)
-                            .foregroundStyle(WidgetPalette.text)
-                            .lineLimit(1)
-                        // The ticking timer shows ONLY when exactly one agent is working, so it
-                        // unambiguously tracks THAT agent and stops the moment it finishes. The
-                        // Live Activity is one aggregate per machine, so with several agents working
-                        // a single "since" is meaningless (and the timer would appear to run on as
-                        // the busiest agent changes) — show the "N working" count instead.
-                        if context.state.status == .working, context.state.workingCount == 1,
-                           let since = context.state.workingSince {
-                            HStack(spacing: 5) {
-                                Text("Working").font(.caption).foregroundStyle(WidgetPalette.color(.working))
-                                WorkingTimer(since: since, font: .caption)
-                            }
-                            .lineLimit(1)
-                        } else {
-                            Text(summary(context.state))
-                                .font(.caption)
-                                .foregroundStyle(WidgetPalette.color(context.state.status))
-                                .lineLimit(1)
-                        }
+                DynamicIslandExpandedRegion(.bottom) {
+                    if context.state.needsYouCount > 0 {
+                        ActivityAction(
+                            title: "Open",
+                            destination: context.state.deepLinkURL,
+                            outlined: context.isStale
+                        )
+                        .padding(.horizontal, 8)
+                        .padding(.top, 6)
                     }
                 }
             } compactLeading: {
-                StatusDot(status: context.state.status)
+                StatusMark(
+                    status: context.state.status,
+                    diameter: 10,
+                    isUnconfirmed: context.state.markIsUnconfirmed,
+                    isStale: context.isStale
+                )
             } compactTrailing: {
-                // ONLY the count that demands action gets the scarce compact slot. A ticking
-                // timer here widened the notch pill (and kept growing as it counted up), so the
-                // working state stays a bare dot in the compact view; its live elapsed timer
-                // lives in the EXPANDED view and the lock screen, where width isn't constrained.
-                if context.state.needsYouCount > 0 {
-                    Text("\(context.state.needsYouCount)")
-                        .font(.caption2).bold()
-                        .foregroundStyle(WidgetPalette.color(.needsYou))
-                }
+                CompactCount(state: context.state)
             } minimal: {
-                StatusDot(status: context.state.status)
+                StatusMark(
+                    status: context.state.status,
+                    diameter: 14,
+                    isUnconfirmed: context.state.markIsUnconfirmed,
+                    isStale: context.isStale
+                )
+                .frame(width: 36, height: 36)
             }
-            .keylineTint(WidgetPalette.color(context.state.status))
+            .widgetURL(context.state.deepLinkURL)
+            .keylineTint(context.state.markColor)
         }
-    }
-
-    /// One-line summary line: prefer the actionable count, else the status word.
-    private func summary(_ s: AgentActivityAttributes.ContentState) -> String {
-        AgentActivitySummary.line(s)
     }
 }
 
-/// The lock-screen / banner layout: status dot, agent name + summary, host on the
-/// right. Kept compact and legible on the dark banner tint.
-private struct LockScreenView: View {
+private struct CompactCount: View {
+    let state: AgentActivityAttributes.ContentState
+
+    @ViewBuilder
+    var body: some View {
+        if state.needsYouCount > 0 {
+            Text(state.needsYouCount.compactCount)
+                .font(WidgetFont.plexSemiBold(13))
+                .monospacedDigit()
+                .foregroundStyle(WidgetPalette.waiting)
+        } else if state.workingCount > 0 {
+            Text(state.workingCount.compactCount)
+                .font(WidgetFont.plexMedium(13))
+                .monospacedDigit()
+                .foregroundStyle(WidgetPalette.working)
+        }
+    }
+}
+
+private struct ExpandedHero: View {
+    let state: AgentActivityAttributes.ContentState
+    let isStale: Bool
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 1) {
+            if state.needsYouCount > 0 {
+                HStack(spacing: 7) {
+                    StatusMark(
+                        status: state.status,
+                        diameter: 10,
+                        isUnconfirmed: state.markIsUnconfirmed,
+                        isStale: isStale
+                    )
+                    if state.needsYouCount > 1 {
+                        Text(verbatim: "\(state.needsYouCount)")
+                            .font(WidgetFont.plexSemiBold(26))
+                            .monospacedDigit()
+                            .foregroundStyle(WidgetPalette.waiting)
+                    }
+                }
+                Text("need you")
+                    .font(WidgetFont.geist(12))
+                    .foregroundStyle(WidgetPalette.textDim)
+            } else {
+                StatusMark(status: state.status, diameter: 10, isStale: isStale)
+            }
+        }
+    }
+}
+
+private struct ExpandedHeadline: View {
+    let state: AgentActivityAttributes.ContentState
+    let isStale: Bool
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 2) {
+            Text(state.activityHeadline(isStale: isStale))
+                .font(WidgetFont.geistSemiBold(16))
+                .foregroundStyle(WidgetPalette.text)
+                .lineLimit(1)
+            if let question = state.question, !question.isEmpty, !isStale {
+                Text(question)
+                    .font(WidgetFont.plex(12))
+                    .foregroundStyle(WidgetPalette.textDim)
+                    .lineLimit(1)
+            }
+            if isStale {
+                Text("Connection stale")
+                    .font(WidgetFont.plex(11))
+                    .foregroundStyle(WidgetPalette.textFaint)
+                    .lineLimit(1)
+            } else if state.needsYouCount > 0, let since = state.blockedSince {
+                HStack(spacing: 4) {
+                    Text("waiting")
+                    Text(Date(timeIntervalSince1970: since), style: .timer)
+                        .monospacedDigit()
+                }
+                .font(WidgetFont.plex(11))
+                .foregroundStyle(WidgetPalette.textFaint)
+                .lineLimit(1)
+            } else if state.status == .working, let since = state.workingSince {
+                Text(Date(timeIntervalSince1970: since), style: .timer)
+                    .font(WidgetFont.plex(11))
+                    .monospacedDigit()
+                    .foregroundStyle(WidgetPalette.textFaint)
+                    .lineLimit(1)
+            }
+        }
+    }
+}
+
+private struct FleetTotals: View {
     let hostLabel: String
     let state: AgentActivityAttributes.ContentState
 
     var body: some View {
-        HStack(spacing: 12) {
-            StatusDot(status: state.status, diameter: 12)
-            VStack(alignment: .leading, spacing: 2) {
-                Text(state.headline)
-                    .font(.headline)
-                    .foregroundStyle(WidgetPalette.text)
-                    .lineLimit(1)
-                // Timer ONLY for a single working agent (see the Dynamic Island note): it then
-                // tracks that agent and stops when it finishes. With several working, the "since"
-                // is ambiguous and looks like it never stops, so show the "N working" count. The
-                // live ticking time IS the motion here — the dot can't animate on a Live Activity.
-                if state.status == .working, state.workingCount == 1, let since = state.workingSince {
-                    HStack(spacing: 5) {
-                        Text("Working").font(.subheadline).foregroundStyle(WidgetPalette.color(.working))
-                        WorkingTimer(since: since)
+        VStack(alignment: .trailing, spacing: 2) {
+            Text(hostLabel)
+            Text("\(state.workingCount) working")
+            Text(state.totalCount == 1 ? "1 agent" : "\(state.totalCount) agents")
+        }
+        .font(WidgetFont.plex(11))
+        .foregroundStyle(WidgetPalette.textFaint)
+        .lineLimit(1)
+    }
+}
+
+private struct LockScreenView: View {
+    let hostLabel: String
+    let state: AgentActivityAttributes.ContentState
+    let isStale: Bool
+    @Environment(\.isLuminanceReduced) private var isLuminanceReduced
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 14) {
+            HStack(alignment: .top, spacing: 14) {
+                lockHero
+                VStack(alignment: .leading, spacing: 3) {
+                    Text(state.activityHeadline(isStale: isStale))
+                        .font(WidgetFont.geistSemiBold(18))
+                        .foregroundStyle(WidgetPalette.text)
+                        .lineLimit(1)
+                    if let question = state.question, !question.isEmpty, !isStale {
+                        Text(question)
+                            .font(WidgetFont.plex(15))
+                            .foregroundStyle(WidgetPalette.textDim)
+                            .lineLimit(1)
+                    } else if state.needsYouCount > 0, !isStale {
+                        Text(AgentActivitySummary.line(state))
+                            .font(WidgetFont.plex(13))
+                            .foregroundStyle(state.markColor)
+                            .lineLimit(1)
                     }
-                    .lineLimit(1)
-                } else {
-                    Text(summary)
-                        .font(.subheadline)
-                        .foregroundStyle(WidgetPalette.color(state.status))
-                        .lineLimit(1)
+                    if !isLuminanceReduced {
+                        lockDetail
+                    }
                 }
+                Spacer(minLength: 0)
             }
-            Spacer(minLength: 8)
-            VStack(alignment: .trailing, spacing: 2) {
-                Text(hostLabel)
-                    .font(.caption)
-                    .foregroundStyle(WidgetPalette.textFaint)
-                    .lineLimit(1)
-                if state.totalCount > 0 {
-                    Text(state.totalCount == 1 ? "1 agent" : "\(state.totalCount) agents")
-                        .font(.caption2)
-                        .foregroundStyle(WidgetPalette.textFaint)
-                        .lineLimit(1)
-                }
+
+            if state.needsYouCount > 0 {
+                ActivityAction(
+                    title: "Open Herdrup",
+                    destination: state.deepLinkURL,
+                    outlined: isStale || isLuminanceReduced
+                )
             }
         }
         .padding(.horizontal, 16)
-        .padding(.vertical, 12)
+        .padding(.vertical, 14)
     }
 
-    private var summary: String { AgentActivitySummary.line(state) }
+    @ViewBuilder
+    private var lockHero: some View {
+        if state.needsYouCount > 0 {
+            VStack(alignment: .leading, spacing: -1) {
+                HStack(alignment: .center, spacing: 7) {
+                    StatusMark(
+                        status: state.status,
+                        diameter: 12,
+                        isUnconfirmed: state.markIsUnconfirmed,
+                        isStale: isStale
+                    )
+                    Text(verbatim: "\(state.needsYouCount)")
+                        .font(WidgetFont.plexSemiBold(34))
+                        .monospacedDigit()
+                        .foregroundStyle(WidgetPalette.waiting)
+                }
+                Text("need you")
+                    .font(WidgetFont.geist(13))
+                    .foregroundStyle(WidgetPalette.textDim)
+            }
+            .fixedSize(horizontal: true, vertical: false)
+        } else {
+            StatusMark(status: state.status, diameter: 12, isStale: isStale)
+                .frame(width: 34, height: 34, alignment: .topLeading)
+        }
+    }
+
+    @ViewBuilder
+    private var lockDetail: some View {
+        if isStale {
+            if let updatedAt = state.updatedAt {
+                HStack(spacing: 3) {
+                    Text("last seen")
+                    Text(Date(timeIntervalSince1970: updatedAt), style: .relative)
+                    Text("· \(hostLabel) unreachable")
+                }
+                .font(WidgetFont.plex(11))
+                .foregroundStyle(WidgetPalette.textFaint)
+                .lineLimit(1)
+            } else {
+                Text("\(hostLabel) unreachable")
+                    .font(WidgetFont.plex(11))
+                    .foregroundStyle(WidgetPalette.textFaint)
+                    .lineLimit(1)
+            }
+        } else {
+            Text("\(hostLabel) · \(state.workingCount) working · \(state.totalCount) agents")
+                .font(WidgetFont.plex(13))
+                .foregroundStyle(WidgetPalette.textFaint)
+                .lineLimit(1)
+        }
+    }
 }
 
-/// A filled status circle; colour carries the meaning (amber = needs you, blue =
-/// working, red = stopped, faint = idle). STATIC on purpose: a Live Activity is
-/// rendered from snapshots and iOS frame-interpolates ONLY time-driven views
-/// (`Text(timerInterval:)` / `ProgressView(timerInterval:)`), so a pulsing or
-/// spinning dot genuinely cannot animate here. The working state's MOTION comes
-/// from the live `WorkingTimer` beside the dot, not from the dot itself.
-private struct StatusDot: View {
+private struct ActivityAction: View {
+    let title: String
+    let destination: URL
+    let outlined: Bool
+
+    var body: some View {
+        Link(destination: destination) {
+            Text(title)
+                .font(WidgetFont.geistSemiBold(15))
+                .frame(maxWidth: .infinity, minHeight: 44)
+                .foregroundStyle(outlined ? WidgetPalette.text : WidgetPalette.ground)
+                .background(outlined ? Color.clear : WidgetPalette.text)
+                .clipShape(RoundedRectangle(cornerRadius: 14, style: .continuous))
+                .overlay {
+                    if outlined {
+                        RoundedRectangle(cornerRadius: 14, style: .continuous)
+                            .strokeBorder(WidgetPalette.hairline, lineWidth: 1)
+                    }
+                }
+        }
+    }
+}
+
+private struct StatusMark: View {
     let status: AgentActivityAttributes.Status
     var diameter: CGFloat = 10
+    var isUnconfirmed = false
+    var isStale = false
 
+    @ViewBuilder
     var body: some View {
-        Circle()
-            .fill(WidgetPalette.color(status))
+        switch status {
+        case .needsYou:
+            if isStale || isUnconfirmed {
+                Circle()
+                    .strokeBorder(WidgetPalette.waiting, lineWidth: max(1.5, diameter * 0.16))
+                    .frame(width: diameter, height: diameter)
+            } else {
+                Circle()
+                    .fill(WidgetPalette.waiting)
+                    .frame(width: diameter, height: diameter)
+            }
+        case .working:
+            Circle()
+                .strokeBorder(WidgetPalette.working, lineWidth: max(1.5, diameter * 0.16))
+                .frame(width: diameter, height: diameter)
+        case .idle:
+            Circle()
+                .fill(WidgetPalette.textFaint)
+                .frame(width: max(4, diameter * 0.42), height: max(4, diameter * 0.42))
+                .frame(width: diameter, height: diameter)
+        case .stopped:
+            ZStack {
+                Circle()
+                    .strokeBorder(WidgetPalette.died, lineWidth: max(1.5, diameter * 0.15))
+                Capsule()
+                    .fill(WidgetPalette.died)
+                    .frame(width: diameter * 0.62, height: max(1.5, diameter * 0.14))
+            }
             .frame(width: diameter, height: diameter)
+        }
     }
 }
 
-/// A live, up-counting elapsed-time readout ("1:23") for the working state — the
-/// one thing iOS actually animates in a Live Activity (`Text(_, style: .timer)` is
-/// frame-interpolated, unlike `symbolEffect`/spinners). `since` is Unix SECONDS.
-private struct WorkingTimer: View {
-    let since: Double
-    var font: Font = .subheadline
-
-    var body: some View {
-        Text(Date(timeIntervalSince1970: since), style: .timer)
-            .font(font)
-            .monospacedDigit()
-            .foregroundStyle(WidgetPalette.color(.working))
-    }
+private enum WidgetFont {
+    static func geist(_ size: CGFloat) -> Font { .custom("Geist-Regular", fixedSize: size) }
+    static func geistSemiBold(_ size: CGFloat) -> Font { .custom("Geist-SemiBold", fixedSize: size) }
+    static func plex(_ size: CGFloat) -> Font { .custom("IBMPlexMono", fixedSize: size) }
+    static func plexMedium(_ size: CGFloat) -> Font { .custom("IBMPlexMono-Medm", fixedSize: size) }
+    static func plexSemiBold(_ size: CGFloat) -> Font { .custom("IBMPlexMono-SmBld", fixedSize: size) }
 }
 
-/// The widget's own copy of the app's status palette (DesignSystem is app-only).
-/// Hex values match `Palette` in App/DesignSystem.swift.
-enum WidgetPalette {
-    static let ground     = Color(hex6: 0x13162A)
-    static let text       = Color(hex6: 0xEEF0F7)
-    static let textFaint  = Color(hex6: 0x666D91)
-    static let waiting    = Color(hex6: 0xE9A63C) // amber — needs you
-    static let working    = Color(hex6: 0x5B9BE8) // blue — running
-    static let died       = Color(hex6: 0xE2584E) // red — stopped / exited
-    static let idle       = textFaint
+private enum WidgetPalette {
+    static let ground = Color(hex6: 0x13162A)
+    static let hairline = Color(hex6: 0x2E3358)
+    static let text = Color(hex6: 0xEEF0F7)
+    static let textDim = Color(hex6: 0x99A0BC)
+    static let textFaint = Color(hex6: 0x666D91)
+    static let waiting = Color(hex6: 0xE9A63C)
+    static let died = Color(hex6: 0xE2584E)
+    static let working = Color(hex6: 0x5B9BE8)
 
     static func color(_ status: AgentActivityAttributes.Status) -> Color {
         switch status {
         case .needsYou: return waiting
-        case .working:  return working
-        case .stopped:  return died
-        case .idle:     return idle
+        case .working: return working
+        case .idle: return textFaint
+        case .stopped: return died
         }
     }
 }
 
+private extension AgentActivityState {
+    var markIsUnconfirmed: Bool {
+        needsYouCount > 0 && unconfirmedCount >= needsYouCount
+    }
+
+    var markColor: Color { WidgetPalette.color(status) }
+
+    var displayHeadline: String {
+        needsYouCount == 0 ? "Nothing needs you" : headline
+    }
+
+    func activityHeadline(isStale: Bool) -> String {
+        guard isStale, needsYouCount > 0 else { return displayHeadline }
+        return "\(needsYouCount) may need you"
+    }
+
+    var deepLinkURL: URL { AgentActivityDeepLink.url(agentID: agentID) }
+}
+
+private extension Int {
+    var compactCount: String { self >= 100 ? "99+" : String(self) }
+}
+
 private extension Color {
-    /// Builds a colour from a 0xRRGGBB literal (sRGB), matching DesignSystem's helper.
     init(hex6: UInt32) {
         self.init(
             .sRGB,
-            red:   Double((hex6 >> 16) & 0xFF) / 255,
+            red: Double((hex6 >> 16) & 0xFF) / 255,
             green: Double((hex6 >> 8) & 0xFF) / 255,
-            blue:  Double(hex6 & 0xFF) / 255,
+            blue: Double(hex6 & 0xFF) / 255,
             opacity: 1
         )
     }
--- a/Shared/AgentActivityState.swift
+++ b/Shared/AgentActivityState.swift
@@ -52,26 +52,32 @@ struct AgentActivityState: Codable, Hashable {
     /// A plain Double (not Date) so the daemon-pushed JSON decodes identically,
     /// dodging Swift's Date reference-date Codable strategy. `nil` when not working.
     var workingSince: Double?
+    /// When the headline agent entered the blocked state (Unix seconds). The widget
+    /// renders this with a system timer so the waiting age advances without pushes.
+    var blockedSince: Double?
+    /// Daemon-truncated prompt copy for the one-line expanded and Lock Screen layouts.
+    var question: String?
+    /// A daemon-owned safe answer. Nil means the activity must offer Open, not Approve.
+    var defaultAnswer: String?
+    /// Stable pane identifier used by the activity's deep link.
+    var agentID: String?
+    /// When this snapshot was produced (Unix seconds), for honest stale-state copy.
+    var updatedAt: Double?
 
-    /// Hand-written for TWO additive directions, both of which otherwise make the whole
-    /// activity undecodable, which ActivityKit answers by DROPPING the activity so it
-    /// can never be reclaimed or ended.
+    /// Hand-written because additive metadata and enum cases must not make an existing
+    /// Live Activity undecodable. ActivityKit drops an undecodable activity, so it could
+    /// then be neither reclaimed nor ended.
     ///
-    /// 1. An absent `unconfirmedCount`, from an activity persisted by, or a payload
-    ///    minted by, a build predating the field. A stored-property default does not
-    ///    help: Codable synthesis never consults it.
-    /// 2. An UNRECOGNISED `status` word, from a newer app or daemon pushing a bucket
-    ///    this widget binary lacks. Raw-value decoding throws `dataCorrupted` there,
-    ///    which is the same drop-the-activity outcome as case 1. HerdrKit already
-    ///    treats this hazard as first-class with `AgentStatus.unrecognised(String)`;
-    ///    this file cannot carry that arm without a new case, so it falls back to
-    ///    `.needsYou`, matching HerdrKit's rule that something uninterpretable
-    ///    SURFACES rather than sinking. A wrongly-attention-grabbing lock screen is
-    ///    recoverable; a silently dropped activity is not.
+    /// `unconfirmedCount` defaults to zero for payloads predating that count. The prompt,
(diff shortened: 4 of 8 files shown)


jerryfane:
> Agent: `herdr-kimi-review`
> Runtime: `kimi`
> Job: `local-review-herdr-kimi-review-18d4e0eac82346b9-1`

**Decision:** `failed`
**Summary:** error: failed to run prompt: provider managed:kimi-code has no credential configured
See log: /root/.gitmoot/cache/tools/read-only/7eb12dc5c7899343/home/.kimi-code/logs/kimi-code.log: exit status 1
**Diagnostics:** error: failed to run prompt: provider managed:kimi-code has no credential configured
See log: /root/.gitmoot/cache/tools/read-only/7eb12dc5c7899343/home/.kimi-code/logs/kimi-code.log: exit status 1

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d4e15a2a929aaa-1`

**Decision:** `failed`
**Summary:** OMP requires a provider-qualified model such as provider/model; got ""
**Diagnostics:** OMP requires a provider-qualified model such as provider/model; got ""

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d4e25a907bf7d0-1`

**Summary:** Reviewed PR #248 at exact head 0c61c63379d448d8c62e0b691ff49a02044ede20 (worktree .git/HEAD matches; PR head.sha matches; 11 files, base main). The additive AgentActivityState decoder, the deep-link round-trip, GramStaging.stageData, the widget font wiring, and the photo upload/post/prompt idempotency ladder are all correct. Two P1 rendering defects block merge. (1) staleDate is now Date()+90s at both ActivityKit sites (LiveActivityController.swift:63,90) while the only update() call site is change-driven (.onChange(of: fullList) at HerdrApp.swift:2426) and no heartbeat exists anywhere in the client -- so 90s after a STABLE roster the activity renders 'Connection stale' and '<host> unreachable' and hedges the count to 'N may need you', while the session is healthy and even foregrounded. That is a false connectivity assertion in exactly the scenario this redesign exists for: one agent blocked, waiting for you. (2) The new widget-private displayHeadline discards the real headline whenever needsYouCount == 0, so a DIED agent renders a red stopped mark above the words 'Nothing needs you' (agent name gone), a working agent loses its identity to the same constant, and LiveActivityController.connecting's 'Connecting...' becomes a confident 'Nothing needs you' before the first agent.list arrives. AgentList.swift:338-352 documents the red-Stopped-over-'2 working' self-contradiction this exact rule was written to remove, and Shared/AgentActivityState.swift:1-19 plus LiveActivityController.swift:110-125 document why display decisions must not live where no test can execute them -- these three new decisions (displayHeadline, activityHeadline, markIsUnconfirmed) sit in a private extension in HerdrWidgets, which no test target can see. Also P2: canPerformAction reads UIPasteboard.general.itemProviders on every edit-menu presentation (an unsolicited content read that prompts 'Allow Paste?' in production and which the green UI test structurally cannot catch, because the harness wri
[truncated]
**Severity:** `P1`

**Findings**
- **90s staleDate with change-only updates makes the activity claim the host is unreachable while it is healthy** (P1)
  - detail: Both ActivityKit content sites changed from staleDate: nil (never stale) to Date().addingTimeInterval(90). The only caller of update() is .onChange(of: fullList, initial: true) at HerdrApp.swift:2426; AgentList is Equatable, so an UNCHANGED roster produces no update. A grep over App/ and Sources/HerdrKit for Timer/refresh/heartbeat paths finds nothing that re-pushes content periodically, and the server pushes on notification events (change-driven too). Consequence: a roster that is stable -- which is precisely 'one agent is blocked, waiting for you', the state the whole redesign is built around -- goes isStale 90 seconds later. The new copy then asserts, as fact, things a missing push does not establish: ExpandedHeadline renders Text("Connection stale"), lockDetail renders "<hostLabel> unreachable" (and "last seen <relative>"), StatusMark hollows out, the Open button goes outlined, and activityHeadline(isStale:) downgrades 'N need you' to 'N may need you'. This fires even with the app connected in the FOREGROUND. Fix directions: drive staleDate from a real liveness signal (last successful transport poll), or re-push on a cadence shorter than the stale window, or keep the stale window but change the copy so it claims only what is known ('not updated recently'), never unreachability.
  - files:
    - App/LiveActivityController.swift:63
    - App/LiveActivityController.swift:90
    - App/HerdrApp.swift:2426
    - HerdrWidgets/AgentLiveActivity.swift
  - id: f1
- **displayHeadline overwrites the headline for every non-attention state: 'Nothing needs you' above a red stopped agent, and during connect** (P1)
  - detail: var displayHeadline: String { needsYouCount == 0 ? "Nothing needs you" : headline } and activityHeadline(isStale:) falls through to it for every non-stale state. AgentList.activityContent sets headline = lead?.title (the agent's own name) and needsYouCount = count of .needsYou + .unrecognised only -- a .stopped lead yields statusWord "stopped" with needsYouCount 0. So: (a) an agent that DIED renders the red stopped StatusMark above the words 'Nothing needs you', with the agent's name discarded; (b) a working lead renders 'Nothing needs you' above its elapsed timer, also losing the name; (c) LiveActivityController.connecting (headline 'Connecting...', status .idle, needsYouCount 0) renders 'Nothing needs you' during the handshake, before any roster is known -- an all-clear asserted with zero information, which AgentList.swift:351 explicitly says isQuiet refuses to do. AgentList.swift:338-352 names the red-Stopped-over-'2 working' contradiction as the defect this rule was written to remove; this reintroduces the same class one layer down. Compounding it: the override lives in a private extension inside HerdrWidgets, the one target no test can execute -- the exact hazard Shared/AgentActivityState.swift:1-19 was split out to fix and that LiveActivityController.swift:110-125 records as already shipped once. If the headline genuinely should change for a quiet roster, that decision belongs in AgentList.activityContent (or in Shared/, which AgentActivityStateTests can reach), not beside the layout.
  - files:
    - HerdrWidgets/AgentLiveActivity.swift
    - Sources/HerdrKit/AgentList.swift:353-385
    - App/LiveActivityController.swift:143-145
  - id: f2
- **canPerformAction performs an unsolicited pasteboard content read on every edit menu; the green UI test cannot catch the resulting Allow-Paste prompt** (P2)
  - detail: ReplyTextView.canPerformAction(_:withSender:) evaluates UIPasteboard.general.itemProviders.contains(where: { $0.hasItemConformingToTypeIdentifier(UTType.image.identifier) }) to decide whether to advertise paste. UIKit calls canPerformAction while BUILDING the edit menu -- before the user has asked to paste -- and reading .itemProviders is a content read, not a metadata query. For a pasteboard written by ANOTHER app (a photo copied from Photos or Safari: the only real use case here) that read raises the system 'Allow Paste?' prompt / paste banner on every long-press of the reply composer. The non-alerting API for this check is UIPasteboard.general.hasImages (or detectPatterns), which is what the override should use; the read inside paste(_:) is user-initiated and fine as-is. Note why CI cannot see this: TerminalInteractionHarness's 'photo-pasteboard' command writes UIPasteboard.general.image from the app under test, and same-app reads are exempt from the prompt -- so testCopiedPhotoOffersPasteAndSendsAttachment passes while the production path prompts. This also reverses the privacy posture recorded for PR #244 ('No programmatic pasteboard reads were added ... matching the stated privacy goal').
  - files:
    - App/HerdrApp.swift
  - id: f3
- **onFocusChange mutates @State from inside updateUIView on two common paths** (P2)
  - detail: TerminalReplyField.updateUIView calls resignFirstResponder() (isFocused false branch) and assigns view.isEditable = isEnabled, both of which end editing synchronously and therefore run Coordinator.textViewDidEndEditing -> parent.onFocusChange(false) -> replyFocused = false while SwiftUI is still in its update pass. Two routine triggers: (a) the collapse-keyboard chevron sets replyFocused = false (HerdrApp.swift:4826; same at :4246 and :4888), the re-render calls resignFirstResponder, the delegate writes replyFocused again during that update; (b) any send sets sending = true, which flips isEnabled false, which sets isEditable = false on a first-responder text view, ending editing inside updateUIView. SwiftUI reports this as 'Modifying state during view update, this will cause undefined behavior'; @State has no same-value short circuit, so the write always publishes. Minimum fix: guard the callback (if replyFocused != newValue) and defer the write off the update pass; better, do not toggle isEditable to express 'sending'.
  - files:
    - App/HerdrApp.swift
  - id: f4
- **Every send disables the composer, so the keyboard drops for the whole round trip and focus is not restored** (P2)
  - detail: isEnabled: !replyDictating && !sending && !loadingReplyPhoto, and updateUIView applies it as view.isEditable. The base TextField was disabled only on replyDictating, never on sending. sending is held across the network round-trip, and in the photo path across upload + post + prompt + a 300ms sleep + refresh. Setting isEditable = false on an editing UITextView ends editing, so the software keyboard visibly dismisses on every reply; nothing re-focuses afterwards (the isFocused branch cannot re-acquire while isEditable is still false, and replyFocused has by then been set false by f4). Consecutive replies therefore require re-tapping the field each time. [INFERENCE on the UIKit end-editing mechanic -- no simulator available in this seat; the !sending gate and the isEditable assignment are both in the diff.]
  - files:
    - App/HerdrApp.swift:4841
  - id: f5
- **Return/Command-Return semantics from #247 are reversed, typed multiline replies become impossible, and none of it is in the PR description** (P2)
  - detail: Coordinator.textView(_:shouldChangeTextIn:replacementText:) returns false for replacement == "\n" and submits instead, and the send button's .keyboardShortcut(.return, modifiers: .command) is deleted. PR #247 -- approved two heads ago at d3f6bc3 -- deliberately made Return insert a newline and Command+Return send, 'matching Gram and keeping hardware-keyboard submission available'. After this head a newline can no longer be TYPED into a reply at all (only pasted, since a multi-character replacement containing newlines is not intercepted), and hardware Command+Return no longer sends anything. That may be the intended product call, but the PR body describes only the Island/Lock Screen redesign; the whole UIKit composer rewrite, the photo-paste attachment feature, and this input-semantics reversal are undisclosed scope in a PR titled 'Redesign Dynamic Island and Lock Screen activity'. Please state the reversal explicitly (and confirm the owner wants it) rather than landing it as a side effect of a widget redesign.
  - files:
    - App/HerdrApp.swift
  - id: f6
- **testReplyContainingOnlyNewlinesIsNotSendable no longer tests the invariant it names** (P2)
  - detail: The test types "\n\n" and asserts no send button. With the new shouldChangeTextIn interceptor each "\n" is consumed and returns false, so the newlines never enter the buffer: the test now passes because newlines are uninsertable, not because canSend trims .whitespacesAndNewlines. The predicate #247 added -- and that a prior review specifically accepted this test as pinning -- is now reachable only via paste and has zero coverage. Either re-point the test at the paste route (paste "\n\n" via the harness pasteboard, then assert no send button) or delete it; leaving a green test that names an invariant it can no longer fail on is worse than having none.
  - files:
    - HerdrUITests/TerminalControlTests.swift:235-248
  - id: f7
- **question is rendered but never produced and is erased by the next local update; defaultAnswer is dead** (P3)
  - detail: question is the primary subtitle in both ExpandedHeadline and LockScreenView, but LiveActivityController.state(from:) does not set it (AgentList.ActivityContent has no such field), so the app can never produce one and -- worse -- the first foreground roster change after a daemon push replaces the content state and ERASES a pushed question line. defaultAnswer is fully dead: declared, documented, decoded with decodeIfPresent, asserted nil in a test, never written and never read by any view. Either wire both through AgentList.ActivityContent, or drop defaultAnswer until the daemon endpoint it is waiting for exists (the PR body already concedes that endpoint does not exist).
  - files:
    - Shared/AgentActivityState.swift:58-61
    - App/LiveActivityController.swift:126-140
  - id: f8
- **postReplyPhoto reintroduces the unreachable-trailing-throw retry shape and duplicates GramView's post retry** (P3)
  - detail: for attempt in 0..<3 { ... } followed by throw GramError.invalidFileData: the loop either returns or throws on attempt 2, so the trailing throw is unreachable and reports a wrong error type if the compiler's reasoning ever changes. The PR #244 review had this project convert exactly this shape to a while-loop for exactly this reason ('the upload_in_progress retry loop was converted from a for-loop with unreachable trailing throw to a while-loop, preserving the exact same attempt/sleep semantics'). Semantics here are correct (3 attempts, 2 one-second sleeps, rethrow on the third), but this is a second copy of gram-post retry logic, now in App/ where no test can reach it, rather than a reuse of the tested one.
  - files:
    - App/HerdrApp.swift:5135-5152
  - id: f9
- **paste(_:) never falls through to super, so mixed image+text items lose their text and a non-intent pane cannot paste text at all** (P3)
  - detail: ReplyTextView.paste(_:) hands the first image-conforming provider to onPasteImage and returns without calling super.paste(sender). An item carrying both an image and text (copying an image from Safari registers public.png plus public.url) therefore drops the text entirely. And when pasteReplyPhoto's guard fails -- pane not .intent, or the agent has no name -- the paste becomes a no-op plus actionNote 'Photo attachments need a named agent with a ready prompt', so with an image anywhere on the pasteboard, plain-text paste into the composer is impossible on a rawKeys pane. Prefer image handling only when no text representation exists, and fall through to super on every bail-out path.
  - files:
    - App/HerdrApp.swift
  - id: f10
- **updatedAt mixes device and daemon clocks; widget bundle carries unregistered fonts; two Island metrics unverifiable here** (P3)
  - detail: updatedAt is stamped Date().timeIntervalSince1970 on local updates but comes from the daemon on pushes, and lockDetail renders it as Text(..., style: .relative) ('last seen ...'), so clock skew produces a nonsense or negative age. Separately, project.yml adds App/Fonts as a resources path to the widget target, which copies all 10 files (Geist-Bold/Medium, IBMPlexMono-Bold, and both OFL .txt files included) while UIAppFonts registers only the 5 the widget uses -- list the 5 files explicitly instead of the directory. Two layout details I could not verify without a device and flag only as risks: FleetTotals puts three 11pt lines in the narrow expanded-trailing region, and the minimal presentation wraps a 14pt StatusMark in .frame(width: 36, height: 36), which is wider than the minimal slot.
  - files:
    - App/LiveActivityController.swift:137
    - project.yml:299-320
    - HerdrWidgets/AgentLiveActivity.swift
  - id: f11

**Tests Run**
- static review only -- bash and eval are approval-gated in this seat, so no build, no swift test, no simulator run; I did not execute any command I cite
- verified exact head: .git/HEAD = 0c61c63379d448d8c62e0b691ff49a02044ede20 and GitHub PR #248 head.sha identical
- read the complete PR diff via the GitHub API files endpoint (11 files: App/HerdrApp.swift +463/-27, HerdrWidgets/AgentLiveActivity.swift +327/-127, Shared/AgentActivityState.swift +61/-18, App/LiveActivityController.swift, App/TerminalInteractionHarness.swift, HerdrUITests/TerminalControlTests.swift, Sources/HerdrKit/AgentList.swift, Sources/HerdrKit/GramStaging.swift, two test files, project.yml)
- traced the live-activity update graph in the worktree: grepped App/ and Sources/HerdrKit for liveActivity/LiveActivityController/activityContent/staleDate/Timer/refreshInterval; confirmed exactly two staleDate sites and exactly one update() call site (HerdrApp.swift:2426, .onChange(of: fullList)), with no periodic re-push
- cross-checked every widget font PostScript name against App/DesignSystem.swift:71-89 (IBMPlexMono, -Medm, -SmBld, Geist-Regular, Geist-SemiBold all match the app's proven names) and globbed App/Fonts to confirm all 5 registered filenames exist
- traced the photo path end to end: pasteReplyPhoto -> GramView.Staging.copy/GramStaging.stageData (GramView.swift:1586-1609) -> finishReplyPhotoPaste -> send() intercept -> sendPromptWithPhoto -> postReplyPhoto -> submitPrompt (HerdrApp.swift:5190-5203); confirmed uploadID/messageID latch-back, single staged-dir delete after a successful post, and retry-without-re-upload on prompt failure
- confirmed deep-link type agreement: AgentList.activityContent agentID = lead?.info.paneID -> AgentActivityDeepLink.url -> onOpenURL -> PushCenter.tapped(paneID:) (PushCenter.swift:40), and that project.yml registers the herdrup scheme under CFBundleURLTypes
- read GITMOOT_PRIOR_VERDICTS (/root/.gitmoot/.../evidence/prior-verdicts.json, as_of 2026-09-13T12:51:01Z, 25 verdicts): no prior verdict exists for PR #248 at any head, so this is a first review; used the #244 and #247 entries as the source for the pasteboard-privacy posture and the Return/Command-Return decision this head reverses
- CI status at this head taken from the job instructions (run 34754811001 green), not from any run of my own

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d4e65811c7cfc0-1`

**Summary:** Re-reviewed jerryfane/herdrup PR #248 at head 4dce5bd91d5da630211f6c6958a41a03d4d3dd24 (worktree .git/HEAD == 4dce5bd9..., single reflog entry: 'clone: from .../readonly-seat' at that exact sha, so no stale-checkout risk; no origin/main or branch refs exist in .git/packed-refs, only tags). EVIDENCE MODE IS static_only AND NOT BY CHOICE: bash, eval and task were all refused in this seat ('Tool "bash" requires approval but no interactive UI available'), so no git plumbing (cat-file/rev-parse/diff), no swift build, no swift test, no simulator run could be executed, and I could not compute the changed-file list; the review is code reading with grep/read at this head. All six blockers are closed: (f1) staleDate: nil at LiveActivityController.swift:72 and :101 with the remaining widget copy reduced to 'Not updated recently' (AgentLiveActivity.swift:132) and 'No update from <host>' (:261); (f2) zero occurrences of displayHeadline/activityHeadline anywhere (only an explanatory comment at AgentLiveActivity.swift:372-380), both surfaces render state.headline (:121, :181), markIsUnconfirmed now in Shared/AgentActivityState.swift:126-128 pinned by AgentActivityStateTests.swift:153-167 and wired into a real Linux test target (Package.swift:69-75); (f3) canPerformAction reads the non-prompting detection property hasImages (HerdrApp.swift:3879-3884); (f10) paste(_:) falls through to super on no-image/text-present/handler-declined (:3886-3901) with onPasteImage typed (NSItemProvider) -> Bool (:3749, :4937); (f4) textViewDidEndEditing guards on parent.isFocused and hops the write off the update pass (:3820-3831); (f5) the field is disabled only while dictating (:4867); (f7) the newline test now pastes via a new harness command and asserts the premise (TerminalControlTests.swift:240-261 + TerminalInteractionHarness.swift:472-473,525). ONE NEW P2: the f5 fix leaves the composer editable and focused for the whole round trip while `reply = ""` at HerdrApp.swift:5081 (and :5154) is still
[truncated]
**Severity:** `P2`

**Findings**
- **The f5 fix makes the composer editable during a send, but `reply = ""` after the await is still unconditional, so a reply typed during the round trip is silently wiped** (P2)
  - detail: f5 was fixed by narrowing the gate to `isEnabled: !replyDictating` (:4867), which is correct for the keyboard. It also removes the only thing that made the deferred clear safe. `sendTapped` -> `send(.submitText(text))` snapshots the outgoing text into `plan`, then inside the Task at :5064 does `try await submitPrompt(...)` — which calls `client.prompt(..., waitUntil: anyAgentStatus, timeoutMs: 6000)` (:5238-5242) — and only afterwards executes `if case .submitText = action { reply = "" }` (:5081). During that window the field is editable (nothing sets isEditable=false anymore) and still first responder (the Return path at :4873-4880 does not clear replyFocused; only the phone arrow-tap at :4914 does), which is exactly the flow the f5 fix exists to enable: press Return, keep typing the next reply. When the server confirms, the unconditional clear deletes the newly typed text. The same shape exists on the photo path (`reply = ""` at :5154, after gramUploadFile + gramPost + submitPrompt, i.e. a much longer window). The user cannot rescue the text by sending it either: `onReturn` bails while `sending` (:4874) and `canSend` includes `!sending` (:4135), so the text sits unsendable until the exact moment it is erased. At the prior head the field was disabled for the whole send, so the unconditional clear could not lose anything; this is a regression introduced by this head.
  - file: App/HerdrApp.swift
  - line: 5081
  - rationale: Clear only what was actually sent — e.g. capture the submitted string and do `if reply == sentText { reply = "" }`, or clear optimistically at submit time and restore on throw. No test covers it: HerdrUITests has no consecutive-send/keyboard-survives case (grep of TerminalControlTests for terminal-send-button shows taps only in the photo test at :293-299), so neither the f5 fix nor this loss has a receipt.
  - state: open
  - uid: jerryfane/herdrup#248-new-replyclear
- **90s staleDate with change-only updates makes the activity claim the host is unreachable while it is healthy** (P1)
  - continues_uid: jerryfane/herdrup#248-f1
  - detail: CLOSED at this head. Checked by reading both ActivityKit call sites: `Activity.request(attributes:content: ActivityContent(state: state, staleDate: nil), pushType: .token)` at :72 and `activity.update(ActivityContent(state: state, staleDate: nil))` at :101, each with the reasoning comment (:62-70, :99-100). A repo-wide grep for `staleDate` returns only those two sites (the other `isStale` hits are HerdrKit's unrelated StreamLiveness and the widget's `context.isStale` plumbing). The remaining stale wording asserts nothing about reachability it cannot know: 'Not updated recently' (HerdrWidgets/AgentLiveActivity.swift:132), 'last update <relative> · no update from <host>' (:246-256) and 'No update from <host>' (:258-262). Consequence worth noting, not a defect: with staleDate nil at both sites, `context.isStale` can now only be raised by an APNs payload's stale-date, so every isStale branch in the widget is unreachable through the in-app path.
  - line: 72
  - state: answered
  - uid: jerryfane/herdrup#248-f1
- **displayHeadline overwrites the headline for every non-attention state: 'Nothing needs you' above a red stopped agent, and during connect** (P1)
  - continues_uid: jerryfane/herdrup#248-f2
  - detail: CLOSED at this head. A repo-wide grep for `displayHeadline|activityHeadline|Nothing needs you` returns a single hit — the explanatory comment at AgentLiveActivity.swift:372-380 — and no code. Both surfaces now render the daemon/HerdrKit headline directly: `Text(state.headline)` in ExpandedHeadline (:121) and in LockScreenView (:181); the quiet-roster wording is only `AgentActivitySummary.line(state)` and only under `state.needsYouCount > 0` (:192-195), so a stopped agent keeps its own name over the red mark and 'Connecting…' (LiveActivityController.swift:155) is no longer answered with an all-clear. `markIsUnconfirmed` moved to Shared/AgentActivityState.swift:126-128 and is executed by a test: AgentActivityStateTests.swift:153-167 pins true for 1/1 and 2/2, false for 2/1, 0/0 and 0/3 — and that target genuinely runs, because Package.swift:69-75 declares the ActivityKit-free `AgentActivityState` target (path Shared, sources limited to the one file) plus its testTarget. I could not execute `swift test` to see it pass (bash/eval/task refused).
  - line: 121
  - state: answered
  - uid: jerryfane/herdrup#248-f2
- **canPerformAction performs an unsolicited pasteboard content read on every edit menu; the green UI test cannot catch the resulting Allow-Paste prompt** (P2)
  - continues_uid: jerryfane/herdrup#248-f3
  - detail: CLOSED at this head. `canPerformAction` now tests `UIPasteboard.general.hasImages` (:3880) — a detection property, which by contract answers from item type metadata and does not trigger the paste notification — and otherwise defers to `super` (:3883). The only content read (`itemProviders`) is inside `paste(_:)` at :3893, i.e. after the user chose Paste. Unchanged, pre-existing and not a regression: forcing `true` for `paste:` whenever the pasteboard holds an image also offers Paste while the view is non-editable (dictation), and `pasteReplyPhoto` does not gate on `replyDictating` (:4938-4942); the prior head had the same reachability with the prompting read.
  - file: App/HerdrApp.swift
  - line: 3879
  - state: answered
  - uid: jerryfane/herdrup#248-f3
- **onFocusChange mutates @State from inside updateUIView on two common paths** (P2)
  - continues_uid: jerryfane/herdrup#248-f4
  - detail: CLOSED at this head for both named paths. `textViewDidEndEditing` (:3820-3831) now (a) returns early unless `parent.isFocused`, which kills the equal-value publish on the resign path — `parent` is refreshed at the top of `updateUIView` (:3785), so the declared state is already false when the resign at :3794 calls back — and (b) defers the genuine false write with `Task { @MainActor in notify(false) }`, off the update pass. I traced the isEditable=false path too (dictation): isEditable=false ends editing, the queued Task lands after the pass, and the following pass resigns and hits the guard, so it converges rather than ping-pongs. Residual I checked and am NOT filing as a defect: the symmetric begin path is still synchronous (`updateUIView` -> `becomeFirstResponder()` at :3796 -> `textViewDidBeginEditing` -> `parent.onFocusChange(true)` at :3817), but it is unreachable, because nothing in the app sets the focus flag true programmatically — grep for `replyFocused =` yields only :4265, :4845, :4914 (all `false`) and the binding at :4869 — so `becomeFirstResponder` from the update pass only happens after UIKit already began editing, when the view is already first responder.
  - file: App/HerdrApp.swift
  - line: 3820
  - state: answered
  - uid: jerryfane/herdrup#248-f4
- **Every send disables the composer, so the keyboard drops for the whole round trip and focus is not restored** (P2)
  - continues_uid: jerryfane/herdrup#248-f5
  - detail: CLOSED as written: `isEnabled: !replyDictating` (:4867) with the reasoning comment at :4857-4866; `sending` and `loadingReplyPhoto` are gone from the gate and double-sends stay prevented where they happen (`canSend` :4135, `onReturn` :4874). The dictation disable that TerminalControlTests.swift:174-176 uses as its 'dictation actually started' observable still holds, so that test's precondition is unaffected. Two caveats recorded rather than silently accepted: no receipt pins the new behaviour (no UI test sends twice and checks the keyboard/focus survives), and the now-editable in-flight window is what the new P2 above exploits.
  - file: App/HerdrApp.swift
  - line: 4867
  - state: answered
  - uid: jerryfane/herdrup#248-f5
- **Return/Command-Return semantics from #247 are reversed, typed multiline replies become impossible, and none of it is in the PR description** (P2)
  - continues_uid: jerryfane/herdrup#248-f6
  - detail: Code half CHECKED at this head and unchanged, which matches the claim that only the disclosure was fixed: `shouldChangeTextIn` intercepts any exactly-\n replacement and submits (`parent.onReturn(...); return false`, :3846-3851), `returnKeyType = .send` (:3768), and there is no newline-insert affordance — grep for `keyboardShortcut` finds only ⌘K/⌘//⌘=/⌘-/⌘0 (:1938-1952) and ⌘F/⌘R/Escape (:4496-4518), no Command-Return anywhere, and the vendored terminal's keyCommands are Ctrl-only (LiveTerminalView.swift:481-490). So typed multiline replies remain impossible and paste is the only route that puts a newline in the composer — which the f7 test now states explicitly (TerminalControlTests.swift:236-239). Disclosure half NOT VERIFIABLE from this seat: there is no network and gh is unauthenticated (`pr://248` returned 'GitHub CLI is not authenticated'), so I cannot read the PR body to confirm the composer scope and the Return reversal are now described. Not blocking on my side: the behaviour is deliberate, documented in-tree, and the gate can read the body directly.
  - file: App/HerdrApp.swift
  - line: 3846
  - state: answered
  - uid: jerryfane/herdrup#248-f6
- **testReplyContainingOnlyNewlinesIsNotSendable no longer tests the invariant it names** (P2)
  - continues_uid: jerryfane/herdrup#248-f7
  - detail: CLOSED at this head, and it now asserts the premise as well as the conclusion. The test seeds the pasteboard through the new harness command `newline-pasteboard` (TerminalInteractionHarness.swift:472-473, registered in the command grid at :525 so the fixture button exists), long-presses, taps Paste, then asserts `field.value == "\n\n"` with the message 'the newlines must actually reach the composer for this to test anything' (:253-254) BEFORE asserting the send button is absent (:255-256), and finally types real text to prove the button does appear (:258-260). I checked the arming path is real rather than incidental: the paste is not swallowed by the Return interception, because `shouldChangeTextIn` only diverts a replacement equal to "\n" (:3847) and a two-newline paste is inserted by `super.paste`; and the assertion pins the trimming rule specifically, because `hasReplyContent` is `replyPhoto != nil || !reply.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty` (:4131-4133) and drives the button swap at :4900. Could not execute the XCUITest (no simulator, no shell).
  - file: HerdrUITests/TerminalControlTests.swift
  - line: 240
  - state: answered
  - uid: jerryfane/herdrup#248-f7
- **paste(_:) never falls through to super, so mixed image+text items lose their text and a non-intent pane cannot paste text at all** (P3)
  - continues_uid: jerryfane/herdrup#248-f10
  - detail: CLOSED at this head. `paste(_:)` (:3886-3901) only consumes the gesture when all of `hasImages`, `!hasStrings`, an image-conforming provider, AND `onPasteImage?(provider) == true` hold; every other path reaches `super.paste(sender)`. The declining half is now expressible: `onPasteImage` is `(NSItemProvider) -> Bool` (:3749) and `pasteReplyPhoto` is `@discardableResult ... -> Bool` returning false for an in-flight send/load and for a pane with no named intent agent (:4936-4948), so a non-intent pane pastes text normally and still gets the 'Photo attachments need a named agent' note. Tradeoff introduced by the `!hasStrings` arm, documented in the comment as 'text wins': a photo copied from Safari, which registers a text/URL representation alongside the image, will now never attach — it always pastes as text. Acceptable as stated, but it is the opposite of the Safari case the comment cites as motivation, and no test covers the mixed pasteboard (the harness only sets `UIPasteboard.general.image`, TerminalInteractionHarness.swift:464-469).
  - file: App/HerdrApp.swift
  - line: 3886
  - state: answered
  - uid: jerryfane/herdrup#248-f10
- **question is rendered but never produced and is erased by the next local update; defaultAnswer is dead** (P3)
  - continues_uid: jerryfane/herdrup#248-f8
  - detail: Confirmed STILL OPEN at this head, as the PR body declares. `state(from: AgentList)` (:137-151) copies headline/status/counts/timers/agentID/updatedAt and sets neither `question` nor `defaultAnswer`, so every foreground update overwrites a push-supplied question with nil. A repo-wide grep for `defaultAnswer|\.question` finds readers only in the widget (`state.question` at AgentLiveActivity.swift:125 and :187) and declarations/tests in Shared — `defaultAnswer` has no reader at all, and there is no Approve affordance anywhere, so the 'Nil means offer Open, not Approve' contract at Shared/AgentActivityState.swift:60-61 is unexercised.
  - line: 137
  - state: open
  - uid: jerryfane/herdrup#248-f8
- **postReplyPhoto reintroduces the unreachable-trailing-throw retry shape and duplicates GramView's post retry** (P3)
  - continues_uid: jerryfane/herdrup#248-f9
  - detail: Confirmed STILL OPEN at this head, as the PR body declares. `postReplyPhoto` (:5164-5181) is unchanged: `for attempt in 0..<3` either returns the posted message or, on the third `upload_in_progress`, rethrows at `if attempt == 2 { throw error }`, which leaves `throw GramError.invalidFileData` after the loop reachable only to satisfy the compiler — and it would surface a misleading 'invalid file data' error if it ever executed. The duplicate of GramView's post retry also remains.
  - file: App/HerdrApp.swift
  - line: 5164
  - state: open
  - uid: jerryfane/herdrup#248-f9
- **updatedAt mixes device and daemon clocks; widget bundle carries unregistered fonts; two Island metrics unverifiable here** (P3)
  - continues_uid: jerryfane/herdrup#248-f11
  - detail: Confirmed STILL OPEN at this head, with one part now latent rather than visible. (a) Clock mixing unchanged: `updatedAt: Date().timeIntervalSince1970` (:150) is the device clock while the decoder accepts a daemon-minted value (Shared/AgentActivityState.swift:95); the only consumer is the relative-time line in `lockDetail` (AgentLiveActivity.swift:246-256), which sits behind `isStale` — and with staleDate nil at both request/update sites that branch is unreachable via the in-app path, so a skewed daemon clock can now only bite through an APNs stale-date. (b) Fonts: project.yml copies the whole `App/Fonts` folder as widget resources (:302-303) while the widget Info.plist registers five files (:315-320); the five PostScript names WidgetFont asks for — Geist-Regular, Geist-SemiBold, IBMPlexMono, IBMPlexMono-Medm, IBMPlexMono-SmBld (AgentLiveActivity.swift:338-342) — are covered, so the residue is unused bundle weight (Geist-Medium/Bold, IBMPlexMono-Bold), not a missing glyph. (c) The two Dynamic Island metrics remain unverifiable in this seat: no simulator and no shell (bash/eval/task all refused), so nothing about rendered geometry can be measured here.
  - line: 150
  - state: open
  - uid: jerryfane/herdrup#248-f11

**Tests Run**
- No commands executed. bash, eval and task were all refused in this seat with 'requires approval but no interactive UI available', so git plumbing (git cat-file -t / git rev-parse <sha>:<path> / git diff merge-base), swift build, swift test --filter AgentActivityStateTests and any xcodebuild/simulator run were impossible. Evidence mode is static_only for that reason, not by choice.
- Head identification without git: .git/HEAD contains 4dce5bd91d5da630211f6c6958a41a03d4d3dd24 (detached, no branch refs — .git/refs/{heads,remotes} empty, .git/packed-refs holds only v0.1.x tags), and .git/logs/HEAD has exactly one entry: 0000000 -> 4dce5bd91d5da630211f6c6958a41a03d4d3dd24 'clone: from /root/.gitmoot/worktrees/jerryfane--herdrup/delegations/local-review-herdr-ios-review-opus-18d4e65811c7cfc0-1/readonly-seat'. So the reviewed tree is a fresh clone pinned at the assigned head; the reused-stale-worktree failure mode cannot apply here. Because no base ref exists locally, the changed-file set could not be computed — findings cite files read at this head rather than a diff.
- Read in full at this head: App/LiveActivityController.swift (all 158 lines), HerdrWidgets/AgentLiveActivity.swift (all 398 lines), Shared/AgentActivityState.swift (all 201 lines), Package.swift, Tests/AgentActivityStateTests/AgentActivityStateTests.swift, App/HerdrApp.swift:3700-3960 (TerminalReplyField/Coordinator/ReplyTextView), :4130-4136 (hasReplyContent/canSend), :4480-4570 (paneKeyboardShortcuts), :4790-4990 (replyBar, pasteReplyPhoto), :4990-5240 (send, sendPromptWithPhoto, postReplyPhoto, handleReplyChange, submitPrompt), :5270-5320 (prefill), :5340-5360 (sendTapped), App/TerminalInteractionHarness.swift:430-530, HerdrUITests/TerminalControlTests.swift:150-352, project.yml (target/font sections).
- Greps used as checks: 'displayHeadline|activityHeadline|Nothing needs you' -> 1 hit, a comment at HerdrWidgets/AgentLiveActivity.swift:373 (no code) — f2; 'staleDate|isStale|markIsUnconfirmed' -> staleDate only at LiveActivityController.swift:72,101, both nil — f1; 'replyFocused =' -> 4265, 4845, 4914 all false plus the binding at 4869, no programmatic true — f4 begin-path reachability; 'defaultAnswer|\.question' -> readers only at AgentLiveActivity.swift:125,187, no producer in App/ — f8; 'reply = ""|sending = true' -> clears at HerdrApp.swift:5081, 5154, 5302 all after awaits — the new P2; 'testReplyContainingOnlyNewlines' + harness 'paste|newline' -> the new newline-pasteboard command and its registration — f7.
- PR body could not be read: pr://248 failed with 'GitHub CLI is not authenticated', so the f6 disclosure claim is unverified from this seat (its code half was verified in-tree).

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d4e6e28356bff9-1`

**Decision:** `approved`
**Summary:** f12 is closed at 06a3bbc2. Both send paths now clear only what was actually sent: App/HerdrApp.swift:5085 `if case .submitText(let sent) = action, reply == sent { reply = "" }` and App/HerdrApp.swift:5158 `if reply == text { reply = "" }`. I traced every producer of the compared string for a lost-clear regression and found none: sendTapped() (5359) passes `text ?? reply`; the only caller passing an explicit argument is onReturn (4879) with `textView.text` captured in shouldChangeTextIn (3847), and textViewDidChange (3833-3837) writes that same string through the binding to `reply` on every keystroke, so sent == reply at completion on the Return path as well as the button path; usePrompt (5046-5047) assigns `reply` then sends it; the plan-side trimming/refusal paths never reach the clear (`.refused` returns at 5079), and the comparison is against the raw action payload, not the plan text, so trimming cannot desynchronise it. No stale-composer case either: the only way the guard skips the clear is `reply != sent`, which means the composer holds text that was never sent. Residual behaviour worth knowing, not a defect: because the clear is deferred until after the await, the sent text stays visible for the whole round trip (seconds on the photo path: upload + post + prompt), so a user who appends to it ends up with sent-prefix + new text and a second send re-sends the prefix. Clearing earlier would lose text on a failed send, so the tradeoff is defensible. All prior P1/P2 rows check out at this head. Three P3s remain open as declared (f8, f9, f11) and I am adding two new P3s: the third clear site deliverPrefillOnce (5306) still clears unconditionally while the composer is editable, and textViewDidBeginEditing (3816) still writes the focus binding synchronously from inside updateUIView. Neither blocks. APPROVE.
**Severity:** `P3`

**Findings**
- **Deferred clear is now conditional on both send paths; no lost-clear or stale-composer regression found** (P2)
  - continues_uid: jerryfane/herdrup#248-f12
  - detail: Checked at 06a3bbc2 by reading App/HerdrApp.swift:5043-5167 and 5250-5360 and enumerating every producer of the `.submitText` payload. Line 5085 guards the clear with `reply == sent`; line 5158 guards the photo path with `reply == text`. Lost-clear analysis: sendTapped() (5359) sends `text ?? reply`; the sole explicit-argument caller is onReturn (4873-4879) which forwards `textView.text ?? ""` from shouldChangeTextIn (3846-3848), and Coordinator.textViewDidChange (3833-3837) assigns `parent.text = textView.text` on every edit, so the binding and the payload are the same string; usePrompt (5045-5048) sets `reply` then sends it; rawKeys `.submitText` still reaches the clear via the `.text` plan. Refusals and photo-guard bails return before the clear, which is correct because nothing was sent. Stale-composer analysis: the guard only skips when the composer differs from what was sent, i.e. when it holds unsent text. Note, not a defect: the clear stays deferred, so the sent text remains in the field for the whole round trip and an append during that window survives as sent-prefix + new text, which the next send re-sends; the alternative (clear at send time) loses text when the prompt fails, so the current tradeoff is the safer one.
  - file: App/HerdrApp.swift
  - line: 5085
  - rationale: The finding asked for the clear to be conditional on the composer still holding the sent text; both named paths now do exactly that, and the equality key cannot strand the composer because the compared string is always the composer's own value at send time.
  - state: answered
- **staleDate is nil on both the start and the update push** (P1)
  - continues_uid: jerryfane/herdrup#248-f1
  - detail: Read App/LiveActivityController.swift:55-102 at this head. `ActivityContent(state: state, staleDate: nil)` at line 72 (request) and line 101 (update), with the reason documented at 62-70 and 99-101: nothing re-pushes on a cadence (the only caller is `.onChange(of: fullList)`), so an elapsed window would assert unreachability from mere roster stability. The 90s window is gone from both call sites; no other staleDate occurrence exists in the repo (grep for `staleDate` returns only these four lines).
  - line: 72
  - rationale: The P1 was that a stale window plus change-only pushes makes a healthy host read as unreachable; removing the window removes the false claim.
  - state: answered
- **No headline override remains in the widget target; the headline is the lead agent's name from HerdrKit** (P1)
  - continues_uid: jerryfane/herdrup#248-f2
  - detail: Grepped `headline`/`Nothing needs you`/`needsYouCount == 0` across HerdrWidgets, App, Shared, Sources and Tests at this head. The widget renders `state.headline` unmodified (AgentLiveActivity.swift:121, 183); the only remaining mention of the override is the prohibition comment at 372-379. The value is decided in Sources/HerdrKit/AgentList.swift:376 (`headline: lead?.title ?? "No agents"`), App/LiveActivityController.swift:137-151 is a pure field copy, and the connect handshake keeps its own "Connecting…" state (LiveActivityController.swift:154-156). Tests execute the rule: ActivityContentTests.testTheHeadlineNamesTheLeadAgentOnANonEmptyRoster (273-284) pins the name against the empty-roster-constant mutant and AgentListTests.testAStoppedPaneIsTheHeadlineWhenNothingElseExists / testAConfirmedNeedsYouOutranksAnUnconfirmedOne pin the ranking. Not executed here (no toolchain in this sandbox).
  - line: 372
  - rationale: The P1 was an all-clear headline rendered over a red stopped mark and over the connect state; the override is deleted, the decision now lives where a test runs it, and the connect state is a separate constant.
  - state: answered
- **canPerformAction now answers from pasteboard metadata, not content** (P2)
  - continues_uid: jerryfane/herdrup#248-f3
  - detail: Read App/HerdrApp.swift:3873-3901. `canPerformAction` tests `UIPasteboard.general.hasImages` only (3880) and otherwise defers to `super` (3883); the `itemProviders` content read is confined to `paste(_:)` (3893), which is user-initiated. No `itemProviders`, `string`, `image`, or `items` access remains on the menu-building path, so the Allow-Paste prompt the UI test could not observe can no longer be raised by opening the edit menu.
  - file: App/HerdrApp.swift
  - line: 3880
  - rationale: The P2 was an unsolicited content read while UIKit builds the menu; hasImages answers the same predicate without a content read.
  - state: answered
- **The end-editing focus writes now hop off the update pass; both cited paths are fixed** (P2)
  - continues_uid: jerryfane/herdrup#248-f4
  - detail: Read App/HerdrApp.swift:3776-3850. Both paths that end editing from inside updateUIView — `resignFirstResponder()` (3792) and `isEditable = isEnabled` (3785) — land in textViewDidEndEditing (3820-3831), which now guards on `parent.isFocused` and dispatches the binding write through `Task { @MainActor in notify(false) }` instead of writing during the pass. The `nativeFocusPendingStateSync` flag (3809, 3816, 3821, 3787, 3791) keeps a native tap's responder alive across text-binding renders so the hop cannot resign a field the user just focused. One synchronous write remains on the begin path; I report it separately as a new P3 rather than fold it into this row, because it is a different call site and an idempotent write.
  - file: App/HerdrApp.swift
  - line: 3820
  - rationale: The two paths this finding named write asynchronously now and are additionally short-circuited when the declared state already agrees.
  - state: answered
- **The composer stays editable and focused across a send; only dictation disables it** (P2)
  - continues_uid: jerryfane/herdrup#248-f5
  - detail: Read App/HerdrApp.swift:4856-4926. `isEnabled: !replyDictating` (4867) — neither `sending` nor the photo upload nor `pendingPrefill` gates editability any more, so `isEditable = false` is never set on a first-responder text view mid-send and the keyboard no longer drops; `isFocused: replyFocused` (4868) keeps the responder. Double-send is still prevented where it can happen: `canSend` excludes `sending` (4134-4136) and disables the button (4922), and `onReturn` bails on `sending` (4874). This is also the precondition for the f12 fix, which is why the conditional clear was required.
  - file: App/HerdrApp.swift
  - line: 4867
  - rationale: The P2 was a keyboard drop for the whole round trip with no focus restoration; the gate that caused it is removed and focus is preserved by the same binding.
  - state: answered
- **Return submits deliberately and multiline input survives via paste; code semantics verified, PR body not readable offline** (P2)
  - continues_uid: jerryfane/herdrup#248-f6
  - detail: Read App/HerdrApp.swift:3844-3849 and 4873-4880 at this head. The interception is exact-equality — `guard replacement == "\n" else { return true }` — so only a bare Return submits; a multiline paste arrives as one replacement string containing newlines and is inserted untouched, which HerdrUITests/TerminalControlTests.testReplyComposerGrowsUpwardThenScrollsWithoutMovingSend (335-348, `reply-multiline-pasteboard`) exercises, and `returnKeyType = .send` (3768) labels the key accordingly. So typed-Return-as-send is deliberate and multiline text is not impossible, only not typeable. I cannot verify the PR description at this head: this reviewer has no network and no git plumbing (bash/eval denied), so the declaration half of this finding rests on the prior round's record, not on my own observation.
  - file: App/HerdrApp.swift
  - line: 3846
  - rationale: The behaviour is now internally consistent and test-covered, and the newline guard is narrow enough that it does not swallow pastes; only the PR-body claim is outside what I can check here.
  - state: answered
- **testReplyContainingOnlyNewlinesIsNotSendable arms its premise via the pasteboard and asserts it** (P2)
  - continues_uid: jerryfane/herdrup#248-f7
  - detail: Read HerdrUITests/TerminalControlTests.swift:234-260. The test now loads a `newline-pasteboard` fixture (244), pastes it, and asserts `field.value == "\n\n"` with the message "the newlines must actually reach the composer for this to test anything" (253-254) BEFORE asserting the send button is absent (255-256), then types "message" and requires the button to appear (258-259). The premise is therefore armed and pinned, so the test can no longer pass because newlines are uninsertable; it tests the `canSend` trimming rule it names. Not executed (no simulator or toolchain in this sandbox).
  - file: HerdrUITests/TerminalControlTests.swift
  - line: 240
  - rationale: The finding was a vacuous test whose setup could not reach the condition; the premise assertion plus the positive-control typing step closes exactly that.
  - state: answered
- **paste(_:) now falls through to super on every bail-out** (P3)
  - continues_uid: jerryfane/herdrup#248-f10
  - detail: Read App/HerdrApp.swift:3886-3901. The image branch requires `hasImages && !hasStrings` and an `onPasteImage` handler returning true; every other outcome — no image, text also present (the Safari image+URL case), or a handler that declines because there is no named intent agent — reaches `super.paste(sender)` at 3900. Mixed image+text keeps its text and a non-intent pane can paste text again.
  - file: App/HerdrApp.swift
  - line: 3892
  - rationale: The finding was a swallowed paste with no super fallthrough; all three bail-out paths now reach UIKit.
  - state: answered
- **question is still rendered but never produced locally, and defaultAnswer is still dead** (P3)
  - continues_uid: jerryfane/herdrup#248-f8
  - detail: Unchanged at 06a3bbc2. `LiveActivityController.state(from:)` (137-151) builds ContentState without `question` or `defaultAnswer`, and it is the only local producer, so every `.onChange(of: fullList)` update erases a question that arrived by push; the widget renders it at HerdrWidgets/AgentLiveActivity.swift:125 and 187. `defaultAnswer` (Shared/AgentActivityState.swift:61, 93, 113) has no reader anywhere — grep across App, HerdrWidgets, Shared and Sources finds only its declaration, decode and init. Declared in the PR body and P3, so it does not hold the merge.
  - line: 149
  - rationale: Still a rendered field with no local producer and a decoded field with no reader; recorded, not blocking.
  - state: open
- **postReplyPhoto still ends in an unreachable trailing throw** (P3)
  - continues_uid: jerryfane/herdrup#248-f9
  - detail: Unchanged at 06a3bbc2. App/HerdrApp.swift:5174-5186: the `for attempt in 0..<3` loop either returns, rethrows a non-retryable APIError, or throws on `attempt == 2`, so `throw GramError.invalidFileData` at 5185 cannot execute and reports a misleading error kind if the loop bounds ever change. Still duplicates GramView's post retry rather than sharing it. P3 and declared.
  - file: App/HerdrApp.swift
  - line: 5185
  - rationale: Same shape as the earlier occurrence; a compiler-satisfying terminator that lies about the failure mode, cheap to fix but not merge-blocking.
  - state: open
- **updatedAt is still a device clock beside daemon-derived timestamps; widget fonts ARE registered; Island metrics still unverifiable here** (P3)
  - continues_uid: jerryfane/herdrup#248-f11
  - detail: Checked all three sub-claims at this head. (1) Clock mixing stands: `updatedAt: Date().timeIntervalSince1970` (LiveActivityController.swift:149) is the device clock while `workingSince`/`blockedSince` come from daemon `completedUnixMs` via AgentList.activityContent, so the honest-stale copy compares two clocks. (2) The font sub-claim no longer holds: project.yml:296-320 gives HerdrWidgets `App/Fonts` as a resources build phase (302-303) and registers Geist-Regular, Geist-SemiBold, IBMPlexMono-Regular, IBMPlexMono-Medium, IBMPlexMono-SemiBold under UIAppFonts, which covers every PostScript name WidgetFont requests (AgentLiveActivity.swift:338-342: Geist-Regular, Geist-SemiBold, IBMPlexMono, IBMPlexMono-Medm, IBMPlexMono-SmBld). (3) The two Island metrics remain unverifiable in this environment — no device, simulator, or toolchain.
  - line: 149
  - rationale: Two of three parts still open; the font part is answered by the widget target's own Info.plist registration, and the metrics part is unfalsifiable without hardware rather than wrong.
  - state: open
- **deliverPrefillOnce is the third clear site and is still unconditional, so text typed during the auto-deliver round trip is wiped** (P3)
  - detail: New at this head only in the sense that the f12 fix did not cover it: App/HerdrApp.swift:5300-5313 awaits `client.prompt(pane: paneID, text: reply)` (5305) — capturing the composer value as an argument — then executes `reply = ""` (5306) with no `reply == sent` guard. The composer is editable throughout, because `isEnabled: !replyDictating` (4867) does not gate on `autoDelivering` after the f5 fix, and `canSend` (4134-4136) does not either. A character typed during that await is neither sent nor kept. The code already acknowledges this hazard for one input source only: the MicButton comment at 4893-4895 says isActive drops on `autoDelivering` "so an in-flight dictation stops before the auto-deliver clears the reply" — dictation was gated, typing was not. Same one-line remedy as 5085: capture the sent text and compare before clearing.
  - file: App/HerdrApp.swift
  - line: 5306
  - rationale: Identical defect class to f12 at a site the fix missed; narrow window (a pending pre-fill's delivery attempt) and silent rather than destructive to sent data, so P3 rather than P2.
  - state: open
- **textViewDidBeginEditing still writes the focus binding synchronously from inside updateUIView** (P3)
  - detail: App/HerdrApp.swift:3786-3790 calls `view.becomeFirstResponder()` from inside updateUIView, which begins editing synchronously and reaches textViewDidBeginEditing (3815-3818), where `parent.onFocusChange(true)` writes `replyFocused` (4869) during SwiftUI's own update pass — the exact pattern the sibling handler documents as "Modifying state during view update, this will cause undefined behavior" (3822-3827). It is benign in effect: the branch only runs when `isFocused` is already true, so the write is equal-valued and cannot diverge or loop; but the same comment notes @State has no same-value short circuit, so it still publishes and still trips the runtime warning. The remedy that already exists for the end path applies verbatim: guard on the declared state and hop off the pass.
  - file: App/HerdrApp.swift
  - line: 3816
  - rationale: Residual instance of the state-during-update pattern at a call site the f4 fix left synchronous; cosmetic and idempotent, so it is recorded rather than blocking.
  - state: open

**Tests Run**
- Reviewed SHA 06a3bbc2c960cc16a5a8813d1ccb96bf1717cee3, read from .git/HEAD in the assigned worktree /root/.gitmoot/execbackends/local/8fbf668d05bcc4f828613d5f2f5ec5e1/workspace; it equals the head named in the dispatch. Per the standing header I could not run the git plumbing (git cat-file / git diff / git rev-parse <sha>:<path>) or produce a blob-id discriminator: both the bash and eval tools are refused by this session's approval policy ('requires approval but no interactive UI available'), so the head was established by reading the ref file directly and every claim below cites exact lines at that tree.
- No build and no test execution: no Xcode toolchain, no simulator, and no shell in this sandbox. Evidence mode is static_only; the test files cited (HerdrKitTests/ActivityContentTests.swift, AgentListTests.swift, AgentActivityStateTests.swift, HerdrUITests/TerminalControlTests.swift) were read, not run.
- f12 verification: read App/HerdrApp.swift:5043-5167 (send(_:), sendPromptWithPhoto, postReplyPhoto) and 5250-5360 (deliverPrefillOnce, deliverPrefillIfNeeded, sendTapped); confirmed the two conditional clears at 5085 and 5158.
- Lost-clear enumeration: grepped App/HerdrApp.swift for `sendTapped(`, `onReturn`, `canSend`, `trimmingCharacters` over lines 3700-5060 and traced every `.submitText` payload producer — sendTapped() 5359, onReturn 4873-4879 with shouldChangeTextIn 3846-3848, usePrompt 5045-5048 — against Coordinator.textViewDidChange 3833-3837 to show the compared string is always the composer's own binding value.
- Composer plumbing: read App/HerdrApp.swift:3739-3945 (TerminalReplyField, Coordinator, ReplyTextView incl. canPerformAction/paste) and 4837-4926 (field wiring, MicButton, send button, canSend).
- Live Activity: read App/LiveActivityController.swift:100-160; grepped `staleDate`, `displayHeadline`, `headline`, `question`, `defaultAnswer`, `Nothing needs you`, `needsYouCount == 0` across App, HerdrWidgets, Shared, Sources and Tests.
- Widget packaging: read project.yml:280-340 and grepped `UIAppFonts`/`Geist`/`IBMPlex` to check the widget target's font registration against WidgetFont's PostScript names in HerdrWidgets/AgentLiveActivity.swift:337-343.

PR #249, opened 2026-09-13
Accept any pasted file in the terminal composer

jerryfane:
The terminal composer accepted only images, so "paste a photo" worked while "paste the file you just copied" did nothing at all.

Nothing in the delivery path was image-specific — staging, the Gram upload channel and the prompt reference are byte-agnostic — so the restriction lived entirely in the pasteboard type filter.

## What changed

- `PastedFile` (new, file-private) decides what counts as a pasted file: anything with concrete bytes or a package, minus text, rich text, links and contact cards, which stay ordinary pastes. Both the text view and the pane read this one rule, so they cannot disagree.
- The edit-menu check reads `UIPasteboard.types` — type metadata only, no content read, so no "Allow Paste?" prompt while UIKit builds the menu. The content read stays inside the user-initiated `paste(_:)`.
- `paste(_:)` still prefers text when the pasteboard carries both, and still falls through to UIKit on every bail-out path.
- `PromptPhoto` → `PromptAttachment` with an `isImage` flag that chooses only the chip glyph (`photo` vs `doc`) and the wording of the prompt reference (`[Photo attached via …]` vs `[File <name> attached via …]`). Same staging, same upload, same gram post, same 100 MiB cap.
- Accessibility identifiers follow: `terminal-attachment` and `terminal-attachment-loading`.

## Verification

Focused `TerminalControlTests` on iPhone at head `2be091c`: [run 34775654311](https://github.com/jerryfane/herdrup/actions/runs/34775654311), 14/14 passing.

- New `testCopiedFileOffersPasteAndSendsAttachment` copies a PDF through the harness, pastes it, and asserts the chip appears, sends without caption text, and clears after prompt delivery — the non-image branch end to end.
- `testCopiedPhotoOffersPasteAndSendsAttachment` still passes, so the image path is unchanged.
- Linux HerdrKit and vendored SwiftTerm suites pass.

Requested by owner feedback recorded in workflow note 150401 (photo paste confirmed working on TestFlight build 139; file paste expected). No merge or release authorization is claimed here.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d4f8c7429004f7-1`

**Decision:** `approved`
**Summary:** Reviewed PR #249 at exact head 2be091cb064fa5b375b79d36d829c5e982cb85f8 (worktree .git/HEAD read directly and matches; no prior verdict for #249 in the frozen prior-verdicts set, which covers 244/246/247/248 only). Static/read-only: bash, eval and hub are all approval-gated in this seat and were refused, so nothing was executed and no build was attempted; the iOS-only target cannot be built here regardless. VERDICT: the generalization is correct and I found no P0-P2 defect. Type filter (HerdrApp.swift:3745-3762) is right for every case the brief names. isAttachment requires public.data OR com.apple.package conformance, then subtracts public.text, public.url, public.vcard, public.rtf and com.apple.flat-rtfd. public.rtf and public.vcard already conform to public.text/public.url-adjacent hierarchies so their explicit exclusions are belt-and-braces, but the com.apple.flat-rtfd exclusion is LOAD-BEARING: flat-RTFD conforms to public.data and public.composite-content, not public.text, and it is the flavor UIKit actually writes for rich text with attachments, so without line 3751 an attributed-text paste could have been read as bytes. public.file-url conforms to public.url, so a copied link or file-URL flavor cannot be picked. UTType(_:) returns nil for undeclared identifiers, so unknown app-private flavors are skipped rather than guessed. Plain text cannot be swallowed: PastedFile.isAttachment excludes public.text-conforming types AND paste(_:) is additionally gated on !UIPasteboard.general.hasStrings (3910), so a pasteboard carrying both a string and a file always goes to UIKit. Fall-through is provably total: paste(_:) is one &&-chained if with a single early return, so no-file-type, text-present, no-matching-provider, nil handler and a handler returning false all reach super.paste(sender) (3904-3919); pasteReplyAttachment returns false on every bail-out (4966, 4971, 4979) and sets a visible actionNote for the non-intent-pane case rather than no-oping. PASTEBOARD PRIVAC
[truncated]
**Severity:** `P3`

**Findings**
- **Text-preference branch (the newest and riskiest guard) has no receipt** (P3)
  - detail: Every paste fixture is single-flavor: newline-pasteboard and reply-multiline-pasteboard are string-only, photo-pasteboard is image-only, file-pasteboard is PDF-only. Nothing exercises a pasteboard carrying BOTH a string and a file, which is exactly the new `!UIPasteboard.general.hasStrings` gate at HerdrApp.swift:3910 and the failure mode with the worst blast radius (a swallowed plain-text paste). A future narrowing of that guard would keep all 14 focused tests green. Cheap fix if wanted: one harness case using UIPasteboard.general.setItems([[UTType.utf8PlainText.identifier: "copied-text", UTType.pdf.identifier: pdfData]]) asserting the text reaches the field AND terminal-attachment never appears - which pins the premise (both flavors present) alongside the conclusion (text won), not just the outcome.
  - file: HerdrUITests/TerminalControlTests.swift:300-320, App/TerminalInteractionHarness.swift:475-485
  - id: f1
- **The isImage-dependent prompt reference is unobservable to any test** (P3)
  - detail: The only user-visible behavioral difference the rename introduces for non-images is the reference wording and download path (noun `File <name>` vs `Photo`, stem herdr-file- vs herdr-photo-). TerminalInteractionDriver.roundTrip forwards agent.prompt and every gram.* method to a fresh MockTransport (TerminalInteractionHarness.swift:105), whose canned Self.agentPrompted / Self.gramPosted replies discard the request text, so the delivered prompt string is not recorded anywhere the UI test can query. The new receipt therefore proves delivery but cannot distinguish the image wording from the file wording; the two tests would still pass if `noun` were hardcoded to "Photo". Pinning it needs new harness plumbing (record the last agent.prompt text into the driver snapshot), so this is a gap to be aware of, not a defect.
  - file: App/HerdrApp.swift:5181-5188, App/HerdrApp.swift:7392
  - id: f2
- **com.apple.package types pass the filter but can never stage** (P3)
  - detail: isAttachment admits com.apple.package conformers (the comment at 3746 intends .rtfd bundles), but stageCopy copies the directory and then requires `.fileSizeKey` > 0, which is nil for a directory, so the guard at GramStaging.swift:49-54 deletes the item and returns nil. Net effect for a package paste: pasteReplyAttachment already returned true, so paste(_:) consumed the gesture and the user gets `Couldn't add that attachment.` with nothing inserted, instead of falling through to UIKit. Practically unreachable - UIKit writes com.apple.flat-rtfd (excluded) rather than com.apple.rtfd, and any realistic rich-text pasteboard also carries a string flavor, which the hasStrings gate catches first - so the `|| type.conforms(to: .package)` disjunct currently buys an error toast rather than a capability. Harmless, but it is dead-ish surface that reads as supported.
  - file: App/HerdrApp.swift:3749, Sources/HerdrKit/GramStaging.swift:36-60
  - id: f3
- **Observations, no action requested** (P3)
  - detail: (a) canPerformAction ignores isEditable, so Paste is offered on the composer while replyDictating holds it non-editable, and pasteReplyAttachment guards !sending/!loadingReplyAttachment but not !replyDictating - an attachment can be staged mid-dictation. Benign: canSend already excludes replyDictating, so it cannot be sent until dictation ends. (b) Both attachment tests conclude with an absence-shaped wait (chip `exists == false`), which an app crash would also satisfy; this is the pre-existing shape inherited from the photo test, not something this PR introduced. (c) No UI test can detect a pasteboard-privacy regression here, because the harness writes the pasteboard from inside the app under test and same-app reads never raise the Allow Paste alert - the privacy property at 3760/3910 is review-verified only.
  - file: App/HerdrApp.swift:3897-3902, HerdrUITests/TerminalControlTests.swift:316-319
  - id: f4

**Tests Run**
- Head identity: read .git/HEAD verbatim = 2be091cb064fa5b375b79d36d829c5e982cb85f8, identical to the dispatched head; no reliance on the job record, which is derived from the worktree and therefore not an independent witness.
- Prior verdicts: read evidence/prior-verdicts.json (as_of 2026-09-13T19:42:01Z) and grepped for pull_request 249 and head_sha 2be091cb - no matches; existing entries cover 244 (2), 246 (5), 247 (2+), 248 (3).
- Static read of App/HerdrApp.swift:3689-3963 (PastedFile, TerminalReplyField, ReplyTextView.canPerformAction/paste), 3984-4013 (PromptAttachment + state), 4152-4158 (hasReplyContent/canSend), 4780-4960 (chip strip, replyBar wiring), 4960-5260 (pasteReplyAttachment, finishReplyAttachmentPaste, removeReplyAttachment, send interception, sendPromptWithAttachment, postReplyAttachment), 7359-7395 (MockTransport routing).
- Static read of App/TerminalInteractionHarness.swift:1-10 (#if DEBUG && canImport(UIKit) gate), 11-150 (driver roundTrip: only agent.list/pane.set_pty_size/pane.send_text/agent.read intercepted; everything else falls to a fresh MockTransport), 180-300 (snapshot fields - no prompt text), 430-540 (file-pasteboard PDF fixture and the 19-entry command grid).
- Static read of Sources/HerdrKit/GramStaging.swift:1-105 (stageCopy/stageData size gates, safeFileName) and App/GramView.swift:1586-1609 (Staging.maxFileBytes = 100 MiB, Staging.copy delegating to stageCopy).
- Static read of HerdrUITests/TerminalControlTests.swift:200-388 (new testCopiedFileOffersPasteAndSendsAttachment plus the replyAttachmentChip and pasteIntoReply helpers shared with the photo test).
- Rename-completeness grep (repo-wide, all roots): PromptPhoto|onPasteImage|onPastePhoto|replyPhoto|loadingReplyPhoto|pasteReplyPhoto|finishReplyPhotoPaste|removeReplyPhoto|sendPromptWithPhoto|postReplyPhoto -> 0 matches.
- Callsite grep (repo-wide): terminal-attachment|photo-pasteboard|file-pasteboard|onPasteFile|pasteIntoReply -> all references accounted for in HerdrApp.swift, TerminalInteractionHarness.swift and TerminalControlTests.swift only.
- Pasteboard-read audit grep (repo-wide): rtfd|vCard|hasStrings|UIPasteboard.general.(types|itemProviders) -> exactly three read sites, all in HerdrApp.swift:3760/3910/3911, confirming no other surface reads item content.
- NOT RUN: no build, no XCTest, no git plumbing. bash, eval and hub each returned 'requires approval but no interactive UI available' in this seat. The reported green focused TerminalControlTests iPhone run 34775654311 (14/14) is taken as dispatch input, not as evidence I produced.

PR #250, opened 2026-09-13
Show upload progress and stop the composer clipping the typed line

jerryfane:
Two reader reports from TestFlight build 140.

## 1. Upload progress (new)

A large photo or file uploaded with no feedback beyond a dimmed chip, so a slow link was indistinguishable from a hang. `gramUploadFile` already reports `(bytesSent, totalBytes)` on the main actor — Gram's composer uses it — and the terminal composer ignored it.

The attachment chip now shows a determinate bar plus "Uploading 42% · 8.4 MB / 20.0 MB" (percentage only under 1 MB, where the byte pair is noise). Sizes render through HerdrKit's single `GramFile.displaySize` formatter rather than a second copy. While the upload runs, the chip's remove button is replaced by the progress label, so the file cannot be pulled out from under an in-flight upload.

## 2. "Typing past two lines looks like nothing is being written"

Reported behaviour: from the third line on, typed text seems not to appear, and it comes back on the fourth line.

**I could not reproduce this in CI**, on iPhone or iPad, including with a receipt that measured the composer's own caret rect, content size and scroll offset. That receipt passed both with and without the fix, so it proved nothing and was deleted rather than kept as a green test that cannot fail.

What this change does address is the mechanism that best fits the report, found by reading the sizing path:

- `fittingHeight` measured through `UITextView.sizeThatFits`, whose answer depends on `isScrollEnabled` and on how much TextKit has lazily laid out. If it under-reports by one line, SwiftUI hands the view a two-line frame with three lines of text in it, and with scrolling off the third line — the one being typed — is clipped. Height is now measured from the string with `boundingRect`, which cannot lag the text or change with scroll state.
- Scrolling was toggled off below the three-line cap, which meant the caret could not be scrolled into view in exactly the state where it had left the view. Scrolling is now always on; only bouncing tracks overflow, and a composer whose content fits is pinned back to offset zero so a stale offset cannot blank the field.

That is a reasoned fix for an unreproduced report, not a verified one — stated plainly so the next reader does not over-trust it. If it persists on the next build, a screen recording plus the device's Settings › Display & Brightness › Text Size would narrow it further.

## Verification

Focused `TerminalControlTests` on iPhone at head `dbb3cef`: [run 34784140141](https://github.com/jerryfane/herdrup/actions/runs/34784140141), 14/14 passing, including the existing three-line growth, overflow scrolling and multiline-paste receipts.

Not verified: the progress bar's appearance (a CI upload completes instantly through the mock, and there is no device here to watch it) and the clipping report itself.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d517d0f0b40664-1`

**Decision:** `approved`
**Summary:** Reviewed PR #250 at exact head dbb3cefc68b2c0d2c8718c2d8add2e416e3c54b2 (git rev-parse HEAD matches; net diff f8ff661..dbb3cef is App/HerdrApp.swift only, 92+/36-, plus the probe deletion in dbb3cef touching TerminalInteractionHarness.swift and TerminalControlTests.swift). No prior verdict exists for #250 in the frozen prior-verdicts set (as_of 2026-09-14T05:10:42Z; it covers 244/247/248/249). No toolchain in this seat: `which swift swiftc xcodebuild` is empty, so nothing was compiled or executed and the reported green run 34784140141 is taken as the requester's evidence, not mine.

(1) UPLOAD PROGRESS — correct, and it mirrors an already-shipped pattern. `gramUploadFile(fileURL:onProgress:)` exists at Sources/HerdrKit/HerdrClient.swift:644 with `onProgress: (@MainActor @Sendable (Int, Int) -> Void)?` and every call site is `await onProgress?(...)` (lines 696, 716, 725), so the closure hops to the main actor before touching `replyUploadBytes`; the @State write is isolation-safe with no Sendable hole. `staged.size` is `Int` (GramStaging.swift:8) and `GramFile.displaySize(of: UInt64)` is public (Gram.swift:42), so `uploadLabel` type-checks and `UInt64(sent)` can never trap (byte counts are non-negative). CLEARING ON EVERY EXIT: `defer { sending = false; replyUploadBytes = nil }` is the first statement in the Task body, before the `do`, so it covers the success path, both `catch` arms, the two mid-flow `return`s (HerdrApp.swift:5202, 5233), and cancellation (`CancellationError` thrown out of the `await`). STRANDING: `replyUploadBytes` is non-nil only between 5204 and either 5208 or the defer, so the X button can be hidden only while an upload is genuinely in flight; and it was already `.disabled(sending)` for that whole window before this PR, so hiding it removes no capability the reader had. Same tuple/closure/clear shape as the reviewed GramView.swift:1237 path.

(2) COMPOSER SIZING — measurement is right at the cap, always-on scrolling does not strand an offset, but
[truncated]
**Severity:** `P3`

**Findings**
- **ceil() in fittingHeight fights the +0.5 epsilon at exactly the 3-line cap** (P3)
  - detail: maximumHeight = lineHeight*3 + 22 (3789) while fittingHeight returns ceil(3*lineHeight) + 22. The overflow test `fittingHeight > maximumHeight + 0.5` therefore reduces to `ceil(3L) - 3L > 0.5`, i.e. it fires spuriously whenever frac(3*lineHeight) lands in (0, 0.5). Geist-Regular@15 metrics are not in the repo so I cannot say which side it falls on; with the `.systemFont(ofSize: 15)` fallback (lineHeight 17.8955, frac(53.6865)=0.6865) it does not fire, i.e. the code is correct only by 0.19pt of luck. When it does fire, a field whose three lines exactly fit gets alwaysBounceVertical = true (rubber-band on non-overflowing content) and skips the pin-to-zero branch at the cap. Harmless beyond that — the SwiftUI height is still clamped to maximumHeight so no line is clipped, and scrollRectToVisible clamps to offset 0 because contentSize == bounds there. Fix is one character of intent: compare box.height unrounded, or widen the epsilon past 1.0. Same rounding also makes every height (including the empty-field minimum) up to 1pt taller than before.
  - file: App/HerdrApp.swift:3947,3956
  - id: f1
- **The fittingHeight comment states a mechanism the deleted probe contradicted** (P3)
  - detail: The doc comment asserts as established fact that "the box could still report a two-line height with three lines of text in it" and that the third line "was clipped away". The only instrumentation ever built to observe that — ReplyComposerProbe's `caretVisible` (caretBottom <= offsetY + bounds.height + 1 && caretTop >= offsetY - 1) plus testReplyComposerKeepsTheCaretVisibleAtEveryLineCount — would have gone false in exactly that state (caretBottom ~76 vs bounds 58, offset 0), and it reported PASS without the fix. That is a negative result: the hypothesized stale-sizeThatFits clipping does not occur in the simulator. Worse, the stated mechanism does not explain the reader's specific symptom — a line drawn ABOVE the box requires content taller than the frame AND a positive contentOffset, which under-reported height with offset pinned at 0 cannot produce (it would spill BELOW). The change is still the right change (see f3 for the mechanism that does fit), but the comment should say "could not be reproduced; this removes two candidate causes" rather than narrating one as diagnosed. Load-bearing because this file's comments are the project's primary verification artifact for iOS-only code that CI cannot compile.
  - file: App/HerdrApp.swift:3926-3934
  - id: f2
- **Better-supported root cause for 'drawn above the box' is the old early return, and the PR does not say so** (P3)
  - detail: The old refreshScrollMode reset contentOffset only after `guard isScrollEnabled != shouldScroll else { return }`, so the pin ran on the scroll-mode EDGE only. Setting isScrollEnabled = false does not reset a UIScrollView's contentOffset, and UITextView adjusts contentOffset itself on text change/first-responder, so a nonzero offset acquired while shouldScroll was already false was permanently uncorrectable — content shifted up inside a non-clipping-radius box is the one configuration that paints text outside the box's top edge, matching the reader's wording far better than clipping does. The new unconditional pin closes that. Worth recording in the commit body because it is the part of this fix that is actually targeted at the reported symptom; as written the PR credits the measurement change instead. Secondary note: `layer.cornerRadius = 20` (3782) is set without `masksToBounds`, so the rounded background never clips content; if a future height mismatch appears, an explicit `clipsToBounds` is the cheap containment.
  - file: App/HerdrApp.swift:3954-3961
  - id: f3
- **Deleting the caret receipt was right; keeping it would have been worse** (P3)
  - detail: Asked directly: no, deleting it was not wrong. A test that passes identically with and without the fix defends no contract, and leaving it in place would have advertised coverage for a bug it demonstrably cannot see — plus it kept a DEBUG-only mutable static (ReplyComposerProbe.geometry) and a receipt key alive in shipping code paths for nothing. Verified the removal is complete: no reference to ReplyComposerProbe survives anywhere in the tree, and the `wait` helper the test used lives in TerminalResizeTests.swift:38 so nothing is orphaned. The only thing that should have survived is the NEGATIVE RESULT, and the commit subject ("could not tell the bug apart") carries it only partially — see f2.
  - file: HerdrUITests/TerminalControlTests.swift (deleted in dbb3cef)
  - id: f4
- **A stat that returns 0 wedges the bar at 0% for the whole upload** (P3)
  - detail: The chip is seeded with the trustworthy `staged.size`, then the client's first callback overwrites it with its own stat (HerdrClient.swift:650, `?? 0`). If that stat fails after FileHandle open succeeded, total becomes 0, uploadLabel returns "Uploading 0%" forever (the >= 1MB branch is skipped, percent is hardcoded 0) and ProgressView sits at 0 while bytes actually move — the precise "reads as a hang" failure the feature exists to prevent, until the terminal `(offset, offset)` report at 725 snaps it to 100%. Narrow window, but `replyUploadBytes = (sent: sent, total: max(total, staged.size))` in the closure would close it for free.
  - file: App/HerdrApp.swift:5204-5207
  - id: f5
- **Programmatic text sets no longer scroll to the caret (behavior change, arguably better)** (P3)
  - detail: updateUIView's text-changed branch calls refreshScrollMode() but not requestCaretReveal(); the old code got an implicit reveal from the shouldScroll transition setting shouldRevealCaretAfterLayout. So inserting a saved prompt longer than three lines (reply = p.text at 5130) or seeding initialReply (4152) now leaves the field at offset 0 showing the TOP of the text instead of scrolled to the end. I consider that an improvement for wholesale inserts, and the first keystroke restores caret-following via textViewDidChange, so no action needed — recording it only so the delta is not mistaken for a regression later.
  - file: App/HerdrApp.swift:3806-3811
  - id: f6

**Tests Run**
- git rev-parse HEAD -> dbb3cefc68b2c0d2c8718c2d8add2e416e3c54b2 (matches the review head)
- git diff f8ff661..dbb3cef --stat and full diff read (App/HerdrApp.swift, 92+/36-)
- git show --stat dbb3cef and f2d1113 (confirmed the probe/receipt deletion scope: 3 files, 73 deletions)
- which swift swiftc xcodebuild -> empty; no build or XCTest executed, evidence is static_only
- grep -rn ReplyComposerProbe . -> no matches (probe removal is complete)
- read Sources/HerdrKit/HerdrClient.swift:589-737 (onProgress is @MainActor @Sendable, invoked only via await; total widening at 716/725)
- grep StagedAttachment / GramFile.displaySize (staged.size: Int; displaySize(of: UInt64) public)
- read App/HerdrApp.swift:3740-3983, 4250-4300, 4890-5006, 5136-5259 (composer, container VStack, replyBar, send path)
- arithmetic check of ceil(3*lineHeight)+22 vs lineHeight*3+22+0.5 for both Geist and the systemFont fallback
- cross-checked the upload-progress pattern against the existing App/GramView.swift:1237 call site

PR #252, opened 2026-09-14
Lowercase the widget name, add a gradient, and accept copied or dropped files

jerryfane:
Two reader requests against TestFlight build 142.

## 1. Widget wording and background

- The Lock Screen action read "Open Herdrup". The product is **herdrup**, lower case, so the button says "Open herdrup".
- The Lock Screen activity was a flat `activityBackgroundTint`. It now paints a two-stop gradient behind the content — `#1B1F3A` top-leading to `#13162A` bottom-trailing — eight points of lightness apart: enough to read as depth, not enough to compete with the status marks, which are the only saturated things on that surface. The flat tint stays as the base so a system fallback still matches.

## 2. Files from a Mac or an iPad

Command-V of a file copied in Finder or Files pasted the **path as text** and attached nothing, and there was no drop target at all.

- **Why paste failed:** `public.file-url` conforms to `public.url`, and the type rule excluded links so a pasted web address would not become an attachment. A file url names bytes; a web url names a page. It is now accepted explicitly, and it wins over an accompanying text representation — Finder puts the path on the pasteboard beside the file, and deferring to that text is what produced the visible bug. Everything else still defers to text.
- **A file url is a reference, not bytes.** Loading it as a file representation yields a temp file containing the path string, so the handler resolves the URL object and copies from the real file inside a security scope, which is what makes a document outside the app container readable at all. The chip carries the file's own name and MIME.
- **Drag and drop:** the whole reply bar is a drop target (`.onDrop(of: [.item])`), not just the 40-point text view, with a border while a drag hovers. Each dropped provider goes through the same staging path as a paste, so the 10-file cap, the type rule, the named-agent guard and the serial upload are shared rather than re-stated.

## Verification

Focused `TerminalControlTests` on iPhone at head `6253e12`: [run 34861556246](https://github.com/jerryfane/herdrup/actions/runs/34861556246), 16/16 green, including new `testACopiedFileURLAttachesTheFileRatherThanItsPath` — it seeds the pasteboard the way Finder does (file url plus the path as text), pastes, and asserts the chip is named `dropped-note.txt` while the composer text stays empty. That test fails on the old rule, which pasted the path.

Not verified: drag and drop itself (XCUITest cannot synthesise a cross-app drag), and the widget's rendered gradient — no test target can see a Live Activity, so both need eyes on a device.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d53afa9b63dc6d-1`

**Summary:** Head 6253e1263ea5a160e506717cb05baa8cb3905e3a confirmed local (git log = 6253e12 'Lowercase the widget name, add a gradient, and accept copied or dropped files'; 4 files, +125/-11). No prior verdict exists for PR #252 in GITMOOT_PRIOR_VERDICTS (frozen 2026-09-14T15:54:43Z; PRs 228/251 and earlier only), so this is a first-round review.

(1) WIDGET — clean. 'Open herdrup' is the only product-name string in HerdrWidgets (grep Herdrup|herdrup over HerdrWidgets returns exactly AgentLiveActivity.swift:207); remaining repo hits are provisioning-profile names and comments in project.yml/.github/workflows/testflight.yml, and both CFBundleDisplayName entries (app + widgets, project.yml:141/311) are already lowercase 'herdrup'. Gradient cannot wash out the status marks: I computed WCAG contrast for every mark/text colour against BOTH stops. Worst mark at the lightest corner 0x1B1F3A is died 0xE2584E at 4.40:1 (was 4.88 on flat ground); working 5.60 (6.21), waiting 7.66 (8.49), idle/textFaint 3.19 (3.54), text 14.15 (15.68). Stop-to-stop ratio is 1.109 — 8 points of lightness, matching the doc comment. Always-On / luminance-reduced is safe: both stops are near-black (relative luminance 0.0152 and 0.0088), so the unconditional .background adds no meaningful emission, and the only sub-4.5:1 foreground (textFaint lockDetail) is already suppressed in that state by `if !isLuminanceReduced` at line 198. activityBackgroundTint(WidgetPalette.ground) still matches the gradient's bottom-trailing stop exactly, so a system fallback or a system-inserted gutter cannot show a seam of a different hue. One pre-existing wart very slightly worsened: the outlined ActivityAction border (hairline 0x2E3358, used when isStale || isLuminanceReduced) goes 1.47:1 -> 1.33:1 against the lighter corner; the 15pt title still renders at 14:1, so the control stays legible.

(2) FILES — the paste fix is right; the drop path has one real defect. isAttachment's fileURL early-return before the .url exclusion is t
[truncated]
**Severity:** `P2`

**Findings**
- **Multi-item drop silently discards every file after the first, yet reports the drop as accepted** (P2)
  - detail: `providers.map { pasteReplyAttachment($0) }.contains(true)` calls the handler eagerly for all providers. The first call sets `loadingReplyAttachment = true` synchronously (line 5120 / 5142) before its async load; every subsequent provider then trips `guard !sending, !loadingReplyAttachment else { return false }` at line 5093 and returns false WITHOUT setting `actionNote` — that guard is the one bail-out with no user-facing message. `.contains(true)` is still true, so SwiftUI plays the accept animation for the whole drop session. Dragging three files out of Files or Finder therefore stages one and drops two with zero feedback. This is specific to drop: Command-V is inherently one item, so the serial design was sufficient for paste, but `providers` is an array and the strip already holds up to GramView.Staging.maxAttachments (10). Either queue the remaining providers and drain them as each staging completes, or take only the first and say so (e.g. actionNote 'One file at a time.') so the loss is not silent.
  - file: App/HerdrApp.swift
  - id: F1
  - line: 5067-5069
- **When the file-url handler declines, paste(_:) falls through to super.paste and inserts the file PATH — the exact outcome this PR set out to remove** (P3)
  - detail: The new condition admits file-url pasteboards, but the fallthrough is unchanged: if `onPasteFile?(provider)` returns false, super.paste inserts the accompanying utf8 text, i.e. '/Users/…/dropped-note.txt'. Before this change that outcome was unavoidable for all file copies; now it is a deterministic surprise for three declines that have nothing to do with 'this is not a file': cap reached (line 5094), a paste already loading (line 5093 — a quick double Command-V), and no named intent agent (line 5098). In those cases the user sees both an explanatory actionNote (except the loading case, see F1) and a pasted path. Suppressing the text paste when PastedFile.pasteboardHasFileURL() is true would keep the composer clean and leave the note as the only signal.
  - file: App/HerdrApp.swift
  - id: F2
  - line: 3934-3943
- **Bare-file-url branch copies a dragged FOLDER recursively into tmp before rejecting it** (P3)
  - detail: A folder drag vends public.folder + public.file-url. public.folder conforms to public.directory, not .data/.package, so the middle selector at line 5109 skips it and the bare-file-url selector at 5110 accepts it — a path that was unreachable before this PR, because the old filter required .data or .package. GramStaging.stageCopy then runs FileManager.copyItem on a directory, which succeeds recursively, and only afterwards fails the `resourceValues(forKeys: [.fileSizeKey]).fileSize` guard (nil for a directory) and removes the tree. Dragging a large directory means an unbounded recursive copy into the temp dir before the 'Couldn't add that attachment.' message — exactly what stageCopy's own doc comment ('an unstat-able pick must never reach an unbounded copy') warns against. A `source.resourceValues(forKeys: [.isDirectoryKey])` check (or a `!type.conforms(to: .directory)` exclusion) before staging closes it.
  - file: App/HerdrApp.swift
  - id: F3
  - line: 5119-5140
- **Non-materialised iCloud item fails with a generic message on the reference path** (P3)
  - detail: The bare-file-url branch copies with plain FileManager semantics — no NSFileCoordinator read and no startDownloadingUbiquitousItem — so an iCloud Drive file that is still a placeholder fails the copy and surfaces only 'Couldn't add that attachment.' Providers that vend concrete bytes avoid this because loadFileRepresentation materialises for you; a Finder/Files reference-only copy of an evicted file does not. Worth either a coordinated read or a more specific note, since this is a plausible iPad case and is not covered by the new test.
  - file: App/HerdrApp.swift
  - id: F4
  - line: 5122-5136
- **Doc comment contradicts the staging path it feeds** (P3)
  - detail: 'A file url wins over the file's own concrete type when both are advertised' is true only for provider LOOKUP; pasteReplyAttachment deliberately prefers the concrete type (line 5109 before 5110) precisely because a file url is a reference. As written the comment reads as a statement about which representation gets staged, which is the opposite of the implemented — and correct — precedence.
  - file: App/HerdrApp.swift
  - id: F5
  - line: 3767-3771

**Tests Run**
- git log --oneline -3 / git show --stat HEAD — confirmed review head 6253e1263ea5a160e506717cb05baa8cb3905e3a and the 4-file diff (App/HerdrApp.swift +90, App/TerminalInteractionHarness.swift +12, HerdrUITests/TerminalControlTests.swift +19, HerdrWidgets/AgentLiveActivity.swift +15)
- grep 'Herdrup|herdrup' over HerdrWidgets — single hit, AgentLiveActivity.swift:207 'Open herdrup'; repo-wide sweep over *.swift/*.plist/*.yml/*.pbxproj shows the remaining capitalised strings are provisioning-profile names and comments only; project.yml:141/311 CFBundleDisplayName already lowercase
- WCAG relative-luminance/contrast computation (python) for waiting E9A63C, working 5B9BE8, died E2584E, textFaint 666D91, textDim 99A0BC, text EEF0F7, hairline 2E3358 against both gradient stops 0x1B1F3A and 0x13162A; stop-to-stop ratio 1.109
- Read of PastedFile (HerdrApp.swift:3743-3780), the canPerformAction/paste overrides (3918-3944), pasteReplyAttachment/finishReplyAttachmentPaste/removeReplyAttachment (5084-5220), replyBar drop chain (5040-5081), LockScreenView + WidgetPalette (AgentLiveActivity.swift:1-383), and GramStaging.stageCopy/stageData/safeFileName (Sources/HerdrKit/GramStaging.swift:1-125) to trace the cap, named-agent guard and serialisation
- grep for other paste/drop entry points (onDrop|dropDestination|override func paste|itemProviders across App/) — the reply bar is the only drop target, GramView has no file-paste path, LiveTerminalView.paste only disarms Ctrl, so no second convention diverges from PastedFile
- NOT RUN, and not claimed: no Swift/Xcode toolchain on this Linux host (which swift swiftc xcodebuild all empty), so the cited TerminalControlTests iPhone run 34861556246 (16/16, incl. testACopiedFileURLAttachesTheFileRatherThanItsPath) was reviewed as source, not executed. The test's pasteboard is same-process (TerminalInteractionHarness sets UIPasteboard.general.items directly), so it does not exercise the cross-process sandbox extension, the security scope, or any drag; the PR's 'drag gesture and rendered gradient unverified' caveat is accurate and F3/F4 live in that unverified region.

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d53de326126dc7-1`

**Decision:** `approved`
**Summary:** Head e2c10c3aadb9fd7e9f103900255f2d8a59f29041 confirmed local (git rev-parse = e2c10c3 'Stage every dropped file, refuse folders and placeholders, never paste the path'; delta vs 6253e12 = App/HerdrApp.swift only, +74/-7; whole PR vs merge-base ae94dcf = 4 files, +191/-10; tree clean). Prior verdict read from GITMOOT_PRIOR_VERDICTS (frozen 2026-09-14T16:48:44Z): herdr-ios-review-opus, PR 252, head 6253e126, changes_requested, P2, 5 findings, static_only. All five are addressed.

(1) MULTI-ITEM DROP — fixed, and it cannot strand the serialising flag. acceptDroppedFiles (5105-5122) never writes loadingReplyAttachment; the only writers are pasteReplyAttachment:5162 (file-url branch) and :5209 (typed branch), and every completion path reaches finishReplyAttachmentPaste, which clears it at 5255: loadObject's completion always fires, and loadFileRepresentation -> nil falls into loadDataRepresentation whose completion also always fires. So the 5s `break` at 5115 abandons the remaining providers but leaves the composer usable. I checked the unusual loop form too: `for _ in 0..<100 where loadingReplyAttachment` is a Swift for-in FILTER, not an early exit, so it does not break out when the flag clears — it simply stops sleeping and spins the rest of the range as no-ops. Net semantics match the doc (wait while loading, cap 5s per file).

(2) DECLINED FILE-URL PASTE — fixed. paste(_:) 3935-3947 consumes a file-url pasteboard unconditionally, so a cap hit / in-flight load / agent-less pane can no longer drop the path into the composer. It cannot swallow a genuine text paste: pasteboardHasFileURL() is UIPasteboard.contains(pasteboardTypes:) (3779), metadata-only, and public.url does not conform to public.file-url, so the Safari image+link case, web links and plain text still reach the `!hasStrings` branch (3948) and fall through to super.paste. The Finder/Files trap (file url + path text on the same item) is exactly what the new harness scenario 'file-url-pasteboard' builds, and 
[truncated]
**Severity:** `P3`

**Findings**
- **Two declines still set no note, and both new consume-everything paths widen their reach** (P3)
  - detail: pasteReplyAttachment returns false WITHOUT setting actionNote in two guards: 5135 (`sending` or `loadingReplyAttachment`) and 5150-5155 (no usable type). Two consequences of this round's change: (a) Command-V while a previous attachment is still staging, or during a send, now returns at 3946 having consumed the paste — nothing stages, no path text, no note, so the gesture is a silent no-op; the new comment at 3942-3945 explicitly claims 'a load already in flight ... says so in the note', which is false, and 5099 calls this same bail-out 'the one bail-out with no note'. (b) Dragging selected TEXT onto the reply bar: .onDrop(of: [.item]) now returns true unconditionally, SwiftUI plays the accept animation, the type guard at 5150 declines silently and nothing appears — at 6253e12 the `providers.map{...}.contains(true)` return produced a visible snap-back rejection. Suggested fix: set actionNote in both guards (e.g. 'Still adding the last file — try again in a moment.' and 'That can't be attached.').
  - files:
    - App/HerdrApp.swift:3942-3947
    - App/HerdrApp.swift:5081
    - App/HerdrApp.swift:5135
    - App/HerdrApp.swift:5150-5155
- **Multi-file PASTE still stages only the first item, now with no fallback** (P3)
  - detail: paste(_:) picks `UIPasteboard.general.itemProviders.first(where:)` (3937), so selecting three files in Files and pressing Command-V attaches one, silently discards the other two, and — because the paste is now consumed unconditionally — leaves no text fallback either. This is the same defect acceptDroppedFiles was added to fix for drop, so the two halves of the feature are now asymmetric. A loop equivalent to acceptDroppedFiles over the matching providers would close it.
  - files:
    - App/HerdrApp.swift:3936-3941
    - App/HerdrApp.swift:5105-5122
- **ubiquitousItemDownloadingStatus != .current also refuses a readable stale copy** (P3)
  - detail: 5183-5188 refuses whenever the status is not .current, which includes .downloaded — a local copy that exists and is readable but is out of date. Such a file would be told 'isn't downloaded yet' and refused even though stageCopy would succeed. The precise predicate is `== .notDownloaded`. Low likelihood in practice (.downloaded is deprecated and modern OSes report .current/.notDownloaded), which is why this is P3 rather than higher.
  - files:
    - App/HerdrApp.swift:5183-5188
- **'Folders can't be attached' also fires for document packages, and the 5s timeout abandons files silently** (P3)
  - detail: Two smaller observations on the new refusals. (a) isDirectoryKey is true for PACKAGES too (.pages, .key, .rtfd), so a bare-file-url copy of a package document now reports 'Folders can't be attached — pick the files inside.' That contradicts PastedFile.isAttachment deliberately admitting .package (3756), though it is not a functional regression: stageCopy's fileSizeKey guard returns nil for any directory, so packages never staged — they used to say 'Couldn't add that attachment.' Wording only. (b) When a single staging exceeds the 5s budget, the `break` at 5115 drops every remaining dropped file with no note — a bounded version of the silent loss this round fixed. A note ('Only the first N files were added.') would close it.
  - files:
    - App/HerdrApp.swift:3754-3759
    - App/HerdrApp.swift:5109-5118
    - Sources/HerdrKit/GramStaging.swift:47-54

**Tests Run**
- git rev-parse HEAD / git log --oneline -5 — head confirmed e2c10c3aadb9fd7e9f103900255f2d8a59f29041
- git diff --numstat 6253e12 e2c10c3 — App/HerdrApp.swift +74/-7, single file
- git diff --stat ae94dcf e2c10c3 — 4 files, +191/-10; git merge-base confirms ae94dcf
- git status --porcelain — working tree clean, review of committed head only
- read every changed hunk: HerdrApp.swift 3740-3781, 3919-3957, 5074-5122, 5124-5273; TerminalInteractionHarness file-url-pasteboard scenario; TerminalControlTests new case; AgentLiveActivity gradient
- grep -n over HerdrApp.swift for loadingReplyAttachment / actionNote / finishReplyAttachmentPaste — traced all writers (5162, 5209) and the single clearer (5255); no strand path
- read Sources/HerdrKit/GramStaging.swift:36-60 — confirmed copyItem recursion and late fileSizeKey guard behind the new directory refusal
- read App/HerdrApp.swift:4324-4330 — actionNote is rendered (terminal-action-note), so the new refusal strings do surface
- which swift swiftc xcodebuild -> not found; gh -> GH_TOKEN unset: could NOT run TerminalControlTests, HerdrKit's 516 tests, or fetch CI run 34866084112. Verdict is static_only.

PR #253, opened 2026-09-14
Attach the copied document, not its Finder icon

jerryfane:
Reported from TestFlight build 143: copying a PDF in Finder and pressing Command-V attached `photo-31ec1fb2.icns`, 288 KB, a Mac OS X icon (`ic12`) — the document's Finder ICON, not the document.

## Cause

A Finder copy advertises three things: the file url, the path as text, and the document's icon as an image (`com.apple.icns`, which conforms to `public.image`). `pasteReplyAttachment` picked the image first — a rule written for screenshots — so the icon won, and the image branch named it `photo-<hash>.icns`.

## Change

`App/HerdrApp.swift`: the file url now wins whenever the item carries one. An image sitting beside a file url is always derived from that file — an icon, a thumbnail, or, when the file is itself an image, the same bytes under a worse name. With no file url the item is in-memory (a screenshot, an image copied out of Safari), and an image still wins over any other concrete type.

## Receipt

`testACopiedDocumentBeatsItsFinderIcon` seeds the pasteboard exactly as Finder does — file url + path text + PNG icon — and asserts the chip is `quarterly-report.pdf` with no `.icns` staged. It fails on the old precedence.

Focused TerminalControlTests iPhone run 34889572347: both file receipts green; the run's one failure was `testAppDeactivationDisarmsBeforeNextDirectKey` stalling the simulator main thread for 30s, which passed 17/17 on the same binary's second run.

Not verified: a real Finder copy on hardware — the fixture reproduces the pasteboard shape, not macOS itself.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d54a7b27bf507c-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/herdrup PR #253 at exact head 58599686d5edcfce471c681e2695bdfbdafcb0ac (git rev-parse HEAD matches; base 3504d56; diff = 3 files, +54/-7; git diff --check clean). No prior verdict for #253 in GITMOOT_PRIOR_VERDICTS (frozen 2026-09-14T20:38:47Z; it covers 226/228/229/244/246/247/248/249/250/251/252), so this is round one. EVIDENCE static_only and NOT by choice: `which swift swiftc xcodebuild` is empty and `gh auth status` reports no login, so CI run 34889572347 could neither be re-run nor fetched and the cited green receipts are the requester's evidence, not mine. What I DID execute: git rev-parse/log/diff/-S at this head, git diff --check, and grep/read of the whole staging path (HerdrApp.swift pasteReplyAttachment 5134-5259, PastedFile 3745-3781, the paste gate 3919-3957, GramView.Staging 1802-1837, GramStaging.stageCopy/stageData/safeFileName) plus the harness wiring.

THE FIX IS CORRECT AND THE PRECEDENCE IS THE RIGHT ONE. Q1 (in-memory / mixed regression): none found. public.file-url conforms to public.url, not the reverse, so Safari's Copy Image (public.url + image bytes) misses clause 1 and the image still wins; the same conformance direction governs the gate (pasteboardHasFileURL -> contains(pasteboardTypes:)), so a web url does not divert the paste either. photo-pasteboard (UIPasteboard.general.image) and file-pasteboard (PDF bytes) register no file url and take clauses 2/3 unchanged. No test in the suite asserted an image beating a coexisting file url, so nothing was re-pinned to fit the new order, and testAPastedPhotoStagesAnAttachment plus the two-attachment receipt both key on a `photo-` chip name that only the image branch can produce, which keeps that branch live under test. Dropping `&& !conforms(to: .fileURL)` from clause 3 is a true no-op because clause 1 catches every fileURL-conforming type first.

Q3 (PR #252 guards on the widened path): all still cover, and now cover strictly more. The directory refusal (5183), the iCloud-plac
[truncated]
**Severity:** `P3`

**Findings**
- **The .icns assertion in the new receipt is vacuous: the fixture cannot produce an .icns name** (P3)
  - detail: finder-document-pasteboard registers the icon under UTType.png.identifier (TerminalInteractionHarness.swift:517), so under the PRE-FIX ordering the image branch names the chip `photo-<8 hex>.png` (HerdrApp.swift:5220-5229 -- ext comes from type.preferredFilenameExtension, i.e. "png", and provider.suggestedName is nil for a UIPasteboard.items-built provider). `icons.count == 0` therefore holds in BOTH the buggy and the fixed world; only the first assertion (quarterly-report.pdf exists) discriminates, and it does discriminate, which is why the receipt still earns its place. But the assertion that names the actual field symptom is dead weight. Two ways to make it live: register the icon bytes under UTType("com.apple.icns") rather than .png -- the composer only reads the UTType to derive stem/ext/mime and loadDataRepresentation returns whatever bytes were registered, so no real icns encoder is needed and the fixture becomes faithful to build 143 -- or assert `app.staticTexts.matching(label BEGINSWITH "photo-").count == 0`, which fails against the pre-fix code exactly as written today.
  - file: HerdrUITests/TerminalControlTests.swift:393-395
  - id: f1
- **The precedence flip is broader than images: any item vending BOTH a concrete data type and a file url now takes the URL path, not loadFileRepresentation** (P3)
  - detail: The old clause order was image, then `isAttachment && !fileURL`, then fileURL -- so a provider registering e.g. com.adobe.pdf alongside public.file-url went through loadFileRepresentation (5232), whose coordinated read materialises a non-resident file-provider item. At this head it takes the fileURL branch instead, where the #252 placeholder guard (5190-5196) refuses a not-yet-downloaded iCloud item with "isn't downloaded yet -- opening it in Files once will fetch it." So an iCloud Drive file copied or dropped from Files that previously downloaded-then-attached now needs a second attempt. This is self-healing (startDownloadingUbiquitousItem is kicked off at 5192 and the retry succeeds) and is the guard's intended behaviour, and an explicit note beats an unexplained multi-second stall -- so not blocking. It is called out because the PR's own comment justifies only the image case ("any image beside a file url is derived from that file"), while the change also reverses concrete-bytes-vs-file-url precedence for non-image items. Note isUbiquitousItem is iCloud-only, so third-party file providers (Dropbox, Drive) skip the guard and materialise on open as before.
  - file: App/HerdrApp.swift:5157-5168
  - id: f2
- **No fallback from the file-url branch to the bytes branch when the URL is unusable** (P3)
  - detail: Once clause 1 matches, the branch commits: if loadObject(ofClass: URL.self) yields nil, or GramView.Staging.copy fails (unreadable, security scope not granted, copyItem throws), finishReplyAttachmentPaste(nil) reports "Couldn't add that attachment." even when the same provider still carries usable image or data bytes that the pre-#253 ordering would have staged. Bounded damage: a visible note, never a wrong file, and loadingReplyAttachment is cleared on every path (5262) so the composer is not wedged. Exposure is also limited because file-promise providers register the concrete type with a file representation rather than public.file-url, so promised items largely never reach clause 1. A `?? loadFileRepresentation/loadDataRepresentation` retry on failure would be strictly better. Related one-line hardening, pre-existing from #252: the branch never checks `source.isFileURL`, and NSURL's readable identifier is public.url (which public.file-url conforms to), so a provider registering a web url ahead of its file url could hand loadObject an https URL that then reaches FileManager.copyItem -- which throws, so again only a note.
  - file: App/HerdrApp.swift:5168-5214
  - id: f3
- **The file-url branch skips the .fileSizeKey pre-check that stageCopy's own contract requires of callers** (P3)
  - detail: GramStaging.stageCopy's doc comment (Sources/HerdrKit/GramStaging.swift:33-35) states "The caller stats the source first (an unstat-able pick must never reach an unbounded copy)", and GramView.attach honours it (GramView.swift:1570-1572, size > 0 && size <= maxFileBytes before staging). The composer's file-url branch requests only .isDirectoryKey/.isUbiquitousItemKey/.ubiquitousItemDownloadingStatusKey, so a multi-gigabyte Finder copy is fully copied into tmp and only then rejected by stageCopy's post-copy size check, which can churn or fill the device's temp space for an attachment that was never sendable. Introduced by #252, not by this PR -- but this PR routes strictly more items through that branch, so exposure grows. Adding .fileSizeKey to the existing `keys` set and refusing over maxFileBytes before the copy costs nothing, since resourceValues is already being called.
  - file: App/HerdrApp.swift:5176-5199
  - id: f4
- **A multi-file Finder copy still stages only the first item, silently** (P3)
  - detail: The paste gate resolves `UIPasteboard.general.itemProviders.first(where:)`, so selecting three files in Finder and pressing Command-V attaches one and sets no note, while the drop path correctly loops every provider (acceptDroppedFiles, 5105-5122). Pre-existing from #252 and outside this PR's stated scope; recorded so it is not mistaken for new behaviour, and because the drop path already proves the serialising pattern that would fix it.
  - file: App/HerdrApp.swift:3937-3941
  - id: f5

**Tests Run**
- git rev-parse HEAD -> 58599686d5edcfce471c681e2695bdfbdafcb0ac (matches the exact head under review)
- git log --oneline -3 -> 5859968 'Attach the copied document, not its Finder icon' on top of base 3504d56 (merge of #252)
- git diff 3504d56..58599686 --stat -> 3 files, +54/-7 (HerdrApp.swift, TerminalInteractionHarness.swift, TerminalControlTests.swift); full diff read
- git diff 3504d56..58599686 --check -> exit 0, no whitespace errors
- git log -S 'a screenshot also' -- App/HerdrApp.swift -> comment originated in 2be091c (PR #249) and is removed by this head; no fixture ever backed the 'a screenshot also vends a file URL' claim (photo-pasteboard uses UIPasteboard.general.image, image types only)
- grep of HerdrUITests for pasteboard receipts -> no existing test asserts an image winning over a coexisting file url, so the new precedence re-pins nothing
- read of HerdrApp.swift:5134-5259 (pasteReplyAttachment), 3745-3781 (PastedFile), 3919-3957 (paste gate), GramView.swift:1802-1837 (Staging), GramStaging.swift:1-106 -> #252 directory/iCloud/safeFileName/maxBytes guards all sit inside the fileURL branch; count cap is upstream and branch-independent
- fixture wiring check -> 'finder-document-pasteboard' present in TerminalInteractionControls.commands (571) for command() identifier lookup; harness switch has default: break; command count 28 -> 29 both ceil to 5 rows in the 6-column grid, so no harness layout shift
- which swift swiftc xcodebuild -> empty; gh auth status -> not logged in. NOTHING was compiled or executed; CI run 34889572347 could not be fetched or re-run and is taken as the requester's evidence, not mine

Issue #92, opened 2026-08-14
Fix blank HTML file preview (proper in-app HTML viewer)


jerryfane:
## Problem
Received `.html` file attachments open to a **blank white screen** in the Gram preview. Markdown files render fine; raw HTML does not.

## Root cause (confirmed)
`Sources/HerdrKit/HtmlSandbox.swift:48-58` wraps the received HTML in an `<iframe sandbox srcdoc="…">` inside a shell whose CSP (`:50`) is:

```
default-src 'none'; style-src 'unsafe-inline'; img-src data:;
```

There is **no `frame-src`/`child-src`** directive. Per CSP fallback, `frame-src` → `child-src` → `default-src` = `'none'`, so WebKit **refuses to load the `srcdoc` frame**. Only the outer shell paints, and its `body{background:#fff}` (`:53`) is the blank white screen. The inner document never renders, regardless of its content.

**Why Markdown works and raw HTML doesn't:** both paths share `openFile` → `gramGetFile` → QuickLook (`App/GramView.swift:918/924/199`) and write a correct `.html` temp file (extension/MIME are fine). The only difference: the Markdown path (`Sources/HerdrKit/Markdown.swift:295-332`) inlines the converted body **directly** in the document (no iframe) under the same CSP, so `style-src 'unsafe-inline'` lets it render. The raw HTML path (`GramView.swift:946` → `HtmlSandbox.wrap`) puts the content in a `srcdoc` iframe, which the CSP blocks.

**Why it escaped tests:** `Tests/HerdrKitTests/HtmlSandboxTests.swift` only asserts on the wrapper **string** (CSP substrings, `<iframe sandbox` present, no `allow-scripts`). Nothing ever renders the wrapper in WebKit, so the `frame-src`-blocks-`srcdoc` interaction was never exercised. (The file header even says "rendering behavior is checked on device/buildbox" — which wasn't done.)

## Proposed solution
**Option A (minimal):** allow the frame in the CSP at `HtmlSandbox.swift:50` (add `child-src 'self';`), keeping scripts off (the `sandbox` attribute has no `allow-scripts`) and network off (`default-src 'none'`). Caveat: `srcdoc` frames are `about:srcdoc` and CSP matching for them is subtle — this **must** be confirmed by an actual render, not a string assertion.

**Option B (recommended — a proper viewer):** replace the QuickLook-of-a-sandbox-wrapped-file + iframe approach with a dedicated SwiftUI `WKWebView` HTML viewer that:
[...]

PR #254, opened 2026-09-14
A JavaScript switch for HTML previews, and a glass island and lock screen

jerryfane:
Three owner requests against TestFlight build 143.

## 1. A JavaScript switch for HTML previews, off by default

Settings gains an **HTML PREVIEWS** section with one switch, **Run JavaScript**, and a line under it that states which rendering is in force. Off unless the reader turns it on.

- `WebViewPolicy.javaScriptDefaultsKey` / `javaScriptEnabled(in:)` hold the key and the default in HerdrKit, so the app's `@AppStorage` and the policy cannot drift. Absent means off.
- `HtmlWebView` takes `allowsJavaScript` and sets `allowsContentJavaScript` from it. The other two protections are NOT settings: the network content-rule block and the frozen navigation apply either way, so an enabled script can compute and repaint but cannot fetch, navigate or beacon.
- The flag lives on the `WKWebViewConfiguration` a `WKWebView` copies at init, so a change applies to the next preview opened. The row says so.

## 2. The expanded Dynamic Island

Reported as "overall a mess". The cause is structural: `.leading` and `.trailing` squeeze whatever sits in `.center`, and the old layout put a 16 pt headline there between a 26 pt counter on one side and three lines of totals on the other. Rebuilt so each region has one job:

| Region | Before | Now |
|---|---|---|
| leading | mark + big count + "need you" | status mark + count, one line |
| trailing | host + working + agents, three lines | the machine's name, one line |
| center | headline + question + timer, squeezed | *nothing* |
| bottom | the Open button, only when needed | the whole card: headline, question, timer + fleet counts, and the action |

Nothing is repeated across regions any more; the counts have exactly one home.

## 3. Glass

- **Lock Screen**: `activityBackgroundTint(nil)`, so the system's own translucent surface stays behind the view and the wallpaper shows through. Over it: `.ultraThinMaterial`, a translucent navy wash that keeps the herdrup hue, and a lit hairline edge. Under Always-On (`isLuminanceReduced`) it falls back to the opaque backdrop — a blurred wallpaper at 1 Hz is unreadable and wasteful.
- **Expanded island**: the bottom card sits on the same glass surface, so it reads as a pane rather than text on black.

## Receipts

- `WebViewPolicyTests.testJavaScriptIsOffUntilTheSettingIsTurnedOn` (Linux): an unset key renders without script; the setting is honoured both ways.
- `SettingsTests.testJavaScriptPreviewSwitchFlipsAndExplainsItself` (iPhone): the switch exists and its explanation tracks it in both directions, and the test leaves it off.

Not verified: anything the Live Activity renders. No test target can see a Dynamic Island or a Lock Screen widget, so the layout and the glass need eyes on a device.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d54cc5b8e19db8-1`

**Summary:** Reviewed eff9c9d8ede369ccf89a8cdb568bcc056260011a (base 3504d56, 7 files, +280/-86) by full read of every changed source. The JavaScript switch is mechanically sound: the default is genuinely OFF on a fresh install AND on upgrade from build 143 (both @AppStorage declarations carry `= false`, and there is no `register(defaults:)`, `UserDefaults(suiteName:)` or `defaultAppStorage` anywhere in App/, Sources/ or HerdrWidgets/, so an absent key resolves to false). Enabling script does NOT weaken the other two protections in any path I could trace: the content rule is installed before `loadHTMLString` inside the compile callback (no race), it fails closed on both a nil store and a nil compiled list, the coordinator cancels every navigation after the first (covering script-initiated `location`, subframe loads and meta-refresh), no `uiDelegate` is set so `window.open` cannot create a webview, and the 2-arg-only policy method is correct precisely because omitting the 3-arg `preferences:` variant leaves `config.defaultWebpagePreferences` governing. Removing `.center` from the expanded island is legitimate (regions are individually optional), no state renders empty (leading always has a StatusMark, bottom always has headline+footer across idle/working/needsYou/stale/connecting), and the FleetTotals removal genuinely de-duplicates: counts appear only in the bottom footer, host only in trailing, the needsYou count only in leading. `isLuminanceReduced` is the right Always-On condition and its opaque+outlined fallback is the right direction. Three things block: (1) the lock-screen glass measurably destroys text contrast over a bright wallpaper - computed WCAG shows textFaint falling from 3.54:1 to 1.01-2.10:1 and textDim from 6.90:1 to 1.92-4.09:1, robust across two material-darkness models and both gradient stops, and textFaint is exactly what carries the host/working/agents detail line; (2) the expanded bottom card totals ~129pt alone in its richest state against a ~160pt island
[truncated]
**Severity:** `P2`

**Findings**
- **Lock-screen glass drops textFaint/textDim below usable contrast over a bright wallpaper** (P2)
  - detail: LockScreenView now backs its content with GlassSurface(cornerRadius: 0) = .ultraThinMaterial + glassWash (navy at 0.42-0.62 alpha) instead of the opaque WidgetPalette.backdrop. I computed WCAG 2.x contrast for the three text colours. On the old opaque #13162A: text 15.68:1, textDim 6.90:1, textFaint 3.54:1 (already marginal). On glass over a bright wallpaper, modelling the material output at both #B0B0B0 and #707070 and at both gradient stops (alpha 0.42 and 0.62): text 4.37-9.30:1, textDim 1.92-4.09:1, textFaint 1.01-2.10:1. The collapse is robust across every assumption, so it does not depend on my material model. This matters because textFaint is what `lockDetail` uses for the line that carries the actual fleet state - "hostLabel · N working · M agents" at 13pt, and both stale variants at 11pt - so that line ranges from unreadable to literally invisible (1.01:1). textDim carries the question at 15pt and the "need you" caption at 13pt and fails 4.5:1 in every model. The headline survives only because 18pt semibold qualifies as large text at 3:1. Fix direction: keep activityBackgroundTint(nil) - that part is right - but stop double-layering translucency. Either drop the in-view .ultraThinMaterial and let the system surface be the only transparency, or raise the wash alpha floor substantially, or lift textFaint/textDim on the non-luminance-reduced path. Unverifiable by any test target, as the PR body correctly states, so this needs a device screenshot over a white wallpaper before merge.
  - lines: 264-270, 404-411, 300-323
- **Expanded island bottom card sits at or over the ~160pt height budget in its richest state, with nothing compressible** (P2)
  - detail: Worst case is needsYouCount > 0 with a non-empty question, !isStale and blockedSince set - a common needsYou state, not an exotic one. Bottom region arithmetic: headline Geist-SemiBold 16 (~19pt line) + question IBMPlexMono 12 (~16) + footer IBMPlexMono 11 (~14) + 2x2 inner spacing = ~53; plus 8 outer VStack spacing; plus ActivityAction at minHeight 44; plus 20 of card .padding(.vertical, 10); plus .padding(.top, 4) = ~129pt for the bottom region alone. Add the leading/trailing row (ExpandedMark's plexSemiBold(17) is ~22pt) and the system's own cutout clearance and inter-region padding and the total lands at ~160pt or above. The dangerous part is that nothing can absorb the overage: every Text is lineLimit(1) so it cannot shrink below its line height, and the button's 44 is a minHeight floor rather than a ceiling, so the system clips instead of compressing. One point in the change's favour: WidgetFont uses .custom(..., fixedSize:) throughout, so Dynamic Type does not inflate this and the budget is deterministic. Cheap headroom if you want to keep the card: .padding(.vertical, 10)->6 and .padding(.top, 4)->0 recovers 12pt, button minHeight 44->38 recovers 6 more, or lift the button out of the glass card entirely. Needs a device screenshot in exactly the needsYou+question state.
  - lines: 131-163, 327-349
- **Settings row and HtmlWebView doc assert absolute network isolation that a URL-filter rule cannot enforce once script runs** (P2)
  - detail: To be clear about what I did verify: the two other protections are not mechanically weakened by the flag. The rule is added to userContentController before loadHTMLString in the same callback so there is no window where the document loads unprotected; both the nil-store and nil-list paths load blockedPlaceholderHTML; the coordinator's didStartInitialLoad gate cancels every navigation after the first including subframe and script-initiated ones; and the 2-arg-only policy method is correct, since not implementing decidePolicyFor:preferences:decisionHandler: is exactly what keeps config.defaultWebpagePreferences authoritative. The retargeted comment ("without carrying this flag into it" rather than "without re-disabling JS there") is the right correction. The problem is the guarantee stated to the user. blockNetworkRuleListJSON is a URL filter over ^https?://, ^wss?://, ^ftps?://, ^file: and ^blob:. It can only see loads WebKit routes as page resources. With script enabled, RTCPeerConnection ICE/STUN egress is the standard content-blocker bypass, and stun:/turn: are in neither the blocked set nor blockedSchemes. [INFERENCE - no iOS runtime available here to confirm WebRTC is reachable in this WKWebView configuration; getUserMedia needs permission but ICE gathering historically does not.] The switch being opt-in and off by default is good design, but previewsSection tells the reader at the exact moment they loosen the sandbox that "The preview still cannot reach the network or navigate anywhere", and the HtmlWebView doc says a script "cannot fetch, navigate or beacon". Both are absolutes. Either soften them to what is actually enforced (no network resource loads, no navigation) or neutralize WebRTC before claiming full isolation. A security reassurance shown next to a security-loosening switch has to be exactly true.
  - file: App/HerdrApp.swift
  - lines: 7147-7150; App/HtmlWebView.swift:47-49; Sources/HerdrKit/WebViewPolicy.swift:24-32
- **Default-off is real, but the test cited as pinning it exercises a helper with no production caller** (P3)
  - detail: The default is genuinely off - I confirmed that independently via grep for register(defaults:), UserDefaults(suiteName:), defaultAppStorage and WindowGroup across App/, Sources/ and HerdrWidgets/: nothing registers or redirects this key, there is a single WindowGroup, and both @AppStorage declarations carry `= false`, so key-absent resolves false on a fresh install and on upgrade from a build that never wrote it. But WebViewPolicy.javaScriptEnabled(in:) has zero production callers - GramView.swift:159 and HerdrApp.swift:7133 both read @AppStorage(WebViewPolicy.javaScriptDefaultsKey) directly. So testJavaScriptIsOffUntilTheSettingIsTurnedOn asserts a function nothing invokes, which reduces it to checking that UserDefaults.bool(forKey:) returns false for a missing key - a property of Foundation, not of this app. Meanwhile SettingsTests' header comment states "The DEFAULT of the JavaScript switch is pinned on Linux (WebViewPolicyTests.testJavaScriptIsOffUntilTheSettingIsTurnedOn)". It is not pinned; the two `= false` literals decide it and no test covers them. The doc comment's own claim that the helper "exists so the default is a tested contract rather than an implicit one" is therefore inverted. Either route the real read through javaScriptEnabled(in: .standard) so the tested contract is the shipped one, or delete the attestation from the SettingsTests comment. Separately, nothing verifies the GramView -> HtmlWebView -> allowsContentJavaScript wiring at all; that is inherent to the app target, but worth not over-claiming.
  - file: Sources/HerdrKit/WebViewPolicy.swift
  - lines: 46-52; Tests/HerdrKitTests/WebViewPolicyTests.swift:23-43; HerdrUITests/SettingsTests.swift:3-6
- **The flag does go stale for an open preview, and only the ON wording covers it** (P3)
  - detail: updateUIView is empty and allowsContentJavaScript is captured in the init-time configuration copy, so flipping the switch never reaches a live webview. The ON branch states this correctly ("the change applies to the next preview you open"). The OFF branch says flatly "Scripts in a previewed file are ignored", which would be false for a preview already running script - and OFF is the safety-relevant direction, where a reader turning the switch off expects script to stop. Calibrating the actual risk: this is latent, not live. The preview is a fullScreenCover over a single WindowGroup, so Settings is unreachable without tapping Done, which tears down the webview; there is no App Group suite, Settings.bundle or intent that could write the key from outside while a preview is open. So there is no reachable path today. It becomes reachable the moment the preview is presented as a sheet, Settings becomes reachable over it, or iPad multi-window is supported. Give the OFF branch the same caveat, or key HtmlWebView on the flag (.id(previewJavaScript)) so a change rebuilds the view and the OFF text is unconditionally true.
  - file: App/HtmlWebView.swift
  - lines: 82; App/HerdrApp.swift:7147-7150
- **Island glass card contradicts GlassSurface's own doc comment and is not the iOS idiom it claims** (P3)
  - detail: GlassSurface's doc says it is "Used where the backdrop is the wallpaper or the island - never over an opaque fill, which would throw away the blur and leave only the wash." The expanded Dynamic Island background is an opaque black fill - it is the physical cutout region, not a translucent surface over the wallpaper - so the island call site at line 160 is exactly the case the comment forbids. The rendered result is .ultraThinMaterial over black, i.e. the wash plus a white-at-0.14 border and no actual glass. The stated rationale ("so the island reads like the rest of iOS rather than text floating on black") is also based on a false premise: first-party expanded Dynamic Islands do put content directly on black, so a bordered card inside the island is the less idiomatic choice, not the more. Either accept it as a deliberate brand decision and fix the comment, or drop the card in the island and keep GlassSurface for the lock screen only. On the related question of whether Material is safe in a Live Activity archive: it will not crash - it is not an unsupported-API trap - but a Live Activity is rendered out-of-process from an archived hierarchy, so the blur is not guaranteed to sample the wallpaper, and the whole visual premise of this change depends on that guarantee. [INFERENCE - needs a device screenshot; Apple's documented control for this surface is activityBackgroundTint(_:), and the supported vibrant-surface primitive for lock-screen archives is AccessoryWidgetBackground, neither of which is a Material.]
  - lines: 193-212, 160
- **GlassSurface(cornerRadius: 0) strokes a square border inside the system's rounded container** (P3)
  - detail: The lock-screen Live Activity is clipped by the system to a rounded rect, but GlassSurface(cornerRadius: 0) traces a zero-radius RoundedRectangle with .strokeBorder(glassEdge, lineWidth: 0.75). The corners of that rectangle fall outside the system's rounded clip and are removed, so the intended "lit edge" renders as four straight hairlines that stop short near each corner. strokeBorder also insets by only half the line width (0.375pt), putting the stroke essentially on the clip boundary, so it may be largely clipped along the flat edges too - the edge either disappears or looks broken at the corners. Suppress the border when cornerRadius == 0, or pass the system's activity corner radius so the stroke follows the visible shape.
  - lines: 268, 207-210
- **ExpandedMark's .fixedSize() opts out of truncation and its count formatting diverges from CompactCount** (P3)
  - detail: ExpandedMark applies .lineLimit(1) then .fixedSize() to Text(verbatim: "\(state.needsYouCount)") at plexSemiBold(17). .fixedSize() forces the intrinsic size, which means lineLimit(1) cannot truncate - the text takes its full width in the narrowest region on screen. CompactCount, by contrast, caps at "99+" via Int.compactCount. So a three- or four-digit count overflows the leading corner, which is the exact overflow class this commit set out to eliminate. Use .compactCount here and drop .fixedSize(). Related: ExpandedHost is now the sole carrier of the machine name, since footer's comment deliberately omits it ("The machine's name is the trailing corner's job"), yet it lives in a ~60pt corner at 11pt with truncationMode(.tail) - a realistic hostname like "my-dev-box.local" truncates to a few characters, so the expanded island can lose host identity entirely while the full-width bottom card has ample room for it. That relocates the squeeze problem rather than solving it.
  - lines: 96-104, 68-72, 111-121, 165-166
- **Stale ViewBuilder child-count comment in indexStack** (P3)
  - detail: The comment reads "Grouped subviews keep this builder well under SwiftUI's 10-child ViewBuilder ceiling (7 children + the footers Group)". Adding previewsSection at line 6078 makes it 8 children plus the Group = 9. The comment exists specifically to track that ceiling, so an off-by-one here actively misleads the next person adding a section - they would believe there are two slots left when there is one. Update the count. (Placement itself is correct: previewsSection follows the exact dual pattern of troubleSection/helpSection/supportSection/aboutSection - inline on the iPhone index, in aboutDetail for iPad - so there is no double render in any size class. Minor nit: aboutDetail's subtitle "Trouble, help, support & legal" no longer describes its now-first section.)
  - file: App/HerdrApp.swift
  - lines: 6072-6073

**Tests Run**
- {"command":"git rev-parse HEAD; git log --oneline -5; git diff --stat 3504d56...HEAD","name":"git verification of review target","result":"PASS - head confirmed eff9c9d8ede369ccf89a8cdb568bcc056260011a on base 3504d56; 7 files changed, +280/-86, matching the described scope (WebViewPolicy, HtmlWebView, GramView, HerdrApp, AgentLiveActivity, plus the two test files)"}
- {"command":"git diff 3504d56...HEAD -- <each changed path>; read App/HtmlWebView.swift, HerdrWidgets/AgentLiveActivity.swift, Sources/HerdrKit/WebViewPolicy.swift, App/HerdrApp.swift:6060-6090,6230-6253","name":"full diff and source read of all changed files","result":"PASS - reviewed every changed line in context, including makeUIView, the Coordinator's single 2-arg policy method, the full rule-list JSON, ExpandedBody/ExpandedMark/ExpandedHost/GlassSurface/LockScreenView/ActivityAction/StatusMark/WidgetPalette, and both SettingsView insertion points"}
- {"command":"grep -rn \"register(defaults|UserDefaults(suiteName|defaultAppStorage|WindowGroup\" App/ Sources/ HerdrWidgets/","name":"default-off audit (fresh install and upgrade)","result":"PASS - only match is App/HerdrApp.swift:29 WindowGroup. No defaults registration, no suite redirection, no defaultAppStorage anywhere, single scene. Confirms an absent 'previews.javascript' key resolves to the @AppStorage `= false` on both a fresh install and an upgrade from build 143"}
- {"command":"grep -rn \"javaScriptEnabled|previews\\.javascript|javaScriptDefaultsKey|allowsJavaScript\"","name":"JS flag reference audit","result":"PASS (with finding) - 4 production references: WebViewPolicy.swift:44 key, GramView.swift:159/365, HerdrApp.swift:7133, HtmlWebView.swift:34/50. WebViewPolicy.javaScriptEnabled(in:) at :50 appears only in its own definition and in WebViewPolicyTests - zero production callers, which is finding P3-4"}
- {"command":"grep -rn \"AccessoryWidgetBackground|containerBackground|ultraThinMaterial|Material\" HerdrWidgets/","name":"widget background convention audit","result":"PASS (with finding) - single match, AgentLiveActivity.swift:202 .fill(.ultraThinMaterial). No AccessoryWidgetBackground anywhere in the extension, so this PR introduces Material into the widget target with no existing precedent in the repo"}
- {"command":"python: sRGB linearization + relative luminance + (L1+0.05)/(L2+0.05) for text #EEF0F7, textDim #99A0BC, textFaint #666D91 against opaque #13162A and against glassWash (0.42 and 0.62 alpha) composited over material models #B0B0B0 and #707070","name":"WCAG contrast computation for glass vs opaque backdrop","result":"EXECUTED - old opaque backdrop: text 15.68:1, textDim 6.90:1, textFaint 3.54:1. Glass over bright wallpaper: text 4.37-9.30:1, textDim 1.92-4.09:1, textFaint 1.01-2.10:1. textFaint fails in all four glass models; textDim fails 4.5:1 in all four. Basis for finding P2-1"}
- {"command":"manual layout sum of ExpandedBody in the needsYou+question+blockedSince state against the ~160pt expanded-island ceiling","name":"expanded island height budget arithmetic","result":"EXECUTED (static) - ~129pt for the bottom region alone (53 text block + 8 spacing + 44 ActivityAction minHeight + 20 card padding + 4 top padding), plus ~22pt leading/trailing row plus system cutout clearance. Confirmed WidgetFont uses .custom(fixedSize:) throughout, so Dynamic Type does not compound it. Basis for finding P2-2"}
- {"command":"which swift swiftc; swift --version","name":"Linux swift test WebViewPolicyTests (claimed 3/3)","result":"NOT RUN - 'command not found: swift'. No Swift toolchain in this sandbox. The claimed 3/3 result is accepted as reported but was not independently confirmed"}
- {"command":"gh run view 34895884300 --json displayTitle,status,conclusion,headSha,jobs","name":"CI run 34895884300 (claimed SUCCESS, testJavaScriptPreviewSwitchFlipsAndExplainsItself green)","result":"NOT RUN - gh refused: 'To use GitHub CLI in automation, set the GH_TOKEN environment variable.' No network/auth. The reported SUCCESS at this head, including the app and widget-extension compile, is accepted as reported but not independently confirmed. Note on the test's own design: it branches on leftover simulator state (`if !willRun.exists { row.tap() }`), so in a leftover-ON run it never exercises the OFF->ON transition - normalizing to OFF first would make both directions unconditional"}

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d54ed3deb5637f-1`

**Summary:** Re-reviewed 4d001834690593e7617a69e1b4cb154852e971b5 (base 3504d56; 7 files, +321/-95 vs base; +61/-29 in the new commit) by full read of every changed file. No Swift toolchain exists in this sandbox (`swift`, `swiftc`, `xcodebuild` all absent) and no CI logs were mounted, so I could not execute the cited SettingsTests run 34899430370 or `swift test`; I take those at the sender's word and label my own evidence static_only. What I DID execute is the arithmetic, in a Python kernel over the full worst-case grid.

f1 CHECKED AND DISCHARGED. Your figures are exact. Compositing the wash at 0.80 in gamma-space sRGB over material outputs #E6E6E6/#B0B0B0/#707070 at both stops, the worst case is #E6E6E6 over the 0x1B1F3A stop -> bg rgb(67.6,70.8,92.4), giving glassTextDim 5.89:1 and glassTextFaint 4.53:1. The two additionally survive every other cell (dim 5.89-9.37, faint 4.53-7.22), so AA for small text holds across the grid, not just on average. One correction worth recording: your 'before' numbers of 2.04 and 1.05 are against the OLD 0.62/0.42 wash. At the NEW 0.80 wash the opaque pair would measure 3.53:1 and 1.81:1 - still under AA - so the wash change alone would not have fixed this and the new inks are genuinely load-bearing, not belt-and-braces. No glass-backed TEXT still reads with the old inks: grep for WidgetPalette.(textDim|textFaint) in the widget returns exactly three hits, two of which are the isLuminanceReduced ternaries in `dim`/`faint` (correct - the Always-On branch is the opaque backdrop those were tuned for) and the third is a non-text fill (see new finding f10). The gating is on the same environment value that swaps the background, so the ink and the surface cannot desynchronise.

f2 CHECKED AND DISCHARGED. Recomputing at ~1.30 line-height: headline 20.8 + 2 + one secondary 15.6 + 6 + button 36 + 12 vertical padding = 92.4pt, matching your ~93. The leading/trailing row is set by ExpandedMark's 17pt numeral at ~22.1pt, so content lands near 114.5pt agains
[truncated]
**Severity:** `P2`

**Findings**
- **Lock-screen glass drops textFaint/textDim below usable contrast over a bright wallpaper** (P2)
  - continues_uid: jerryfane/herdrup#254-f1
  - detail: DISCHARGED at 4d00183. Verified by recomputing WCAG 2.x relative-luminance contrast in a Python kernel over the full worst-case grid: material output in {#E6E6E6,#B0B0B0,#707070} x both wash stops {0x1B1F3A,0x13162A}, alpha 0.80, gamma-space sRGB alpha blend. Worst cell is #E6E6E6 over 0x1B1F3A -> bg rgb(67.6,70.8,92.4): glassTextDim (#C9CFE2, :441) = 5.89:1, glassTextFaint (#AEB6D0, :442) = 4.53:1. Your arithmetic is exact. Across the whole grid dim ranges 5.89-9.37 and faint 4.53-7.22, so AA for small text holds in every cell rather than on average. Correction for the record: the 2.04/1.05 'before' figures are against the OLD 0.62/0.42 wash - at the new flat 0.80 wash the opaque pair would still measure only 3.53:1 and 1.81:1, so the wash change alone would NOT have cleared AA and the new inks are load-bearing. No glass-backed text retains the old inks: grep WidgetPalette.(textDim|textFaint) returns three hits, two being the isLuminanceReduced ternaries at :289-290 (correct: that branch renders over the opaque WidgetPalette.backdrop these were tuned for) and one non-text fill at :400 now tracked as a separate P3. Both `dim` and `faint` key off the same isLuminanceReduced value that swaps the background at :277-281, so ink and surface cannot desynchronise. Island-card side also checked: over the island's opaque black the same inks land at 8.6:1 and 11.2:1.
  - rationale: Checked at this head; the fix is correct and the arithmetic reproduces independently.
  - state: answered
- **Expanded island bottom card sits at or over the ~160pt height budget in its richest state, with nothing compressible** (P2)
  - continues_uid: jerryfane/herdrup#254-f2
  - detail: DISCHARGED at 4d00183. Recomputed the richest state (needsYouCount > 0, question present, not stale) at a ~1.30 line-height factor: headline 16pt -> 20.8, inner spacing 2, exactly one secondary line 12pt -> 15.6, outer spacing 6, ActivityAction minHeight 36 (:154), plus 6+6 vertical padding (:159) = 92.4pt, matching your ~93. ExpandedMark's 17pt semibold numeral (:98) sets the leading/trailing row at ~22.1pt, so expanded content lands near 114.5pt against the ~160pt budget, versus ~154pt plus system insets at eff9c9d. Roughly 45pt of headroom, and the `secondary` ViewBuilder (:165-175) makes the two-secondary-lines case structurally unreachable rather than merely unlikely - the if/else cannot emit both. Dropping the counts when a question exists loses nothing load-bearing: needsYouCount stays in the leading corner (:96-101), hostLabel in the trailing (:111-121), and the lock screen's lockDetail still prints 'hostLabel / N working / M agents' (:338). The only signal genuinely lost in that one state is the elapsed 'waiting <timer>', which neither the compact nor the minimal presentation ever carried. The stale case is also right: `!isStale` fails first, so a stale activity with a question still shows 'Not updated recently' rather than a question that is no longer trustworthy. ActivityAction's default of 44 (:355) keeps the lock-screen touch target intact while only the island passes 36.
  - rationale: Checked at this head; worst case now fits with margin and the compression path is structural.
  - state: answered
- **Settings row and HtmlWebView doc assert absolute network isolation that a URL-filter rule cannot enforce once script runs** (P2)
  - continues_uid: jerryfane/herdrup#254-f3
  - detail: STILL OPEN at 4d00183 - re-read all four sites, none changed since eff9c9d (the new commit touches only HerdrApp.swift's ViewBuilder comment/subtitle and AgentLiveActivity.swift). This is the one P2 the handoff did not list among the three addressed.\n\nThe overclaiming sites: (1) WebViewPolicy.swift:5-11 - 'when we render one we disable JavaScript (at the WebView config level) and block ALL network egress, so a hostile file can neither execute script nor beacon its host's data out'. This PR adds `javaScriptDefaultsKey` (:44) and `javaScriptEnabled(in:)` (:50) to THIS VERY FILE while leaving that sentence asserting script is unconditionally disabled, so the type doc now contradicts the API directly beneath it. (2) HtmlWebView.swift:14-19 - '**All network is blocked** ... so a passive subresource load cannot beacon out'; the 'passive' qualifier is the accurate part, but the bolded headline claim is not. (3) GramView.swift:359-361 - 'all network blocked, navigation frozen, and script off unless the reader turned it on' places the absolute claim and the conditional one in the same breath. (4) HerdrApp.swift:7154, the user-facing ON copy - 'The preview still cannot reach the network or navigate anywhere'.\n\nWhy a url-filter cannot back that: blockNetworkRuleListJSON (:24-32) is five literal-prefix scheme rules. A WKContentRuleList trigger matches a URL load, so it covers fetch/XHR/sendBeacon/<a ping>/subresources - genuinely everything the passive path can do, which is why the claim was true before this PR. It does not cover egress that never presents a URL to the loader. [INFERENCE - I could not execute this in a simulator here] the concrete channel is RTCPeerConnection: a data-channel-only peer connection needs no camera/mic permission, and its ICE gathering emits UDP to STUN/TURN with attacker-chosen hostnames, which no content rule inspects. Navigation egress is separately and correctly closed by the coordinator (:109-120), and the fail-closed guards (:63-6
[truncated]
  - file: Sources/HerdrKit/WebViewPolicy.swift
  - rationale: Re-read WebViewPolicy.swift, HtmlWebView.swift, GramView.swift and HerdrApp.swift's previewsSection at this exact head; all four claims are unchanged and this PR is what made the JS clause conditional.
  - state: open
- **Non-text and accent colours on the lock-screen glass still assume the opaque backdrop that f1's fix moved away from** (P3)
  - detail: NEW, and the residue f1's fix did not reach: the dim/faint TEXT pair is now surface-aware, but three non-text/accent colours on the same glass are not. Measured against the identical worst case f1 uses (material #E6E6E6 over the 0x1B1F3A stop, bg rgb(67.6,70.8,92.4)):\n\n(1) StatusMark's .idle branch (:398-402) fills the inner dot with WidgetPalette.textFaint -> 1.81:1. On the lock screen this dot renders on glass (lockHero's else branch, :314) and is the only thing distinguishing idle from working/stopped whenever needsYouCount == 0. WCAG 1.4.11 wants 3:1 for a meaningful non-text indicator. Not blocking because no information is carried exclusively by the dot - the headline (8.04:1) and lockDetail's counts (4.53:1) are both legible in that state - but the mark is the element the design leans on.\n\n(2) WidgetPalette.color(.idle) (:464) returns the same textFaint, so an idle activity's keylineTint and any markColor-tinted control inherit 1.81:1.\n\n(3) LockScreenView:247-250 paints AgentActivitySummary.line at 13pt in state.markColor, which is #E9A63C for needsYou -> 4.35:1, just under AA for that size. On the pre-PR opaque backdrop the same pair measured 7.66:1, so this is a regression the glass surface introduced rather than a pre-existing miss. `working` #5B9BE8 is 3.18:1 and `died` #E2584E is 2.50:1 on the same surface, should either ever land on glass at text size.\n\nSame remedy shape as f1: a glass-aware variant, or lift the accents a little. Sizing note if you touch (3): 13pt is normal text for WCAG, not large.
  - rationale: Computed from the same worst-case composite as f1 in the same kernel run; the surface changed under these colours and only the text pair was re-tuned.
- **Default-off is real, but the test cited as pinning it exercises a helper with no production caller** (P3)
  - continues_uid: jerryfane/herdrup#254-f4
  - detail: STILL OPEN at 4d00183, unchanged. Grepped javaScriptEnabled|previewJavaScript|javaScriptDefaultsKey across the tree: `WebViewPolicy.javaScriptEnabled(in:)` (:50-52) has exactly two references and both are in Tests/HerdrKitTests/WebViewPolicyTests.swift:33,37,41. Both production reads bypass it - GramView.swift:159 and HerdrApp.swift:7134 are `@AppStorage(WebViewPolicy.javaScriptDefaultsKey) private var previewJavaScript = false`. Default-off IS real, and I re-confirmed why: the literal `= false` in both @AppStorage declarations supplies it, and there is no register(defaults:) or defaultAppStorage anywhere to override an absent key. But testJavaScriptIsOffUntilTheSettingIsTurnedOn proves a helper's semantics, not the app's default; what actually prevents drift is the shared key constant, which the test does exercise. SettingsTests.swift:3-6 makes a stronger claim than the test supports ('The DEFAULT of the JavaScript switch is pinned on Linux'). Either route the app through the helper - then the test covers production - or soften that comment to say the KEY is pinned and the default lives in the two @AppStorage declarations.
  - file: Sources/HerdrKit/WebViewPolicy.swift
  - rationale: Re-grepped at this head; the helper still has zero production callers and the UI-test comment still overstates what is pinned.
  - state: open
- **The flag does go stale for an open preview, and only the ON wording covers it** (P3)
  - continues_uid: jerryfane/herdrup#254-f5
  - detail: STILL OPEN at 4d00183, unchanged. HerdrApp.swift:7153-7155: the ON string carries 'the change applies to the next preview you open'; the OFF string is still 'Scripts in a previewed file are ignored. Turn this on only if you need an interactive report to work.' with no staleness note. The asymmetry runs the wrong way - the staleness matters more when a reader turns protection back ON, because HtmlWebView's flag is captured into WKWebViewConfiguration at init (HtmlWebView.swift:39-52) and updateUIView is empty (:82), so a preview already on screen keeps running script until dismissed. Simplest fix is to move the staleness clause out of the ternary so it appears in both states.
  - file: App/HerdrApp.swift
  - rationale: Re-read previewsSection at this head; the OFF branch still omits the clause.
  - state: open
- **Island glass card contradicts GlassSurface's own doc comment and is not the iOS idiom it claims** (P3)
  - continues_uid: jerryfane/herdrup#254-f6
  - detail: STILL OPEN at 4d00183; the only change is cornerRadius 18 -> 16 at :161. GlassSurface's doc (:205-208) says it is for backdrops that are 'the wallpaper or the island - never over an opaque fill, which would throw away the blur and leave only the wash', but the expanded Dynamic Island's own background IS an opaque fill, so the sentence licenses and forbids the same use in one breath. The new 0.80 wash sharpens this: over the island I measure the card at roughly rgb(22,25,46), i.e. a flat ~#161831 where .ultraThinMaterial contributes essentially nothing and only the wash and the 0.75pt edge are visible. That is a fine-looking card, and it is a hair distinct from WidgetPalette.backdrop, but it is a painted panel rather than glass. Fix is one honest sentence in the doc plus dropping the material for the island path, or accepting the card as paint and saying so.
  - rationale: Re-read GlassSurface and its island callsite at this head; the doc still contradicts itself and the 0.80 wash makes the material contribution negligible there.
  - state: open
- **GlassSurface(cornerRadius: 0) strokes a square border inside the system's rounded container** (P3)
  - continues_uid: jerryfane/herdrup#254-f7
  - detail: STILL OPEN at 4d00183, unchanged. LockScreenView:280 still passes cornerRadius 0, so GlassSurface's strokeBorder (:219-222) traces a square 0.75pt rectangle inside the system's rounded Live Activity container; the four corner segments are clipped away and the edge reads as four detached straight lines rather than a pane outline. The fill is unaffected. Either give the lock-screen path the system's container radius, or suppress the stroke when cornerRadius is 0 - the edge is decorative there, since the system already draws the container boundary.
  - rationale: Re-read the lock-screen background branch at this head; the callsite is unchanged.
  - state: open
- **ExpandedMark's .fixedSize() opts out of truncation and its count formatting diverges from CompactCount** (P3)
  - continues_uid: jerryfane/herdrup#254-f8
  - detail: STILL OPEN at 4d00183, unchanged. ExpandedMark:103-104 sets .lineLimit(1) then .fixedSize(), and fixedSize wins - the corner takes its ideal width and can push into the region the doc at :82-83 says is only about three characters wide, which is exactly the squeeze the redesign was meant to end. Separately :97 formats the count as "\(state.needsYouCount)" while CompactCount:69 uses .compactCount, so a fleet of 100 waiting agents renders '100' in the expanded leading corner and '99+' in the compact trailing one - the same number, two spellings, in two presentations of the same activity. Using .compactCount at :97 fixes both, since '99+' is also the widest string the corner would then have to fit.
  - rationale: Re-read ExpandedMark and CompactCount at this head; both the modifier and the formatting divergence are unchanged.
  - state: open
- **Stale ViewBuilder child-count comment in indexStack** (P3)
  - continues_uid: jerryfane/herdrup#254-f9
  - detail: DISCHARGED at 4d00183. Counted the actual children of the VStack at HerdrApp.swift:6075-6088: atAGlanceSection, manageSection, appearanceSection, previewsSection, troubleSection, helpSection, supportSection, aboutSection = 8, plus the footers Group = 9 direct children, one under SwiftUI's 10-child ceiling. The comment at :6072-6074 now states exactly that and adds the useful part - that the next section needs its own Group. aboutDetail's subtitle at :6241 reads 'Previews, trouble, help, support & legal', so the iPad detail no longer hides a section it renders (:6243).
  - file: App/HerdrApp.swift
  - rationale: Counted the children directly at this head; comment and subtitle both match the code.
  - state: answered

**Tests Run**
- git rev-parse HEAD -> 4d001834690593e7617a69e1b4cb154852e971b5 (confirmed review is at the requested head; working tree clean)
- git diff --stat 3504d56..HEAD -> 7 files, +321/-95; git diff eff9c9d..HEAD -> 2 files, +61/-29 (full text of both diffs read)
- Python kernel: WCAG 2.x relative-luminance contrast over material {#E6E6E6,#B0B0B0,#707070} x wash stops {0x1B1F3A,0x13162A} at alpha 0.80, gamma-space sRGB blend. Worst case #E6E6E6/0x1B1F3A: glassTextDim 5.89:1, glassTextFaint 4.53:1 (confirms the sender's figures); textDim 3.53:1 and textFaint 1.81:1 at the same wash; waiting 4.35:1, working 3.18:1, died 2.50:1; opaque-backdrop baseline for waiting 7.66:1
- Python kernel: expanded-island bottom card height, richest state = 92.4pt (headline 20.8 + 2 + secondary 15.6 + 6 + button 36 + 12 padding), leading row 22.1pt, total ~114.5pt vs ~160pt budget
- grep 'WidgetPalette\.(textDim|textFaint)' HerdrWidgets/AgentLiveActivity.swift -> 3 hits: :289, :290 (isLuminanceReduced ternaries, correct) and :400 (idle dot fill on glass, new f10)
- grep 'javaScriptEnabled|previewJavaScript|javaScriptDefaultsKey' across tree -> helper referenced only from Tests/HerdrKitTests/WebViewPolicyTests.swift:33,37,41; production reads are @AppStorage at GramView.swift:159 and HerdrApp.swift:7134
- Full read of AgentLiveActivity.swift (503 lines), HtmlWebView.swift (122), WebViewPolicy.swift (53), WebViewPolicyTests.swift diff, SettingsTests.swift (46), and the changed regions of HerdrApp.swift (6059-6098, 7129-7166) and GramView.swift (154-160, 353-381)
- Manual child count of indexStack's VStack at HerdrApp.swift:6075-6088 -> 8 children + 1 Group = 9
- NOT RUN, no toolchain: `swift`, `swiftc` and `xcodebuild` are all absent (`which` returns nothing; `swift --version` -> command not found), and no CI logs were mounted under the evidence cache. The cited SettingsTests run 34899430370 and the Linux WebViewPolicyTests 3/3 are taken on the sender's word, unverified here. Everything the Live Activity actually renders remains visually unverified, as in the prior round.
- Read GITMOOT_PRIOR_VERDICTS (evidence/prior-verdicts.json, as_of 2026-09-14T21:58:33Z, 41 verdicts): exactly one prior verdict for PR 254 - herdr-ios-review-opus at eff9c9d8, changes_requested, P2, static_only, 9 findings

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d550b1687946d6-1`

**Decision:** `approved`
**Summary:** Third-round review of jerryfane/herdrup PR #254 at exact head d1e998357c45f2210f9e56f1dc2a8a0d8a23fa1f (git rev-parse HEAD matches; delta 4d00183..d1e9983 = 5 files, +52/-27; whole PR vs 3504d56 = 7 files, +363/-112; tree clean). Prior verdicts read from GITMOOT_PRIOR_VERDICTS (frozen 2026-09-14T22:33:06Z): rounds 1 (eff9c9d, changes_requested/P2/9) and 2 (4d00183, changes_requested/P2/10). EVIDENCE static_only AND NOT BY CHOICE: `which swift swiftc xcodebuild` is empty and `gh auth status` reports no login, so SettingsTests run 34902493632 and the Linux WebViewPolicyTests 3/3 could neither be re-run nor fetched; they are the sender's evidence, not mine. WHAT I DID EXECUTE: git rev-parse/log/diff --stat, the full 4d00183..d1e9983 diff, targeted reads of all five changed files plus the unchanged surfaces the ledger made relevant, grep over every JS-flag symbol, and a Python re-implementation of WebKit's case-insensitive url-filter matching against the rule list at this head plus the count-spelling and corner-width arithmetic.

BOTH OPEN FINDINGS ARE CLOSED, and neither changed behaviour (f3 is comment/copy only; f8's only executable change is two layout modifiers and one string spelling).

f3 (P2) DISCHARGED. Every absolute the finding named is gone and each replacement is backed. (a) 'cannot LOAD anything over the network' is backed by the rule list: I re-implemented the url-filter match (NSRegularExpression .caseInsensitive semantics) and all of http/https/ws/wss/ftp/ftps/file/blob block, including uppercase scheme, explicit port and query-string variants, while about:blank and both data: forms stay unblocked, so the #92 render is preserved; the viewer fails closed on both a nil store and a nil compiled list (HtmlWebView.swift:71-83). (b) 'or open another page' is backed: Coordinator.decidePolicyFor cancels every navigation once didStartInitialLoad is set (HtmlWebView.swift:122-127), which covers link taps, form submits, meta-refresh (.other) and subframe navigatio
[truncated]
**Severity:** `P3`

**Findings**
- **Settings row and HtmlWebView doc assert absolute network isolation that a URL-filter rule cannot enforce once script runs** (P2)
  - continues_uid: jerryfane/herdrup#254-f3
  - detail: CLOSED at d1e9983. Checked by reading all four rewritten sites and re-deriving what the implementation can back. WebViewPolicy.swift:5-20 now says a rendered attachment gets no script by default and cannot LOAD anything over the network, then separates the mechanism ('a content rule matches URL loads') from what it does not inspect ('egress that never presents a URL to the loader, RTCPeerConnection ICE gathering being the obvious one'), and bounds the residue to payload + IP + time from an about:blank no-base-URL origin. HtmlWebView.swift:14-29 carries the same split ('Remote loads are blocked' plus the script-ON caveat paragraph), and its inline comment at :54-57 no longer claims a script cannot beacon -- it now says 'It is not a hermetic seal'. GramView's three comments say 'no remote loads' (:164-166, :360, :1722). The Settings ON row (HerdrApp.swift:7155) admits a file 'could signal that you opened it'.

What I ran: a Python re-implementation of WebKit's case-insensitive url-filter matching over the five rules at this head against http/https/ws/wss/ftp/ftps/file:///etc/passwd/blob:, plus HTTP:// , FILE:// , http://1.2.3.4:8080/p and https://h/x?d=stolen -- all 12 blocked, zero escapes -- and against about:blank, data:text/html,hi and data:image/png;base64,AAAA -- none blocked, so the #92 render is preserved. Traced the two remaining absolutes to their enforcement: 'cannot load anything over the network' is the rule plus fail-closed (HtmlWebView.swift:71-83, both the nil-store and nil-list arms load the script-free placeholder); 'or open another page' is Coordinator.decidePolicyFor cancelling every navigation after the first (:122-127), which covers .linkActivated, .formSubmitted, .other/meta-refresh and subframe navigations, with no createWebViewWith implemented so window.open and target=_blank are inert and pushState cannot cross origin from a nil base URL.

Judgement on the two questions asked. No claim remains that the implementation cannot back, with
[truncated]
  - file: Sources/HerdrKit/WebViewPolicy.swift
  - line: 12
  - rationale: The finding was about claims exceeding enforcement. At this head every user-facing and doc claim maps to a mechanism I could verify by matching the rule list and reading the navigation delegate, and the one class of egress that escapes is stated rather than denied.
  - state: answered
- **ExpandedMark's .fixedSize() opts out of truncation and its count formatting diverges from CompactCount** (P3)
  - continues_uid: jerryfane/herdrup#254-f8
  - detail: CLOSED at d1e9983, with one residual named below. Count spelling: ExpandedMark:101 is now Text(state.needsYouCount.compactCount) and CompactCount:69/74 uses the identical private Int.compactCount at :496 (`self >= 100 ? "99+" : String(self)`), so I confirmed by arithmetic over 1/9/99/100/250 that the compact pill and the island corner agree at every value. needsYouCount is Int (Shared/AgentActivityState.swift:27), so the private extension resolves inside the widget target; Text(_: String) is the StringProtocol overload, which does not localize -- the same guarantee Text(verbatim:) gave, so no string-catalog surprise. Truncation: .fixedSize() is gone, replaced by .lineLimit(1) + .truncationMode(.tail) on the HStack, which propagate to the Text. StatusMark is rigid on every status arm (.frame(width: diameter, height: diameter), :392-417), so under a narrow proposal SwiftUI squeezes the flexible Text and keeps the dot -- the right ordering, and it protects the bottom card that f2 was about.

RESIDUAL (non-blocking, unverifiable here): the corner is now the ONLY place the expanded island shows the waiting count -- ExpandedBody's headline is the lead agent's name and its footer prints '<n> working - <m> agents' (:201), never needsYouCount -- so where a grown corner previously cost layout, a truncated one costs the number itself. If the leading region is as narrow as this file's own doc comment claims ('about as wide as three characters', :82-83), then after 4pt padding + 10pt mark + 6pt spacing only ~1.0 characters of 17pt IBM Plex Mono fit, and '99+' would render as '9…' -- a wrong count rather than a clipped one. I cannot measure the real region width in any test target, and the same doc estimate would have made the previous .fixedSize() behaviour catastrophic, so I read it as rhetorical about the ideal budget rather than a hard clip. Filed as an observation, not a request.
  - line: 101
  - rationale: Both halves of the finding are mechanically closed and I verified them by arithmetic and by reading the layout modifiers. The residual is a consequence of the chosen fix that no test target can observe, so it is recorded rather than blocked on.
  - state: answered
- **Lock-screen glass drops textFaint/textDim below usable contrast over a bright wallpaper** (P2)
  - continues_uid: jerryfane/herdrup#254-f1
  - detail: Re-observed at d1e9983 because the ledger's relevance key HerdrWidgets/AgentLiveActivity.swift was touched. STILL DISCHARGED, and the fix is untouched: the delta 4d00183..d1e9983 modifies ExpandedMark ONLY (10 lines, all inside :94-110). Confirmed by reading the surfaces the finding named at this head: glassWash is still the 0.80-opacity two-stop gradient (:435-439), glassTextDim #C9CFE2 and glassTextFaint #AEB6D0 still exist with the contrast rationale on them (:443-448), and LockScreenView still routes all secondary/tertiary ink through the isLuminanceReduced-switched dim/faint computed vars (:295-296), consumed at :250, :316, :335, :340 and :346 -- so the glass path gets the lifted pair and only the opaque Always-On backdrop (:284) gets textDim/textFaint. No new call site in the delta introduces the opaque pair on a glass surface.
  - line: 295
  - rationale: A P2 answered at the prior head must be re-observed when the diff touches its file; the diff touches the file but not the mechanism, and I checked every colour site the finding covered.
  - state: answered
- **Expanded island bottom card sits at or over the ~160pt height budget in its richest state, with nothing compressible** (P2)
  - continues_uid: jerryfane/herdrup#254-f2
  - detail: Re-observed at d1e9983 for the same relevance key. STILL DISCHARGED. ExpandedBody (:133-209) is byte-identical to 4d00183 -- the delta touches only ExpandedMark -- so the one-secondary-line fix stands: the card is headline (geistSemiBold 16, lineLimit(1)) plus EXACTLY ONE secondary line (question at plex 12 or the footer HStack at plex 11, both lineLimit(1)) plus the conditional 36pt action, with 6pt spacing, 2pt inner spacing and 12pt vertical padding, which I total at roughly 87pt against the ~160pt budget the comment at :144-149 states. The delta's only effect on the expanded layout is in the leading corner and is width-only; and it moves in this finding's favour, since dropping .fixedSize() removes the corner's ability to take its ideal width and steal from the regions around it.
  - line: 144
  - rationale: Same re-observation obligation as f1. The height mechanism is unchanged and the one layout change in the delta reduces rather than adds pressure on the regions this finding was about.
  - state: answered
- **Stale ViewBuilder child-count comment in indexStack** (P3)
  - continues_uid: jerryfane/herdrup#254-f9
  - detail: Re-observed at d1e9983 because the delta touches App/HerdrApp.swift. STILL DISCHARGED and still accurate. I counted the builder's children at this head: atAGlanceSection, manageSection, appearanceSection, previewsSection, troubleSection, helpSection, supportSection, aboutSection = 8, plus the footers Group = 9 (:6075-6088), which is exactly what the comment at :6072-6074 claims, including 'one slot left: the next section needs its own Group'. The delta's HerdrApp.swift change is confined to previewsSection's doc comment (:7139-7141) and its two copy strings (:7155-7156) and adds no child.
  - file: App/HerdrApp.swift
  - line: 6072
  - rationale: Answered at the prior head and re-checked here because the file changed; the comment's arithmetic still matches the code it describes.
  - state: answered
- **Default-off is real, but the test cited as pinning it exercises a helper with no production caller** (P3)
  - continues_uid: jerryfane/herdrup#254-f4
  - detail: Checked at this head and NOT fixed -- unchanged, still stands. I grepped every occurrence of javaScriptEnabled|previewJavaScript|allowsJavaScript|javaScriptDefaultsKey across App/, Sources/, Shared/, HerdrWidgets/ and the test targets. WebViewPolicy.javaScriptEnabled(in:) (:59-61) is referenced from exactly two places: WebViewPolicyTests.swift:33/37/41 and its own doc comment at :16. Both production reads go through @AppStorage(WebViewPolicy.javaScriptDefaultsKey) with an explicit `= false` literal -- GramView.swift:159 (consumed at :366 constructing HtmlWebView) and HerdrApp.swift:7134 (the toggle binding at :7146) -- neither of which calls the function the cited test exercises. The shared key constant does prevent the one drift that matters most, and SettingsTests.swift:3-6 is candid that a UI test cannot observe a fresh install, so the default is still genuinely off on install and on upgrade; what the Linux test pins is a parallel helper, not the production read. Non-blocking and unchanged in this round's scope.
  - file: Sources/HerdrKit/WebViewPolicy.swift
  - line: 59
  - rationale: The finding is about a test that does not execute the production path; at this head the production path is still @AppStorage and the helper is still test-only, so the observation is recorded again rather than discharged.
  - state: answered
- **The flag does go stale for an open preview, and only the ON wording covers it** (P3)
  - continues_uid: jerryfane/herdrup#254-f5
  - detail: Checked at this head and NOT fixed -- the asymmetry survives the rewording. The staleness is real: HtmlWebView captures allowsJavaScript into config.defaultWebpagePreferences before WKWebView(frame:configuration:) copies it (:58-60) and updateUIView is empty (:90), so an already-presented preview keeps whatever value it was built with. The ON string still carries the caveat, now as 'Applies to the next preview you open.' (:7155). The OFF string (:7156) reads 'Scripts in a previewed file are ignored, and it can't load anything from the network. Turn this on only for a file you trust that needs to be interactive.' -- it gained a network clause and lost nothing, but it still never says that turning the switch OFF leaves an open scripted preview scripted. That is the direction where silence costs the reader something: they are trying to revoke a permission and the screen implies it took effect. Low reachability (the Settings screen and the Gram preview cover are not co-presented on iPhone), which is why this stays P3.
  - file: App/HerdrApp.swift
  - line: 7156
  - rationale: The finding named an asymmetry between the two strings; both strings were rewritten this round and the asymmetry was not addressed, so the observation is unchanged.
  - state: answered
- **Island glass card contradicts GlassSurface's own doc comment and is not the iOS idiom it claims** (P3)
  - continues_uid: jerryfane/herdrup#254-f6
  - detail: Checked at this head and NOT fixed -- unchanged. GlassSurface's doc still says it is for surfaces where 'the backdrop is the wallpaper or the island -- never over an opaque fill' (:211-214), and ExpandedBody still wraps the island's bottom region in GlassSurface(cornerRadius: 16) (:167) with the justification 'so the island reads like the rest of iOS rather than text floating on black' (:131-132). The Dynamic Island's own backdrop is the opaque black pill, so .ultraThinMaterial there blurs an opaque fill and reduces to the wash plus sheen -- which is the case the doc comment tells the reader not to do. Nothing in the delta touches GlassSurface or the ExpandedBody background. Unverifiable in any test target, and cosmetic, so it stays P3.
  - line: 211
  - rationale: Recorded as still open at this head so the ledger row is answered; the code the finding cites is identical to the head it was raised against.
  - state: answered
- **GlassSurface(cornerRadius: 0) strokes a square border inside the system's rounded container** (P3)
  - continues_uid: jerryfane/herdrup#254-f7
  - detail: Checked at this head and NOT fixed -- unchanged. LockScreenView's non-Always-On background is still GlassSurface(cornerRadius: 0) (:286), and GlassSurface still draws a .strokeBorder(glassEdge, lineWidth: 0.75) on a RoundedRectangle built from that radius (:226-227), so the hairline is a square path inside the system's rounded Live Activity container and its corners are clipped away while its straight runs sit inboard of the actual curve. Not touched by the delta. Cosmetic and unobservable from any test target, so P3.
  - line: 286
  - rationale: Answered for the ledger at this head; the cited construction is byte-identical to the head where it was raised.
  - state: answered
- **Non-text and accent colours on the lock-screen glass still assume the opaque backdrop that f1's fix moved away from** (P3)
  - continues_uid: jerryfane/herdrup#254-f10
  - detail: Checked at this head and NOT fixed -- unchanged. f1's fix lifted only the two text colours; the non-text ink on the glass path still comes from the opaque-navy palette. On the lock screen over an unknown wallpaper: WidgetPalette.waiting #E9A63C paints both the 34pt hero count (:312) and, via markColor, the summary line at :255 and the action tint (:269); WidgetPalette.textFaint #666D91 is the idle StatusMark fill (:406) and the idle mark colour (:470); and glassEdge is a flat white 0.14 (:442). None of these were re-tuned for the 0.80 wash, and the idle mark in particular is the same value f1 measured at 1.1:1 on this surface. The delta does not touch WidgetPalette or any of these call sites.
  - line: 462
  - rationale: Recorded as still open at this head so the row is answered; the palette and every glass-path consumer of it are identical to the head where the finding was raised.
  - state: answered
- **ExpandedMark's new comment asserts one spelling across presentations, but the lock screen still prints the raw count** (P3)
  - detail: NEW at this head, arising from f8's fix rather than from the old code. The comment added at :97-100 states the invariant as cross-presentation: 'The SAME spelling the compact pill uses. Two presentations of one activity printing 100 here and 99+ there is the kind of detail that reads as a bug.' That now holds between CompactCount (:69/:74) and ExpandedMark (:101), but the SAME activity has two further presentations that were not changed and are not mentioned: LockScreenView.lockHero still renders Text(verbatim: \"\\(state.needsYouCount)\") at :309, and the line beneath it comes from AgentActivitySummary.line, which interpolates the raw count ('\\(s.needsYouCount) need you', Shared/AgentActivityState.swift:190/193/195) and is consumed at :253. I computed the divergence: at 100 waiting the compact pill and island corner say '99+' while the lock screen says '100' and '100 need you'; the four agree only below 100. The likely correct resolution is the comment, not the code -- the hero is a 34pt display number with .fixedSize(horizontal: true) (:318) and abbreviating it to '99+' would be a worse read, and the summary line is a sentence rather than a badge -- but as written the file asserts a consistency rule it does not keep, which is the same class of stale-rationale comment the codebase elsewhere goes out of its way to prevent (see the NO HEADLINE OVERRIDE note at :483-490).
  - line: 309
  - rationale: A comment that states an invariant more broadly than the code enforces it misleads the next reader into assuming the other presentations were considered; the cheap fix is to scope the comment to the compact pill and the corner, which is where the bound ('99+ is the widest string it can ever have to fit') actually does work.
- **'The block is absolute while script is off' is the one surviving absolute in the reworded policy doc** (P3)
  - detail: NEW, and the smallest residue of the f3 class. WebViewPolicy.swift:12-14 now reads: 'a content rule matches URL loads, so it stops fetch, XHR, beacons and subresources -- everything a script-less document can attempt, which is why the block is absolute while script is off.' The enumerated four are genuinely covered, and the navigation routes are separately cancelled by the coordinator, so the sentence is right for every channel I can name that goes through the resource loader. The claim that survives verification is the parenthetical one: 'everything a script-less document can attempt'. A script-less document can also carry <link rel=\"dns-prefetch\"> or <link rel=\"preconnect\">, which are speculative DNS/connection operations rather than resource loads; whether WebKit routes those through content extensions is WebKit-internal behaviour I cannot establish in this seat (no Swift toolchain, no simulator, and the Linux test only matches url-filter strings, never WebKit's loader). If they are not filtered, a DNS query to an attacker-controlled authoritative name server is exactly the read receipt the doc attributes only to the script-ON case, and the word 'absolute' would be wrong in the state the product treats as safe. STATED UNCERTAINTY: I could not test this and I am not asserting the hole exists. Note the user-facing strings are unaffected -- HerdrApp.swift:7156 says 'it can't load anything from the network', and a preconnect is not a load -- which is why this is P3 and not a continuation of f3's P2.
  - file: Sources/HerdrKit/WebViewPolicy.swift
  - line: 13
  - rationale: The rest of the rewrite succeeds precisely by replacing absolutes with a named mechanism and a named exception; hedging this one clause the same way (for example 'everything a script-less document can load') would cost nothing and would remove the last sentence in the file whose truth depends on unobservable WebKit behaviour.

**Tests Run**
- git rev-parse HEAD -> d1e998357c45f2210f9e56f1dc2a8a0d8a23fa1f, matching the requested head; git log --oneline -3 -> d1e9983 'Say what the preview actually guarantees, and print one spelling of the count' over 4d00183 over eff9c9d; git status --porcelain clean (EXECUTED)
- git diff --stat 4d00183 d1e9983 -> 5 files, +52/-27 (App/GramView.swift, App/HerdrApp.swift, App/HtmlWebView.swift, HerdrWidgets/AgentLiveActivity.swift, Sources/HerdrKit/WebViewPolicy.swift); git diff --stat 3504d56 d1e9983 -> 7 files, +363/-112; full text diff of the delta read (EXECUTED)
- Python re-implementation of WebKit's case-insensitive url-filter matching over the five rules parsed out of WebViewPolicy.blockNetworkRuleListJSON at this head: 12/12 must-block URLs blocked (http, https, ws, wss, ftp, ftps, file:///etc/passwd, blob:abcdef, HTTP://H/x, FILE:///etc/passwd, http://1.2.3.4:8080/p, https://h/x?d=stolen), 0/3 must-allow URLs blocked (about:blank, data:text/html,hi, data:image/png;base64,AAAA) (EXECUTED)
- Python arithmetic on the count spelling across all four presentations at 1/9/99/100/250: compact pill == island corner at every value; lock-screen hero and AgentActivitySummary.line diverge at >= 100 (EXECUTED)
- Python arithmetic on ExpandedMark's worst-case intrinsic width: 4pt leading padding + 10pt StatusMark + 6pt spacing + 30.6pt for '99+' at 17pt IBM Plex Mono (0.6em advance) = 50.6pt ideal; ~1.0 characters fit if the corner is the ~3-character budget the file's doc comment claims (EXECUTED)
- grep over App/, Sources/, Shared/, HerdrWidgets/ and the test targets for javaScriptEnabled|previewJavaScript|allowsJavaScript|javaScriptDefaultsKey -> production reads are @AppStorage only (GramView.swift:159 -> :366, HerdrApp.swift:7134 -> :7146); WebViewPolicy.javaScriptEnabled(in:) referenced only by WebViewPolicyTests and its own doc comment (EXECUTED)
- Full reads at this head of Sources/HerdrKit/WebViewPolicy.swift (all 62 lines), App/HtmlWebView.swift (all 130 incl. makeUIView and Coordinator), HerdrWidgets/AgentLiveActivity.swift (all 509), Tests/HerdrKitTests/WebViewPolicyTests.swift, the new HerdrUITests/SettingsTests.swift, App/HerdrApp.swift:6050-6115 and :7129-7167, Shared/AgentActivityState.swift count decls (EXECUTED)
- Cross-checked the cited UI test against the reworded strings: SettingsTests asserts label CONTAINS[c] 'will run' and 'are ignored'; HerdrApp.swift:7155 contains 'Scripts in a previewed file will run' and :7156 contains 'Scripts in a previewed file are ignored', so the assertions survive the rewording (EXECUTED, static string comparison)
- NOT RUN, no toolchain: `which swift swiftc xcodebuild` returns nothing and `gh auth status` reports no login, so SettingsTests iPhone run 34902493632, the app/widget-extension compiles in that job, and `swift test` WebViewPolicyTests 3/3 could neither be re-run nor fetched; taken as the sender's evidence (NOT EXECUTED)
- NOT VERIFIABLE by any means available: everything the Live Activity renders (Dynamic Island expanded regions, compact pill, lock-screen widget), as the sender concedes; the contrast and corner-width judgements above are arithmetic over the declared colours and font metrics, not observation (NOT EXECUTED)

PR #255, opened 2026-09-15
Prove a previewed script runs only when the switch is on

jerryfane:
Follow-up to #254. That PR shipped the JavaScript switch with two receipts: the default is off (Linux, `WebViewPolicyTests`) and the Settings row's wording tracks the switch (`SettingsTests`). Neither proves the thing that matters — that a script in a previewed document actually does not run.

## What this adds

- A DEBUG-only harness, `ScreenshotMock.htmlPreview` / `HtmlPreviewHarness`, that renders `HtmlWebView` over a document whose `<script>` rewrites the page from "script did not run" to "script ran". It reads the same `@AppStorage(WebViewPolicy.javaScriptDefaultsKey)` GramView reads and passes it to the same view, so the receipt covers the shipping wiring rather than a stand-in.
- Two tests that read the rendered DOM text out of the web view:
  - `testAPreviewedScriptDoesNotRunWhenTheSwitchIsOff` — the page still says the script did not run.
  - `testAPreviewedScriptRunsWhenTheSwitchIsOn` — the same document rewrites itself.

The second direction is what makes the first one worth having: without it, a viewer that ignored the setting entirely would pass.

Each test forces the value through the argument domain rather than inheriting whatever a previous run on that simulator stored, so neither depends on order. The DEFAULT — an absent key — stays pinned on Linux, where a fresh `UserDefaults` suite can actually be created.

No production behaviour changes: the harness is `#if DEBUG` and nothing else moved.

## Receipt

Focused SettingsTests iPhone run 34911066465 SUCCESS: both new tests green, plus the existing wording test.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d556bbf68af510-1`

**Decision:** `approved`
**Summary:** Test-only follow-up that genuinely closes #254's receipt gap: the DEBUG harness renders the same HtmlWebView with the same @AppStorage(WebViewPolicy.javaScriptDefaultsKey) read that GramView uses, so WKWebView script execution is actually exercised in both directions. #if DEBUG placement is correct (RootView's case at HerdrApp.swift:363 is inside the :312-:422 DEBUG block; the harness and ScreenshotMock are inside the file-level :7603-:8299 DEBUG block), so release builds carry neither. The negative test cannot pass by the page failing to load, because it first asserts the unexecuted text is present and the fail-closed placeholder renders different copy. The argument-domain override is volatile and non-persisting, so it isolates each direction without polluting simulator state or the sibling copy test. One substantive finding, pre-existing from #254 rather than introduced here: the comment claiming the real default is pinned on Linux is inaccurate for this path, because WebViewPolicy.javaScriptEnabled(in:) is called only by tests while the app reads three separate '= false' @AppStorage literals. Approving; the finding is follow-up debt, not a defect in this change. Could not execute the cited iPhone run 34911066465 or the Linux suite: no Swift/Xcode toolchain in this sandbox and gh is unauthenticated.
**Severity:** `P2`

**Findings**
- **Comment claims the shipping default is pinned on Linux, but the pinned helper is dead code** (P2)
  - detail: SettingsTests.swift:4-7 and :51-54 justify forcing the value via launchArguments by asserting the absent-key default is covered by WebViewPolicyTests.testJavaScriptIsOffUntilTheSettingIsTurnedOn. That test exercises WebViewPolicy.javaScriptEnabled(in:) (Sources/HerdrKit/WebViewPolicy.swift:59), whose only references in the repo are Tests/HerdrKitTests/WebViewPolicyTests.swift:33, :37 and :41. No application code calls it. The shipping default is instead three independent literals: App/GramView.swift:159, App/HerdrApp.swift:7143 and App/HerdrApp.swift:7679, each '@AppStorage(WebViewPolicy.javaScriptDefaultsKey) private var previewJavaScript = false'. Changing App/GramView.swift:159 to 'true' would make script execution the default for received HTML/SVG while testJavaScriptIsOffUntilTheSettingIsTurnedOn, testAPreviewedScriptDoesNotRunWhenTheSwitchIsOff (forces NO) and testAPreviewedScriptRunsWhenTheSwitchIsOn (forces YES) all still pass. The argument-domain override itself is the right isolation; the gap is that the compensating receipt it points at does not cover the read the app performs. Pre-existing #254 debt, not introduced by this PR. Suggested follow-up: give the views a single shared default (e.g. have them read WebViewPolicy.javaScriptEnabled, or seed @AppStorage from one WebViewPolicy constant) so the Linux test becomes load-bearing for the shipping path.
  - file: HerdrUITests/SettingsTests.swift
  - line: 53
- **Harness introduces a third unshared copy of the default literal** (P3)
  - detail: HtmlPreviewHarness declares its own '@AppStorage(WebViewPolicy.javaScriptDefaultsKey) private var previewJavaScript = false', a third duplicate alongside App/GramView.swift:159 and App/HerdrApp.swift:7143. Because both new UI tests force the value through the argument domain, the harness's literal is never observed by any assertion, so it can silently drift from GramView's. Harmless today and immaterial to what the tests prove, but it widens the surface described in the P2 finding.
  - file: App/HerdrApp.swift
  - line: 7679
- **Harness document omits the viewport meta that the sibling inline document carries** (P3)
  - detail: The harness document declares only '<meta charset="utf-8">', whereas HtmlWebView.blockedPlaceholderHTML (App/HtmlWebView.swift:96) sets '<meta name="viewport" content="width=device-width,initial-scale=1">'. Without it WebKit lays out at the 980px fallback viewport and scales down, so the paragraph renders very small. XCUITest's .exists on app.webViews.staticTexts reads the accessibility tree rather than hit-testing, so this should not affect the assertions, but adding the meta would make the rendered receipt legible in failure screenshots and remove a needless difference from the other inline document in the codebase.
  - file: App/HerdrApp.swift
  - line: 7684

**Tests Run**
- git diff --check 708fceb..f528c10 - clean, exit 0 (run here)
- git log/diff 708fceb..f528c10 - confirmed head f528c10, 2 files, +67/-1, matches the described scope (run here)
- Preprocessor balance audit of App/HerdrApp.swift #if DEBUG/#endif pairs - confirmed :7603 closes at :8299 enclosing ScreenshotMock and HtmlPreviewHarness, and :312-:422 encloses the new RootView case (run here)
- Cross-reference of WebViewPolicy.javaScriptEnabled and javaScriptDefaultsKey call sites across Sources/, App/, Tests/ and HerdrUITests/ (run here)
- NOT RUN: xcodebuild focused SettingsTests iPhone run 34911066465 - no Xcode toolchain in this sandbox
- NOT RUN: swift build and HerdrKitTests/WebViewPolicyTests on Linux - 'swift: command not found' in this sandbox
- NOT RUN: gh run view 34911066465 - gh present but unauthenticated (GH_TOKEN unset), so the cited green run could not be independently confirmed

PR #257, opened 2026-09-15
Let the system own the lock screen surface, put the island body back, and show download progress

jerryfane:
Owner feedback on TestFlight build 145, with screenshots: the lock screen "just got darker, so uglier", and the expanded island "is still buggier, it was better before". Plus an unrelated ask: a progress bar while DOWNLOADING gram files.

## 1. Lock screen — the material was on the wrong side of the glass

A material drawn INSIDE a Live Activity does not sample the wallpaper. iOS composites its own surface behind the card, so `.ultraThinMaterial` blurred that, and every wash over it only darkened the panel: 0.80 read as paint in build 144, 0.55 came back darker in 145. The in-app gallery disagreed because there a material really does sample what is behind it — the one thing that harness cannot stand in for, and it misled two rounds.

The card now draws **no background at all**. With `activityBackgroundTint(nil)` the background IS the system's own translucent material.

That forced a second change, found by measuring the render rather than shipping it: with the surface ours no longer, its brightness ranges from `#1D1F2E` over a dark wallpaper to `#A3A3A3` over a white one. No fixed ink survives that — near-white measures 2.2:1 at the light end, a dark ink 2.6:1 at the dark end. So lock-screen text now asks the system: `.primary` / `.secondary` / `.tertiary`, which get vibrancy. The hero number goes with them; amber measured **1.4:1** over a white wallpaper, and the number is the hero by size, with the mark beside it carrying hue and shape.

Always-On is untouched: that panel paints its own opaque navy, so the tuned tokens stay.

## 2. Expanded island — the spec's region map lost on device

The design splits the expanded view four ways, with a three-line fleet column trailing. On a real island the corners are narrower than the design assumed: the owner's screenshot shows a mark, a name, and a trailing column collapsed to ellipses.

Content is back in the full-width bottom region, one short thing per corner — the arrangement the owner judged better. Type sizes, weights and tokens stay the spec's; only which region holds what moved. The gallery now frames the corners at **84 pt, tighter than the device**, so a layout that only survives a generous mock fails in the receipt first.

## 3. Gram download progress bar

A gram file arrives as one base64 reply, so there was no progress signal at all — only a spinner. `CitadelTransport.roundTrip(_:onBytesReceived:)` now reports SSH channel bytes as they land, `gramGetFile` forwards them, and both row types (`GramRow`, `SavedGramRow`) show a real bar plus "3.2 MB of 8.5 MB", scaled back from base64 so the figures name file bytes. A row with no size keeps the indeterminate spinner rather than drawing a bar nothing drives.

## Receipts

- `WidgetGalleryTests` iPhone run 34971826820 green; frames inspected: at 84 pt corners the island renders the host label, headline, counts line and full-width button with **no ellipsis collapse** — the build-145 failure reproduced and fixed.
- Linux `swift build` + suite green (the one failure, `LiveRecoveryExecutorTests`, is gated on a live herdr socket and skips in CI).

## Not verified

Vibrancy. The gallery cannot reproduce it, so the lock screen's contrast is now the system's guarantee rather than a number measured here — stated in the file as well as here. And the island's outer geometry, which iOS composes.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5815542759c42-1`

**Summary:** PR #257 at 90d181bc does not compile: commit 90d181b deleted WidgetPalette.islandTextFaint while two live references remain (AgentLiveActivity.swift:141, :231), and project.yml compiles that file into BOTH the app target (line 82) and the HerdrWidgets extension (line 308), so both fail. The green gallery run 34971826820 necessarily predates 90d181b — it is the only commit touching that declaration — so the head run cannot build. Beyond the break: the lock screen still carries fixed ink on the surface it no longer owns (the Open label at 2.2:1 and, worse, StatusMark at 1.14-2.00:1 over the PR's own measured #A3A3A3), which undercuts the stated justification for dropping the amber hero; the documented gramGetFile fallback contract is violated by both of its callers, which arm a determinate bar from file size alone and leave it frozen at 0% on MockTransport/UnixSocketTransport; the saved-row bar resolves totals from the inbox only and is therefore dead for saved grams whose source message is gone; and the per-chunk MainActor hop is unordered, unthrottled and untagged. AnyShapeStyle is the right vehicle and the Always-On branch is complete apart from StatusMark. Island reversal is sound on overflow: I found no state that ellipsizes the bottom region, and no dangling references to the removed types.
**Severity:** `P0`

**Findings**
- **Widget + app targets do not compile: WidgetPalette.islandTextFaint deleted, two references remain** (P0)
  - detail: `git show 90d181b -- HerdrWidgets/AgentLiveActivity.swift | grep islandTextFaint` yields exactly one line: `-    static let islandTextFaint = Color(hex6: 0x767DA3)`. A repo-wide walk (every file, not just *.swift) finds the identifier in exactly two places, both uses, no declaration: AgentLiveActivity.swift:141 (`ExpandedHost.body`) and :231 (`ExpandedBody.age`). The file declares `private enum WidgetPalette` once (offset 22282) and has no `extension WidgetPalette`, so this is a plain 'type WidgetPalette has no member islandTextFaint'. project.yml puts the file in the app target (line 82, `- path: HerdrWidgets/AgentLiveActivity.swift`) and the whole dir in the extension (line 308), so both targets fail and no UI-test run at this head can build. Both surviving uses are on the island, which is drawn on true black — a fixed token is correct there, so restoring the 0x767DA3 constant is the whole fix. This also means the gallery evidence cited for change (2) was captured before the commit that broke the build.
- **Fixed ink is still on the surface you no longer own, and the mark that inherits the hero's meaning is the worst case** (P2)
  - detail: Answering your own question: no, it is not clean. Contrast computed against the PR's measured endpoints (#A3A3A3 light / #1D1F2E dark):
- ActivityAction label, line 441, `WidgetPalette.text` #EEF0F7 → 2.22:1 light. The border was moved to `.tertiary` (line 446) but the 15 pt SemiBold label it outlines was not; 2.2:1 is the exact number cited in the PR body as disqualifying near-white.
- `.activitySystemActionForegroundColor(WidgetPalette.text)`, line 20 → same 2.22:1 on the system-drawn control. Passing nil lets the system pick.
- StatusMark, line 460: `onGlass` was deleted, so the mark is now `WidgetPalette.color(status)` on both surfaces. Light end: needsYou #E9A63C 1.20:1, working #5B9BE8 1.14:1, stopped #E2584E 1.45:1, idle #666D91 2.00:1. All under the 3:1 graphic bar; the working ring is effectively invisible over a white wallpaper.
That last one is the substantive objection to change (1). Dropping the amber hero is defensible on its own (amber #F2B85C measures 1.42:1 over #A3A3A3 — worse than your quoted 1.4:1 for the token, so the call is right), but the stated reason is 'the mark beside it carrying the colour and the shape', and the reversal made that mark strictly less visible than the text it replaced, at 1.20:1. Restoring the deleted glass tints is not the fix either (they were tuned for #595B67, and #F2B85C is 1.42:1 on the new surface). The mark needs a surface it owns — a small opaque/`.regularMaterial` chip behind it, or a `.primary`-inked glyph — if colour is to keep carrying status.
- **gramGetFile's fallback contract is documented and then violated by both callers** (P2)
  - detail: HerdrClient.swift's new doc is explicit: a non-Citadel transport 'gets the plain path and no callbacks, which a caller must treat as "no progress available" rather than "no bytes yet"'. Neither caller does. GramView.openFile (:1706) and saveFile (:1774) set `downloadBytes = Self.expectedBytes(...).map { (0, $0) }` from the message's file size alone, before any evidence the transport can report, and both rows gate the bar on `p.total > 0` only (GramView.swift:2194, SavedGramStore.swift:154). So on any non-Citadel transport the row draws a determinate bar pinned at 0% for the whole download — precisely the 'bar that is not driven by anything' the comments claim to avoid. This is reachable in the shipped binary, not just in tests: MockTransport answers `gram.get_file` (HerdrApp.swift:7794) and its canned `gram.list` carries `file.size: 48213` (:8012), driving the DEBUG Gram route at :418; `UnixSocketTransport` (Transport.swift:179) is the other non-Citadel conformer. Make the capability explicit — a `HerdrTransport` member with a no-op default, or arm `downloadBytes` only on the first callback rather than before the call — so silence renders the old spinner as documented.
- **The saved row's new bar is dead for exactly the rows it was added for** (P2)
  - detail: `expectedBytes(for:in:)` (GramView.swift:1804) resolves totals from `inboxStore.inbox.messages` only, but `savedContent` is explicitly independent of the inbox: 'Saved is local and complete' (:720) and the store exists so 'a saved message survives even after the original is deleted from the server' (:699-700). When the source message is deleted or simply not in the loaded set, the lookup returns nil, `downloadBytes` stays nil and SavedGramRow keeps its spinner — while `SavedGramFile.size` (SavedGramStore.swift:17) sits unused on the row that is already rendering `file.displaySize`. Commit 579ed07 therefore ships a bar that cannot appear in the saved tab's characteristic case. Derive the total from `saved.file.size` on that path.
- **Per-chunk MainActor hop: unordered, unthrottled, and not tagged to a download** (P2)
  - detail: Three separate problems in the same closure (GramView.swift:1710, :1778):
- Ordering: each stdout chunk spawns a fresh `Task { @MainActor in }`. Enqueue order onto the main actor is not FIFO-guaranteed and nothing clamps monotonically, so both the bar and the 'X of Y' label can move backwards. `downloadBytes = (max(previous, min(received, total)), total)` fixes it.
- Throughput, your question: a 40 MB file is ~55.9 MB of base64; at nio-ssh's ~32 KB channel chunks that is ~1.7k hops, each writing `@State` on GramView and re-evaluating the entire page body including the message LazyVStack. The download itself stays off the main thread, but hundreds of full-body invalidations per second while the list scrolls is real churn for no added information. Coalesce on a delta threshold (≥1% or ≥100 ms).
- Identity: the callback carries no download id and the only guard is `downloadBytes?.total != nil`. A callback still in flight when `openFileTask` is cancelled in `.onDisappear` lands after `defer` clears state, but if the owner starts a second download first it writes into that download's `(0, total)` and jumps its bar.
- **Expected-size estimate: the tail stall is the real artifact, not an early finish** (P3)
  - detail: For the sizes that matter the estimate under-shoots rather than pins. `floor(4n/3) + 512` vs actual `4·ceil(n/3) + envelope`: at 40 MB that is 55,924,565 expected against 55,924,056 + ~150 B of `{"id":…,"result":{"type":"gram_file_content","name":…,"mime":…,"size":…,"data_base64":…}}`, so the bar tops out just short of 100% and `min(received, total)` is a backstop, not the normal path. It can pin early only if the reply gains fields or carries a name+mime over ~360 B. What the owner will actually see is at the other end: the last chunk puts the bar at 100%, and then base64-decoding 55.9 MB and writing the temp file happen with no signal at all, which reads as a stall at full. Worth a distinct 'decoding' state. Related and pre-existing, but sharpened by the PR's own 40 MB framing: that path holds LineAccumulator's raw bytes, the decoded String, the decoder's copy of the base64 and the output Data concurrently — roughly 4x the file size resident.
- **No receipt for the only new HerdrKit logic** (P3)
  - detail: `git diff --name-only ac2d03f..HEAD -- Tests` is empty. The byte counter is the one piece of new non-UI behaviour and it is already trivially testable in the existing harness: CitadelTransportTests.swift drives `parseBridgeOutput` with hand-built `AsyncThrowingStream`s at :305, :324, :341, :349, and the function's own doc comment advertises exactly this. Asserting that a two-chunk stdout stream reports cumulative, non-decreasing counts and that the final count equals the byte total is a few lines and defends an observable contract (the monotonic guarantee the callers currently lack).
- **Expanded island: stale state now shows no fleet counts anywhere** (P3)
  - detail: `ExpandedBody.age` (line 207) puts `"N working · M agents"` inside the non-stale `else` branch only, and the trailing corner now carries just the host label. Build 144's `FleetTotals` printed host + working + total in every state, so the arrangement being restored has quietly lost the counts precisely when the card is least trustworthy. On the overflow question I could not construct a failing state: headline and question are lineLimit(1) and truncate by design, and the faint line's worst realistic content ('waiting 1:02:33 · 12 working · 40 agents', ~40 chars at Plex 11 ≈ 265 pt) fits the ~330 pt bottom region. The one residual risk the gallery cannot exercise is `Text(style: .timer)` width reservation squeezing its HStack siblings — the 84 pt corners don't test it because the bottom region is full-width in the mock, so the outer geometry stays unverified as you noted.
- **Dead constant and a comment that now states the opposite of the code** (P3)
  - detail: A census of `WidgetPalette.` uses in the file shows `ground` referenced nowhere after `glassWash` was deleted. And the `activityBackgroundTint(nil)` comment at lines 16-18 still reads 'The view paints its own blur + brand sheen on top of that (see LockScreenView.body)' — the exact claim this PR removes; the body now draws nothing outside Always-On. In a file where the comments are the argument, that one misleads the next reader straight into build 144's mistake. Everything else is clean: `GlassSurface`, `FleetTotals`, `ExpandedHeadline`, `ExpandedAction`, `onGlass`, `washAlpha`, `glassColor`, `glassTextDim/Faint`, `glassWashAlpha` and the 'wash 55' caption have zero remaining references repo-wide, and the test's new expectation matches the gallery's `"\(item.0) · \(backdrop.0)"`.
- **Judgement on AnyShapeStyle and the Always-On branch (no change requested)** (P3)
  - detail: AnyShapeStyle is the right vehicle. A ternary needs a single type and `Color` does not unify with `HierarchicalShapeStyle`; the alternative is duplicating the whole card under `if isLuminanceReduced`, which is strictly worse. The cost is one existential box per accessor per body evaluation, irrelevant at Live Activity refresh rates. The Always-On branch is complete: `primaryInk`/`secondaryInk`/`tertiaryInk`/`heroInk` all return the opaque tokens under `isLuminanceReduced`, the `WidgetPalette.backdrop` fill is retained, and `ActivityAction` receives `onSystemSurface: !isLuminanceReduced` so it takes `textFaint`. The single exception is StatusMark, which lost its branch entirely — correct for Always-On, and the P2 above on the system surface. The vibrancy gap you flagged as unverified is genuinely unverifiable here, and `.primary/.secondary/.tertiary` is the right answer to it; the reasoning only breaks where you kept a fixed colour anyway.

**Tests Run**
- git rev-parse HEAD → 90d181bc437066fc2162d33cf5addfb98c09f73c; git diff --stat ac2d03f..HEAD (6 files, +297/-241); per-commit attribution of the deletion via `git show 90d181b -- HerdrWidgets/AgentLiveActivity.swift | grep islandTextFaint`
- repo-wide identifier walk (python os.walk, all files) for islandTextFaint → 2 hits, both call sites, zero declarations; regex count of `enum|extension WidgetPalette` in the file → 1
- grep census for the removed symbols (glassWashAlpha, glassWash, glassColor, glassTextDim, glassTextFaint, GlassSurface, onGlass, washAlpha, 'wash 55', ExpandedHeadline, ExpandedAction, FleetTotals) across *.swift/*.md/*.yml/*.pbxproj → no dangling references
- project.yml target-membership check: AgentLiveActivity.swift in app target (line 82) and HerdrWidgets extension (line 308); SWIFT_VERSION 5.9, no SWIFT_STRICT_CONCURRENCY (so the non-Sendable callback crossing into the actor is diagnostic-free today, Swift 6 hazard only)
- WCAG 2.x relative-luminance contrast computed in-session for #EEF0F7, #E9A63C, #5B9BE8, #E2584E, #666D91, #F2B85C, #99A0BC against the PR's measured #A3A3A3 and #1D1F2E
- HerdrTransport conformer enumeration (CitadelTransport, UnixSocketTransport, App/HerdrApp.swift MockTransport) and MockTransport gram fixtures (gram.get_file at :7794, gram.list file.size 48213 at :8012, DEBUG route at :418)
- read of HerdrClient.call/encodeRequest/decodeResult to confirm the Citadel bypass path loses nothing (no retry, reconnect or error mapping lives in `call`)
- git diff --name-only ac2d03f..HEAD -- Tests → empty; located the existing parseBridgeOutput stream harness at CitadelTransportTests.swift:305-349
- NOT RUN, hence static_only: `swift build` → 'command not found'; no swiftc/xcodebuild in this environment; `gh run view 34971826820 / 34974454176` refused (no GH_TOKEN), so the CI run head SHAs could not be confirmed directly — the ordering claim rests on git history alone

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5855a193a5d92-1`

**Decision:** `approved`
**Summary:** Re-reviewed at db7b3dd. No Swift toolchain in this sandbox and `gh` is unauthenticated, so the CI claim for run 34980190311 is unverified by me; I substituted a static symbol audit that covers both classes of break the author names. Every WidgetPalette member referenced in AgentLiveActivity.swift is defined (islandTextFaint restored at line 553), and every call site's argument labels match its declaration - onSystemSurface exists on both StatusMark (line 476) and ActivityAction (line 438), and the six StatusMark / two ActivityAction / two ExpandedBody call sites pass only declared labels. Repo-wide grep for the deleted glass symbols (onGlass, glassColor, glassText*, GlassSurface, washAlpha, glassWash, FleetTotals, ExpandedHeadline, ExpandedAction) returns zero hits across App, HerdrWidgets, Sources, Shared, Tests, AgentActivityStateTests, HerdrUITests. I recomputed every contrast figure the new comments quote: over ground #13162A amber 8.49, working 6.21, died 4.88, idle 3.54; on black islandTextFaint 5.23, textDim 8.11, textFaint 4.16 - all as stated. Judgements asked for: the disc is the right answer, not a .primary glyph (status here is hue AND shape, and with the hero number now .primary the mark is the only colour channel left; a mono glyph would collapse four states to one ink). died 4.88 and idle 3.54 both clear the 3:1 non-text bar for a 12pt graphic, and idle is the least load-bearing state, so adequate. The 1% threshold cannot leave the bar visibly short - worst case is 1% of 56MB retired at 99% - but the comment's promise that the final chunk always lands is false, because expectedReplyBytes adds 512 bytes the wire never carries. The id guard closes the cross-id window but not same-id cancel-then-retap. Four P3s, none blocking.
**Severity:** `P3`

**Findings**
- **Both compile breaks closed by static symbol audit; CI claim itself unverified here** (P3)
  - continues_uid: jerryfane/herdrup#257-f1
  - detail: Checked at this head, not taken on trust. (a) WidgetPalette.islandTextFaint is defined again at line 553 and is referenced at lines 143 and 233 - the only two references. (b) The second break you introduced at a771d16 is gone: onSystemSurface is declared on StatusMark (line 476) and on ActivityAction (line 438), and I enumerated every construction in the file - StatusMark at 52, 61, 107, 369, 393, 753, 764; ActivityAction at 162 and 291; ExpandedHost at 41, 731; ExpandedBody at 45, 735; LockScreenView at 8, 703 - and each passes only labels its type declares (washAlpha is gone from both LockScreenView call sites). I also cross-checked all nine WidgetPalette.* references against the statics the enum defines: none missing. What I could NOT do: run swiftc or xcodebuild (no Swift toolchain on this Linux box) or fetch run 34980190311 (gh has no token). So this is a symbol-resolution audit plus your CI report, not an independent compile. Downgraded to P3 as an observation only.
  - line: 553
  - state: answered
- **Disc is the correct answer for the mark; the system-ink label and nil system action colour are right** (P3)
  - continues_uid: jerryfane/herdrup#257-f2
  - detail: activitySystemActionForegroundColor(nil) at line 21 and ActivityAction's label taking AnyShapeStyle(.primary) when onSystemSurface (line 447) remove the last fixed ink from the surface you do not own; the Always-On branch still takes WidgetPalette.text, which is correct because that panel paints its own navy. On the mark: the disc beats a .primary-inked glyph. Status is encoded here as hue AND shape (filled circle / ring / small dot / dashed circle), and once the hero number became .primary the mark is the ONLY carrier of hue on the card; inking it .primary would flatten needsYou, working, idle and stopped to one colour and leave shape alone doing work at 12pt. The disc also puts the hue back on a surface this file owns - I verified the tokens over #13162A: needsYou 8.49, working 6.21, stopped 4.88, idle 3.54, and ground itself is 7.08:1 against the measured #A3A3A3 so the disc reads as a deliberate shape rather than mud. Adequacy at 12pt: all four clear the 3:1 graphical-object bar; died at 4.88 is comfortable, idle at 3.54 is the marginal one and it is drawn as a 5.04pt inner dot, but idle is the state whose misreading costs least. No change requested.
  - line: 480
  - state: answered
- **Disc padding of diameter*0.42 makes the lock hero's mark 22pt beside a 34pt number, and buys no contrast** (P3)
  - detail: padding(diameter * 0.42) on a 12pt mark yields a 22.08pt disc - 1.84x the mark, and on the light-wallpaper case an opaque near-black blob whose area is comparable to the 34pt digit's cap height sitting next to it. Yes, that is too heavy: the disc is the only opaque thing on a card that is otherwise wholly translucent, so it reads as a punched hole rather than a backing. The contrast argument does not defend the size - the ratios I verified (8.49 / 6.21 / 4.88 / 3.54) depend only on the disc fully backing the mark with a margin past antialiasing, not on its diameter, so 0.20-0.25 (a 17-18pt disc at 12pt, ~14pt at the 10pt marks) keeps every measured figure and stops the mark outweighing the hero. The 0.42 factor appears to have been borrowed from the idle dot's own 0.42 inner-diameter scale, where it means something different. Non-blocking taste call with a measured basis.
  - line: 483
  - state: open
- **gramGetFile fallback contract now holds: the non-Citadel path leaves the spinner** (P3)
  - continues_uid: jerryfane/herdrup#257-f3
  - detail: Traced both callers at this head. downloadBytes is set to nil at entry in openFile (line ~1715) and saveFile (line ~1781) and reset in each task's defer; the only writer is report(), which returns early unless a callback actually arrived AND expected is non-nil. gramGetFile takes the citadel branch only when both onBytesReceived and the as? CitadelTransport cast succeed, so a mock or local-socket transport produces zero callbacks, downloadBytes stays nil, and both row types gate their ProgressView on `let p = downloadProgress, p.total > 0` - indeterminate spinner, exactly as the doc comment now promises. The bar also appears on the first chunk rather than at the first 1% step, because the threshold is skipped when downloadBytes is nil.
  - file: App/GramView.swift
  - line: 180
  - state: answered
- **Saved-row total now comes from the row; the dead-bar case is gone** (P3)
  - continues_uid: jerryfane/herdrup#257-f4
  - detail: Verified both directions. SavedGramRow's call sites pass Self.expectedReplyBytes(for: s.file?.size ?? 0) (lines 734-735) and GramRow's pass message.file?.size, so the size fed to the bar is the same UInt64 each row already renders (SavedGramFile.size in App/SavedGramStore.swift:17, GramFile.size in Sources/HerdrKit/Gram.swift:33) - no inbox lookup, so the saved-with-no-server-copy row that returned nil is no longer reachable. expectedReplyBytes returns nil for size 0, which routes those rows back to the spinner rather than a 0-total bar. The scaled label in both sizeLabel and downloadLabel clamps with min(..., file.size), so the 4/3 estimate cannot print a figure larger than the file.
  - file: App/GramView.swift
  - line: 734
  - state: answered
- **Callback hygiene holds; the residual window is same-id cancel-then-retap, not cross-id** (P3)
  - continues_uid: jerryfane/herdrup#257-f5
  - detail: report(_:of:for:) is @MainActor, guards downloadingFileFor == id, clamps with min(received, total) and writes max(clamped, previous), so a late-landing smaller value cannot run the bar backwards - the three rules are genuinely enforced now. Throttle arithmetic on your 40MB example: expected 55,924,565 bytes, 1% step 559,245 bytes, so ~100 writes instead of ~1.7k, as claimed. The id guard closes the window you asked about ONLY when the ids differ. Same id re-tapped after .onDisappear cancels openFileTask (line 412): the captured closure is still live, its id still equals downloadingFileFor, and because the new download starts from downloadBytes = nil, one stale callback from the first attempt lands first and max() then pins the bar at the old high-water mark until the new transfer catches up. Cosmetic overstatement only - a generation counter alongside the id, or capturing a UUID per download, would close it. Not blocking.
  - file: App/GramView.swift
  - line: 1817
  - state: answered
- **report()'s "the final chunk always lands" is unreachable: expectedReplyBytes adds 512 bytes the wire never carries** (P3)
  - detail: The escape hatch `|| clamped >= total` can never fire. expectedReplyBytes returns Int(size*4/3) + 512, while the reply actually carries 4*ceil(size/3) base64 chars plus a ~100-150 byte JSON envelope; for 40MB that is expected 55,924,565 against ~55,924,166 delivered, i.e. expected exceeds the real total by ~400 bytes forever. So the last sub-1% step is always dropped and the bar retires at between 99% and 100% of its track rather than full. Visually that is under a pixel of a 2pt bar and the row disappears immediately after, so the artifact is negligible - but the comment states a guarantee the code does not provide, and the next person to tune the threshold will trust it. Either drop the +512 slack (min() already clamps overshoot) or make the guard a fraction, e.g. clamped >= total - total/100.
  - file: App/GramView.swift
  - line: 1819
  - state: open
- **Tail stall is still the artifact, now bounded and slightly worse at the very end** (P3)
  - continues_uid: jerryfane/herdrup#257-f6
  - detail: Confirmed unchanged in kind at this head. The bar is driven by wire bytes only; after the last chunk arrives the task still base64-decodes result.dataBase64, writes a temp file and presents QuickLook or the export picker, and downloadingFileFor / downloadBytes are cleared only in the defer - so the bar sits parked at ~99-100% for the whole decode-and-write tail on a large file, which is exactly the shape you predicted. The 512-byte overestimate (see the finding above) means it parks a hair short rather than full. No change requested; if it ever matters, the honest fix is a second phase in the label ('preparing') rather than faking the last percent.
  - file: App/GramView.swift
  - line: 1830
  - state: answered
- **Still no receipt for the byte counting, and the harness for one already exists** (P3)
  - continues_uid: jerryfane/herdrup#257-f7
  - detail: Checked: Tests/HerdrKitTests/CitadelTransportTests.swift exercises parseBridgeOutput four times (lines 305, 324, 341, 349) by feeding it a fake AsyncThrowingStream, and grep across Tests and AgentActivityStateTests finds no reference to onBytesReceived. The new logic - `received += buffer.readableBytes; onBytesReceived?(received)` - is cumulative-not-per-chunk and fires before the newline check returns, both of which a single test over the existing fake stream would pin in ten lines (two stdout chunks, assert the collected callbacks are [n1, n1+n2] and that a stderr chunk contributes nothing). This is the only new HerdrKit behaviour in the PR and it is the cheapest test in the diff. Still P3, still worth doing.
  - file: Sources/HerdrKit/CitadelTransport.swift
  - line: 351
  - state: answered
- **Stale expanded island still shows no fleet counts anywhere** (P3)
  - continues_uid: jerryfane/herdrup#257-f8
  - detail: Re-checked at this head and it is unchanged by the region move. ExpandedBody.age puts the counts only in the non-stale branch (line 229); when isStale the HStack prints 'last update <relative>' or 'no recent update' and nothing else, the trailing region now carries only hostLabel (ExpandedHost, line 141), and ExpandedHero suppresses both the number and 'need you' on isStale. So a stale expanded island shows mark + host + summary line + last-update and zero fleet figures, where the old FleetTotals column printed working/total regardless of staleness. Arguably defensible - stale counts are the ones you cannot vouch for - but it is a silent loss and the doc comment on age ('then the fleet counts') does not mention the exception.
  - line: 210
  - state: answered
- **Glass constants and their contradicting comments are fully gone** (P3)
  - continues_uid: jerryfane/herdrup#257-f9
  - detail: Verified by grep across App, HerdrWidgets, Sources, Shared, Tests, AgentActivityStateTests and HerdrUITests: glassWashAlpha, glassWash, glassTextDim, glassTextFaint, glassColor, GlassSurface, washAlpha and onGlass have zero remaining references, and WidgetPalette is down to nine members, all of which are reachable (died is reached through color(_:), the rest directly). The comments that described a wash this file no longer draws are replaced by the measured-on-device note in LockScreenView.body, and the gallery caption no longer advertises an alpha. Closed.
  - line: 533
  - state: answered
- **ExpandedBody.hostLabel is dead: declared, threaded from both call sites, never read** (P3)
  - detail: ExpandedBody declares `let hostLabel: String` and both constructions pass it (line 46 from context.attributes.hostLabel, line 735 from the gallery), but nothing in body, content, headline or age references it - the host now lives exclusively in ExpandedHost in the trailing region. Swift emits no warning for an unread stored property, so this survives as a parameter that looks load-bearing at the call site. Either delete it and the two arguments, or use it (the age line would be the place, and would also fix the stale-branch gap noted on f8).
  - line: 154
  - state: open
- **AnyShapeStyle boxing and the Always-On branch both check out** (P3)
  - continues_uid: jerryfane/herdrup#257-f10
  - detail: The three AnyShapeStyle-returning properties (primaryInk/secondaryInk/tertiaryInk) plus heroInk allocate an existential box per body evaluation; a Live Activity re-renders on state push and on timer ticks in the age line, so the cost is irrelevant and the erasure is the only way to switch between a Color and a HierarchicalShapeStyle in one expression. The Always-On branch is correct: `if isLuminanceReduced { WidgetPalette.backdrop }` paints the opaque navy, every ink property falls back to the tuned tokens, and both StatusMark call sites (lines 369, 393) and the ActivityAction (line 291) pass onSystemSurface: !isLuminanceReduced, so the dimmed card gets no disc and no vibrancy. One doc nit while I was measuring: ActivityAction's comment says textFaint 'measures 3.5:1 there', which is true against flat ground #13162A (3.54) but the Always-On panel is the backdrop gradient, and against its lightest stop #1B1F3A it is 3.19:1 - still over the 3:1 border bar, so the conclusion stands and only the number is generous.
  - line: 338
  - state: answered

**Tests Run**
- git diff ac2d03f..HEAD, per-file, for all six changed files (no toolchain available: swift/swiftc/xcodebuild absent, gh unauthenticated, so no compile and no CI fetch)
- static symbol audit of HerdrWidgets/AgentLiveActivity.swift: all nine WidgetPalette.* references resolved against the enum's statics (none missing); all 16 constructions of StatusMark/ActivityAction/ExpandedHero/ExpandedHost/ExpandedBody/LockScreenView extracted and their argument labels checked against the declarations
- repo-wide grep for the deleted glass symbols (onGlass, glassColor, glassText*, GlassSurface, washAlpha, glassWash, FleetTotals, ExpandedHeadline, ExpandedAction) across App, HerdrWidgets, Sources, Shared, Tests, AgentActivityStateTests, HerdrUITests: zero hits
- WCAG 2.x relative-luminance arithmetic (python) for every token over #13162A, #000000 and the measured #A3A3A3: reproduced amber 8.49, working 6.21, died 4.88, idle 3.54, islandTextFaint 5.23 on black, textDim 8.11, textFaint 4.16; also ground-vs-#A3A3A3 7.08 and textFaint-vs-#1B1F3A 3.19
- disc geometry arithmetic: diameter*0.42 padding on a 12pt mark yields a 22.08pt disc, 1.84x the mark; idle inner dot 5.04pt
- progress-threshold arithmetic for a 40MB file: expectedReplyBytes 55,924,565 vs ~55,924,166 delivered (base64 + ~110B envelope), 1% step 559,245 bytes, ~100 state writes, and `clamped >= total` unreachable
- gallery receipt anchor cross-check: WidgetGalleryTests expects 'needsYou · many · bright', which matches the new caption \(item.0) · \(backdrop.0) with cases[0].0 = 'needsYou · many' and backdrops[0].0 = 'bright'
- grep of Tests/AgentActivityStateTests for parseBridgeOutput/onBytesReceived coverage (four existing fake-stream tests, no byte-count coverage)
- Swift language mode check (Package.swift 5.9, project.yml SWIFT_VERSION 5.9): passing the non-Sendable ((Int) -> Void) callback into the CitadelTransport actor is not an error in this mode

PR #258, opened 2026-09-15
Bring the Live Activity back after the reader swipes it away

jerryfane:
Owner report: "I closed the lockscreen widget before from the lockscreen and now I don't see it anymore." Reproduced from the source; it is a bug, not iOS policy.

## What happens

Swiping a Live Activity away does not hide it — iOS **ends** it. ActivityKit leaves our reference in place with `activityState == .dismissed`, and `update` on a dismissed activity is silently dropped.

`start` guarded on `activity != nil`, so from then on it reported "one is already live", pushed a state nobody would ever see, and returned. `Activity.activities` reclaim on relaunch could adopt a dead one too. The banner stayed gone for the rest of the process's life; the only cure was force-quitting the app.

## The change

`App/LiveActivityController.swift`:

| Before | Now |
|---|---|
| `activity != nil` decided whether one was live | `liveActivity` — the reference only while `activityState == .active`; every decision goes through it |
| Reclaim adopted any activity for this host | Reclaim filters to `.active`, and a dead reference is dropped first |
| Only `start` (connect, `HerdrApp.swift:443`) could mint | `update` mints a replacement when nothing is live, using the host label remembered from the original `start` |
| — | `end()` clears that label, so a session that really disconnected stays gone rather than resurrecting its banner |

The roster hook (`HerdrApp.swift:2430`, `.onChange(of: fullList)`) therefore restores the banner on the next agent state change. Swiping it away is not "stop watching this machine" — the session is live and the app still has something to report.

## Verification

- `WidgetGalleryTests` iPhone run 35008909059 SUCCESS at this head (app + widget compile check and the gallery receipt).
- **Not verified at runtime, and unverifiable here:** that iOS permits minting a replacement shortly after a user dismissal. No test target can see a Live Activity, and `activityState` only transitions on a real device. ActivityKit documents dismissal as ending *that* activity rather than barring further requests, but the device is the only judge. On-device check: swipe the banner away, wait for an agent's state to change, confirm it returns without a force-quit.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5966d7001777c-1`

**Summary:** Reviewed jerryfane/herdrup PR #258 at exact head 8ddcb50857b9656bf16e80852938e29c7a59d5b2 (git rev-parse HEAD matches; git diff 9c74ecb..HEAD = App/LiveActivityController.swift only, +45/-5; tree clean). No prior verdict for #258 exists in GITMOOT_PRIOR_VERDICTS (frozen 2026-09-15T19:50:33Z; it covers 226/228/229/244/246/247/248/249/250/251/252/253/257), so this is round one. EVIDENCE static_only and not by choice: `which swift swiftc xcodebuild` is empty in this seat, so nothing was compiled or executed and the cited WidgetGalleryTests run 35008909059 is taken as the requester's evidence, not mine. What I did run: full-file read at this head, whole-repo call-site grep for start/update/end/pushToken/activityState/staleDate, and Apple-documentation verification of the three ActivityKit semantics the change turns on.

The core diagnosis is right and the central mechanism is right. Routing every liveness decision through one computed property is the correct shape, the dismissed reference genuinely was what made the banner permanent, and I verified the two things you asked me to break and could not: (a) update -> start -> update cannot recurse or double-mint. start calls update only inside `if liveActivity != nil` (:82); the nested update's `guard let activity = liveActivity` (:127) therefore succeeds and terminates at depth 2. Everything is @MainActor and `Activity.request` is the non-async throwing overload (`try`, not `try await`, :97), so there is no suspension point between the liveness check at :82 and the mint at :97 and two mints cannot interleave. (b) hostLabel cannot go stale against a different host. It is written only by start (:81) and cleared only by end (:151); start has exactly one caller, connect (HerdrApp.swift:443), which is reachable only while `credentials == nil` (the ConnectView arm at :277-278), end has exactly one caller, disconnect (:483), and reconnect (:490) reuses the retained credentials and touches neither. The one path that skips the writ
[truncated]
**Severity:** `P2`

**Findings**
- **The .active filter is applied before the orphan sweep, so a still-visible non-active activity is neither adopted nor ended and the new mint duplicates it** (P2)
  - detail: `let existing = Activity<AgentActivityAttributes>.activities.filter { $0.activityState == .active }` (:68-70) feeds BOTH the reclaim selection (:71) and the end-sweep (:73-79). Because the sweep iterates the already-filtered array, a non-active activity can no longer be ended by it. That matters because non-active does not mean off-screen: with the default dismissal policy an `.ended` Live Activity remains on the Lock Screen for up to four hours, and a `.stale` one remains visible until updated. Reachable without any daemon change via the system's own 8h active budget (system ends it, banner persists), and via a server `event: end` or `stale-date` push; our own end() is exempt because it uses dismissalPolicy .immediate (:154). Sequence: >8h session -> system ends the activity, banner still visible and frozen -> next roster change -> update() sees liveActivity == nil (:127) -> start() -> the visible activity is excluded from `existing`, so keep == nil and the sweep never sees it -> Activity.request mints a SECOND banner for the same host. Net effect is a regression in kind: pre-PR that path kept one dead banner, this head shows two, one of them frozen, and nothing clears the orphan until disconnect (end() still sweeps all states). The existing comment at :62-63 ('End every activity we are not keeping (wrong host, or leftover orphans)') describes the property being removed. Fix is two lines and keeps both properties: sweep the unfiltered set, and put the state test inside the selection predicate instead -- `let existing = Activity<AgentActivityAttributes>.activities` then `activity = existing.first { $0.activityState == .active && $0.attributes.hostLabel == hostLabel }`, leaving :72-79 untouched.
  - id: f1
  - line: 68-79
- **`.stale` is treated as dead, but a stale activity is visible and updatable -- updating it is the remedy, and this repo's widget expects that state** (P2)
  - detail: `guard let activity, activity.activityState == .active` (:31) conflates 'not updatable' with 'not live'. ActivityState.stale means the content is out of date and needs an update; the activity is still on screen and update() on it is the documented cure. Under this predicate a stale activity is instead forgotten (:128) and duplicated (f1). This is not hypothetical in this product: HerdrWidgets/AgentLiveActivity.swift branches on `context.isStale` in fourteen places (:11, :37, :48, :56, :65, :113, :124, :179, :201, :210, :267, :272, :361, :380, :393), including a dedicated 'last update' + relative-time line and a hedged headline -- i.e. the widget half is built for a state the controller half now treats as death. Both local sites pass staleDate: nil (:99, :140), so today the only producer is the server's push path (`stale-date` in the APNs payload), which is not in this repo and which I could not inspect; that is the reachability caveat on this finding. The split that makes both questions right: `.active || .stale` for 'can I push into it?' (the liveActivity property and update's guard), `.active` only for 'may I adopt this one as mine?' (the reclaim predicate in f1). As written, the two questions share one answer and the stale case gets the wrong one.
  - id: f2
  - line: 31
- **Recovery is triggered only by a roster mutation -- the one signal this file documents as absent in the state the activity exists for** (P2)
  - detail: The re-mint hangs off update() (:135), and update has exactly one caller: `.onChange(of: fullList, initial: true)` at HerdrApp.swift:2429. The comment at :89-96 in this very file establishes, as the accepted justification for staleDate being nil, that nothing re-pushes on a cadence and that a stable roster therefore produces no updates -- naming the important case as 'a HEALTHY, stable roster -- one agent blocked, waiting for you, the state this activity exists for'. I confirmed the mechanism rather than trusting the comment: AgentList is Equatable over AgentRow(info, status, group, isLive), and AgentInfo's time field is `statusSinceUnixMs` (Sources/HerdrKit/AgentList.swift:291), a server-reported state-entry instant that does not advance while the state holds, so each 5s poll (HerdrApp.swift:1418) yields an equal value and fires nothing. Consequence: swipe the banner while one agent sits blocked waiting for you, and it stays gone for as long as the herd is quiet -- the exact scenario the feature exists to serve. `initial: true` does not cover it: that modifier sits on the `.id(session)`-scoped view (HerdrApp.swift:259) and re-fires on connect/reconnect only, never on a mere foreground. Second, independent reason the trigger is wrong: Activity.request throws ActivityAuthorizationError.visibility when the app is in the background, and the app IS backgrounded at the moment of a Lock Screen swipe, so the first attempt fails by construction and recovery is deferred to the next roster change that lands while foreground. A scene-phase hook (re-attempt on scenePhase == .active, or a foreground update() call) makes recovery depend on the reader returning to the app rather than on the herd happening to move, and is also the natural place to retry after a .visibility rejection.
  - id: f3
  - line: 135
- **The forget path drops `activity` but leaves tokenTask watching the dead activity and `pushToken` published; the dead token is never unregistered on re-mint** (P3)
  - detail: update()'s guard nils `self.activity` (:128) without cancelling tokenTask or clearing pushToken, and observePushToken is not called on that path. If the subsequent start() returns early at `guard enabled` (:57) or the request throws (:103-107), the controller keeps publishing the DISMISSED activity's token, which RootView has already registered with the server (HerdrApp.swift:267, :296-299). HerdrClient.unregisterActivity's own doc comment (Sources/HerdrKit/HerdrClient.swift:368-369) states the server prunes a token only when told to, and the sole unregister caller is disconnect (HerdrApp.swift:471-477) -- so every swipe-and-recover cycle in one session leaves another dead activity token registered server-side. Not user-visible (pushes to a dismissed activity go nowhere) and partly pre-existing, but the PR makes mid-session re-minting routine where before it was rare, so the leak becomes per-cycle rather than per-process. Calling observePushToken() on the forget path already does the right thing -- its `guard let activity else { pushToken = [REDACTED]; return }` (:115) clears the published token and cancels the task in one step. Verified NOT broken: the new activity's token does reach the server, because `.onChange(of: liveActivity.pushToken, initial: true)` (HerdrApp.swift:267) is live for the whole connected session, so background updates survive a re-mint.
  - id: f4
  - line: 127-128
- **Answer to the asked question: the hostLabel-survives-failure retry is correct and load-bearing -- do not add a one-shot flag or backoff** (P3)
  - detail: `self.hostLabel = hostLabel` at :81 precedes the do/catch and the catch (:103-107) clears only `activity`, so a failed mint does leave every later roster change retrying. That is the right behaviour and must be kept: Activity.request throws ActivityAuthorizationError.visibility from the background, which is exactly where the app sits when the reader swipes the banner off the Lock Screen, so the first attempt is EXPECTED to fail and a one-shot latch would disable the fix in its primary scenario. No backoff is needed either -- update is edge-triggered by `.onChange(of: fullList)` and the 5s poll (HerdrApp.swift:1418) bounds attempts at roughly one per 5s, and only while the roster is genuinely moving, so there is no spin. The only refinement worth considering is declining to retry the terminal cases by matching ActivityAuthorizationError -- .unsupported, .unentitled, .attributesTooLarge, .globalMaximumExceeded, .targetMaximumExceeded -- since those cannot succeed later; .denied is already absorbed by `guard enabled` (:57). Acceptable as-is; recorded so the next reader does not 'fix' it into a one-shot.
  - id: f5
  - line: 81-107
- **Verified clean: no recursion, no double-mint, and the redundant nil at :128 is harmless** (P3)
  - detail: start reaches update only through `if liveActivity != nil { observePushToken(); update(state); return }` (:82), so the nested update's `guard let activity = liveActivity` (:127) succeeds and the chain terminates at depth 2; the mutual recursion is bounded by construction, not by luck. Interleaving is impossible too: the type is @MainActor, and Activity.request is called as the non-async throwing overload (:97, `try` not `try await`), so between the liveness read at :82 and the mint at :97 there is no suspension point at which a second update could enter. The detached `Task` at :74 captures the `existing` array and `keep` by value before the mint, so the freshly requested activity is not in that array and cannot be ended by its own sweep. `self.activity = nil` at :128 is redundant (the guard already proved liveActivity is nil, and start repeats the assignment at :67) but costs nothing and documents intent.
  - id: f6
  - line: 82
- **Verified clean: the remembered hostLabel cannot go stale against a reconnect to a different host** (P3)
  - detail: hostLabel is written only at :81 and cleared only at :151, and the call graph closes: start's sole caller is connect (HerdrApp.swift:443), reachable only while `credentials == nil` because ConnectView is presented only on that arm (HerdrApp.swift:277-278); end's sole caller is disconnect (HerdrApp.swift:483); reconnect (HerdrApp.swift:490) and trustAndReconnect (:510) rebuild the transport from the RETAINED credentials and touch neither, so the host cannot change under them. Every host change therefore passes through connect -> start, which overwrites hostLabel before any mint. The only route that skips the write is `guard enabled` (:57), and start re-checks that same guard when update re-enters it, so a label left over from an earlier enabled period can never reach Activity.request. Corollary worth knowing but NOT a finding here: the wrong-host guard at :64-80 is skipped whenever a live activity is already held, so a connect to host B while host A's banner is active would push B's roster under A's label -- byte-identical to the pre-PR `if activity != nil { update; return }`, and unreachable today because of the credentials gating above.
  - id: f7
  - line: 81

**Tests Run**
- git rev-parse HEAD -> 8ddcb50857b9656bf16e80852938e29c7a59d5b2 (matches the requested head); git log --oneline -3 confirms base 9c74ecb; git status --porcelain empty
- git diff 9c74ecb HEAD -- App/LiveActivityController.swift -> 1 file, +45/-5, read in full
- Full read of App/LiveActivityController.swift at this head (197 lines, all bodies)
- Repo-wide grep for LiveActivityController.shared start/update/end call sites -> start: HerdrApp.swift:443 (connect) only; update: HerdrApp.swift:2430 (.onChange(of: fullList, initial: true)) only; end: HerdrApp.swift:483 (disconnect) only
- Repo-wide grep for pushToken producers/consumers -> HerdrApp.swift:267 onChange(initial: true) -> registerActivityPush (:296-299), :472 unregister on disconnect; HerdrClient.swift:361/369 register/unregister contract
- Repo-wide grep for activityState / staleDate / isStale -> staleDate nil at LiveActivityController.swift:99 and :140; context.isStale consumed in 14 places in HerdrWidgets/AgentLiveActivity.swift
- Read HerdrApp.swift:420-515 (connect/disconnect/reconnect/trustAndReconnect) and :2378-2448 (poll .task + the update trigger); agentListPollIntervalNanoseconds = 5_000_000_000 at :1418
- Read Sources/HerdrKit/AgentList.swift equality surface (AgentRow :101-111, AgentList :220-232) and AgentInfo.statusSinceUnixMs (Wire.swift:291) to establish that a quiet roster is Equatable-equal across polls
- Apple-documentation verification of the three semantics the change turns on: ActivityAuthorizationError.visibility ('the app tried to start the Live Activity while it was in the background'); default dismissal policy keeps an ENDED activity on the Lock Screen up to four hours and state does not become .dismissed until the UI is removed; ActivityState.stale means content is out of date and needs an update (i.e. a stale activity is updatable)
- NOT RUN, no toolchain: `which swift swiftc xcodebuild` returned nothing, so no build, no swift test, no simulator run. WidgetGalleryTests run 35008909059 was neither re-run nor fetched and is taken as the requester's evidence.

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5996ca160d79f-1`

**Decision:** `approved`
**Summary:** Re-reviewed jerryfane/herdrup PR #258 at exact head 8cb9058f36cb556d6763f9cf227ba862aa10f6ba (git rev-parse HEAD matches; GitHub API reports head 8cb9058, base 9c74ecb, 2 commits, 2 files changed; worktree clean). Delta vs the previously reviewed 8ddcb50 is one commit touching App/HerdrApp.swift (+9) and App/LiveActivityController.swift. All three P2s from the 8ddcb50 verdict are genuinely fixed, not papered over. (1) TRIGGER: recoverIfEnded (LiveActivityController.swift:154) is called unconditionally from the 5s roster poll (HerdrApp.swift:2414), so recovery no longer depends on the roster MOVING — which is the exact state a blocked agent does not produce. It guards on enabled/liveActivity==nil/hostLabel!=nil and pushes no content when one is live, so a stable roster costs no ActivityKit update budget; .onChange(of: fullList, initial: true) (HerdrApp.swift:2438) still carries real state. Double-mint is impossible in either interleaving: whichever path arrives first makes liveActivity non-nil, and the other degrades to a single content push. The update<->start mutual recursion is bounded because there is no suspension point between the isLive read at start:105 and the one at update:163 (both @MainActor, Activity.request is synchronous), so an activityState flip cannot interleave. (2) STALE: liveActivity (:36) and isLive (:46) both treat .active and .stale as live; only .dismissed/.ended are dead. Correct — the widget half is built entirely for context.isStale (AgentLiveActivity.swift:11,37,113,124,179,201,210,272,361,380,393), so forgetting a stale activity would duplicate a visible banner. (3) SWEEP: the isLive filter is now only on the reclaim predicate (:92-94), Activity.activities is iterated whole (:91,:96-101), so a dismissed same-host activity is no longer adopted as `keep` and gets ended rather than left beside the fresh mint. Strictly better than the pre-PR behaviour. JUDGEMENT ON YOUR FOUR QUESTIONS. (a) Cancelled-poll window: the .task is .id(session)-sco
[truncated]
**Severity:** `P3`

**Findings**
- **hostLabel is set after the `enabled` guard, so enabling Live Activities mid-session never recovers** (P3)
  - detail: start() returns at `guard enabled else { return }` (:75) BEFORE `self.hostLabel = hostLabel` (:104). If the user has Live Activities off in Settings when connect() calls start (HerdrApp.swift:443), hostLabel stays nil for the whole session, and recoverIfEnded's `hostLabel != nil` guard (:155) then blocks recovery forever even though its `enabled` check would now pass. The new periodic hook re-evaluates `enabled` every 5s, so this guard ordering is the single remaining reason a mid-session Settings toggle does not take effect until reconnect. Hoisting `self.hostLabel = hostLabel` above the guard is a one-line fix. Pre-existing behaviour, not a regression of this PR.
  - id: f1
  - line: 75
- **Recovery latency is coupled to `await load()`** (P3)
  - detail: recoverIfEnded is invoked after `await load()` in the poll body, so on the first foreground pass the re-mint waits for the agent.list round trip (or its timeout) even though it does not need a fresh roster: `fullList` is stale-while-revalidate and already holds the last-good list. Calling recoverIfEnded before `await load()` would make foreground recovery immediate. Related nit: `LiveActivityController.state(from: fullList)` is evaluated at the call site on every poll, before the callee's guards, and activityContent (AgentList.swift:353) does one min-scan plus four filter allocations — so the comment's 'one predicate per poll' understates it. Immaterial at 5s, mildly relevant on the DEBUG rosterStress path (HerdrApp.swift:340) where the interval is 50ms and a UI test is measuring roster publication timing; hostLabel is nil there so nothing mints, but the state derivation still runs 20x/s. An autoclosure parameter or a `needsRecovery` predicate would remove both.
  - file: App/HerdrApp.swift
  - id: f2
  - line: 2407
- **No cure path for a `.stale` banner if a stale date ever appears** (P3)
  - detail: Treating .stale as live is the right call and I am not asking you to change it. The observation is that nothing then refreshes a stale activity: recoverIfEnded deliberately pushes no content while liveActivity != nil, and the only other update trigger is .onChange(of: fullList), which cannot fire on the frozen roster this activity exists for. An update is the documented cure for staleness, so a stale banner would stick — the same shape as the bug this PR fixes. Currently unreachable: staleDate is nil at both :122 and :182 and no other staleDate/stale-date producer exists anywhere in the repo, so the widget's isStale branches only light up for a server push carrying one. If the server sender ever starts setting a stale date, recoverIfEnded needs a second clause (push content when liveActivity?.activityState == .stale).
  - id: f3
  - line: 154
- **Dismissed activity's push token is orphaned server-side once per recovery cycle** (P3)
  - detail: I agree the missing unregister is not user-visible from the client, with one correction to your framing: it is not that the token merely stops being published, it is that the server keeps it. disconnect() (HerdrApp.swift:472) unregisters whatever pushToken currently holds, which after a re-mint is the NEW token, so every dismissed activity's token stays registered until the server prunes it on APNs rejection. That is now once per dismissal-recovery cycle rather than never. The server is not in this repo (iOS client only), so I cannot check whether notifications.register_activity is last-writer-wins per device (fine, self-healing) or accumulates (dead tokens forever, and user-visible if the sender iterates tokens and aborts on the first failure — background widget updates would silently stop). No client change needed if it is last-writer-wins; worth confirming server-side. HerdrClient.unregisterActivity exists, but the controller holds no client reference, so a client-side fix means publishing the dead token for RootView to unregister — more than a one-liner, and correctly deferred.
  - id: f4
  - line: 167
- **Under the new trigger, a dismissal is no longer escapable for the session** (P3)
  - detail: Design consequence worth naming because the trigger change amplified it. At 8ddcb50 resurrection needed a roster change, which is rare precisely when the roster is frozen; now it is unconditional on every foreground poll. Dynamic Island activities cannot be swiped away, and a Lock Screen swipe implies the app is backgrounded and (no UIBackgroundModes) suspended, so the banner returns on the next foreground rather than in 5s — the loop is bounded in practice, and the only user controls left are the system Settings toggle (respected via the `enabled` guard) and disconnecting. That matches your stated product decision and the prior 'no one-shot latch' advice, so I am not reversing it; flagging it because the advice was given under the weaker trigger. If you ever want a per-session opt-out, the minimal form is a latch set on the forget path and cleared on the next real fullList change, not a backoff.
  - id: f5
  - line: 171
- **`$0.id != keep` relies on implicit Optional promotion to mean 'end everything'** (P3)
  - detail: In the no-reclaim case `keep` is nil, and `$0.id != keep` promotes the String id to Optional, so the predicate is true for every activity — which is the intended 'end every orphan' behaviour and is what makes the dismissed same-host activity get swept. It reads as a same-host comparison though, and the sweep is the one place where getting it wrong silently leaves a visible orphan. An explicit `keep == nil || $0.id != keep` (or `!(keep.map { $0 == a.id } ?? false)`) documents the intent for whoever edits this next, including the future multi-session change.
  - id: f6
  - line: 96

**Tests Run**
- git rev-parse HEAD / git status --porcelain (head == 8cb9058f36cb556d6763f9cf227ba862aa10f6ba, tree clean)
- git diff 9c74ecb..8cb9058 --stat and full diff (App/HerdrApp.swift +9, App/LiveActivityController.swift +92/-5)
- GitHub API: repos/jerryfane/herdrup/pulls/258 -> head 8cb9058, base 9c74ecb, 2 commits, 2 files
- GitHub API: actions/runs/35016955187 -> head_sha 8cb9058, status completed, conclusion success; jobs 'HerdrKit and SwiftTerm (Linux)' success and 'build-and-uitest' success, including the simulator compile check (app + widget + Metal) and the iPhone/iPad UI test step
- Full-file read of App/LiveActivityController.swift at this head; targeted reads of HerdrApp.swift connect/disconnect/reconnect (425-520), the poll .task and onChange (2380-2440), and the DEBUG rosterStress construction (300-360)
- Repo-wide greps: LiveActivityController.shared call sites, agentListPollIntervalNanoseconds, staleDate/stale_date/relevanceScore producers (none outside LiveActivityController, both nil), isStale consumers in HerdrWidgets, register_activity/unregister_activity, UIBackgroundModes/BGTaskScheduler (project.yml:214 declares none)
- NOT run: swift build / swift test / xcodebuild — no Swift toolchain in this seat (which swift swiftc xcodebuild all empty), so no Swift code was compiled or executed by me

PR #259, opened 2026-09-16
Paint the lock screen card with a top-to-bottom gradient

jerryfane:
Owner's call, after seeing the alternatives on a device: "it's not translucent at all, so we can make the change" — the system surface did not read as glass on their wallpaper either, so the card goes back to painting its own background, this time as a **top-to-bottom gradient**.

## What changes

`LockScreenView` paints `WidgetPalette.cardGradient`, `#1B1F3A` → `#0B0E1A`, vertical.

Owning the surface again returns everything that vibrancy took away:

| | Build 146 (system surface) | Here |
|---|---|---|
| Hero number | System ink — amber measured 1.4:1 over a white wallpaper | **Amber**, 7.7:1 against the light stop |
| Status mark | On a `ground` disc, because the wallpaper was unknown | Bare — every surface it renders on is ours |
| Text tiers | `.primary` / `.secondary` / `.tertiary` via vibrancy | The tuned tokens: text 13.7:1, textDim 6.2:1 |
| Action label + border | System ink / `.tertiary` | `text` and `textFaint` (3.5:1 at the foot) |
| `activityBackgroundTint` | nil | `ground`, matching the gradient's foot so the seam does not show |

## The one trap avoided

My first stops were `#1F2445 → #0D1020`. Lifting the head like that dropped the fleet line's tier (`textFaint`) from **3.19:1 to 2.98:1** — a contrast regression smuggled in under a cosmetic change. The sweep now comes from **deepening the foot**: the top stop is exactly the lift the old flat backdrop used, so the head is no less legible than before.

`WidgetPalette.backdrop` and the `onSystemSurface` plumbing on `StatusMark` and `ActivityAction` are deleted rather than left dead, along with the disc.

## Receipts

- `WidgetGalleryTests` iPhone run 35060895507 SUCCESS at this head (app + widget compile, gallery capture).
- Gradient verified in the captured frames, sampled down one card: `#191F38` at the top, `#131727` mid, `#0E0F1E` at the foot.

## Not verified

How it looks on the owner's own wallpaper — but that is now a smaller question than in the previous three rounds, because nothing about this surface depends on what is behind it.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5b870ae33f403-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/herdrup PR #259 at exact head b2a27a6d7e06bd39d5b344e858d2d356c903aa32 (git rev-parse HEAD matches; worktree clean; git diff 1e8c91f..b2a27a6 = HerdrWidgets/AgentLiveActivity.swift only, +60/-110). Prior verdicts read from GITMOOT_PRIOR_VERDICTS (frozen 2026-09-16T06:13:48Z): no entry for #259; the #257 pair (90d181bc changes_requested P0, db7b3dd approved P3) is the decision this PR reverses. EVIDENCE static_only and not by choice: `which swift swiftc xcodebuild` is empty in this seat, so CI 35060895507 could not be re-run and is taken as the requester's evidence. What I DID execute: git diff/log/show/rev-parse at this head, repo-wide greps for every deleted symbol, and a WCAG relative-luminance computation over all seven palette tokens against every surface in play, including the gradient interpolated at seven positions and the requester's three sampled frame colours.

THE REVERSAL IS COHERENT AND COMPLETE. The surface is ours again (LockScreenView:312 .background(WidgetPalette.cardGradient)), so every consequence follows: fixed tokens at :320-322, amber hero at :327, no disc, no onSystemSurface. Contrast arithmetic checks out where it matters — textFaint at the light stop #1B1F3A is exactly 3.19:1 and the rejected #1F2445 head is exactly 2.98:1, so the author's self-caught regression is real and the shipped stops genuinely avoid it; textDim 6.22 (claimed 6.2) and amber 7.66 (claimed 7.7) also hold. The sweep is monotone in the right direction and my interpolation reproduces the sampled frames (t=0.5 -> #13162A vs sampled #131727).

FIVE JUDGE QUESTIONS, ANSWERED.
(1) textFaint at 11pt: acceptable, and 3.19:1 is conservative rather than the operative number. 3.19 is the head of the gradient, where the only run is the 17pt semibold headline in `text` (14.15:1). The fleet line (lockDetail :388, plex 11) renders roughly 40-55% down the card — interpolated #14172C to #121628, i.e. 3.50-3.55:1, matching the sampled #131727 -> 3.53:1. That tier still 
[truncated]
**Severity:** `P3`

**Findings**
- **Four measured claims in the new comments disagree with the arithmetic** (P3)
  - detail: In a PR whose entire content is colours and the comments recording them, four numbers are off, all in the safe direction: (a) :317 and :487 say text 13.7:1 against #1B1F3A; it is 14.15:1 (the #252 review computed the same 14.15). (b) ActivityAction :412 says textFaint 'measures 3.5:1 against the gradient's foot' — 3.54:1 is textFaint against ground #13162A, the old flat surface; against the foot #0B0E1A it is 3.81:1, so the number was carried over and re-labelled. (c) StatusMark :429-432 says the hue 'measures 4.9:1 or better against all three' surfaces; at the gradient's head, where the mark actually sits, died is 4.40:1 and idle (textFaint) is 3.19:1, and on the island's black idle is 4.16:1. Both still clear the 3:1 non-text bar of WCAG 1.4.11 and idle is unchanged from the pre-#257 head, so no mark is illegible — but '4.9:1 or better' is not true of two of the four statuses. (d) :16-17 says the tint 'matches the gradient's foot'; WidgetPalette.ground #13162A is the exact sRGB midpoint of #1B1F3A and #0B0E1A, 1.08:1 from the foot, and the task brief repeats the same claim. Non-blocking; nothing renders differently.
  - id: f1
- **SystemSurfaceStandIn is now dead paint the PR's own rule would delete** (P3)
  - detail: HerdrWidgets/AgentLiveActivity.swift:656 still puts SystemSurfaceStandIn (:244-248) behind LockScreenView in the gallery, but the card now fills its own frame with an opaque gradient, so the material is completely occluded. The two backdrops (:608-613) survive only as a 10pt halo, which makes the 'bright' and 'dark' lock-card rows pixel-identical apart from that ring — twelve frames carrying six frames of information. This PR deleted backdrop, onSystemSurface and the disc branch rather than leave them dead; by the same rule the stand-in, its doc comment, and the now-false card() doc at :640-642 ('the card itself now draws no background of its own, which is what makes it translucent on device') belong in the same commit. The island and pill rows still need the gallery, so this is a trim, not a removal.
  - id: f2
- **Three comments left behind by the reversal now state the opposite of the code** (P3)
  - detail: :494-498 (islandTextFaint) still reads 'unlike the lock screen, which now sits on the system's surface and asks the system for ink'; :272-274 justifies ink over the status hue with 'a surface whose brightness we do not control'; :601-604 describes the gallery's bright backdrop as 'the worst case for a translucent card' and warns that nothing reproduces the vibrancy applied to '.primary and friends', which no shipping view in this file uses any more. This file's comments are its measurement record and the reason three prior rounds were reviewable at all; leaving the #257 rationale in place invites the next reader to re-derive the wrong surface.
  - id: f3
- **textFaint remains below WCAG AA for 11pt text on this card (pre-existing)** (P3)
  - detail: At every position the fleet line can occupy, textFaint #666D91 on the gradient measures 3.19-3.81:1, against the 4.5:1 AA bar for 11pt non-large text. Unchanged from the flat backdrop that shipped for months and not introduced here, so it is not a merge condition — recorded so the accepted number is on the record with its in-situ value (~3.5:1) rather than only the head-of-gradient 3.19:1.
  - id: f4

**Tests Run**
- git rev-parse HEAD -> b2a27a6d7e06bd39d5b344e858d2d356c903aa32; git status --porcelain -> clean
- git diff 1e8c91f..b2a27a6 --stat -> HerdrWidgets/AgentLiveActivity.swift, +60/-110 (full diff read)
- git grep -n 'onSystemSurface|WidgetPalette.backdrop|glassColor|onGlass|hairline' -> zero hits in HerdrWidgets (hairline hits are App/Palette, unrelated)
- git grep StatusMark( / ActivityAction( / .primary|.secondary|.tertiary over the file -> 7 + 2 call sites, all declared labels; system-ink styles only in DEBUG gallery captions
- git log -S 'activityBackgroundTint(WidgetPalette.ground)' -> a65319f (original), ac4c6f4 (removed for glass), b2a27a6 (restored): the tint is a genuine revert, not a new value
- WCAG relative-luminance computation (python) over 8 tokens x surfaces {#1B1F3A, #0B0E1A, #13162A, #000000, rejected #1F2445/#0D1020, sampled #191F38/#131727/#0E0F1E} plus the gradient at t=0/.25/.35/.45/.55/.75/1
- which swift swiftc xcodebuild -> empty; nothing compiled or run, CI 35060895507 not re-executed

PR #260, opened 2026-09-16
Match the Live Activity to the original design and lighten its gradient

jerryfane:
## Owner request and design authority
Owner requested a lighter gradient bottom and fidelity to the original design files. Authorization record: workflow adhoc/herdr-ios-review-opus-2026-09-12, note 167558.

Compared the original `Herdrup Island - Design.dc.html` and `Herdrup Island - On Device.dc.html` on this machine against compiled simulator screenshots. Preserve the owner's later choices: opaque vertical gradient and the expanded island's full-width bottom content.

## Changes
- Lighten the bottom stop from #0B0E1A to #171B30.
- Place the status mark beside the headline; keep the hero count on the left and the actual app icon on the right.
- Share the existing AppLogo asset between app and widget instead of redrawing it in SwiftUI.
- Match spacing and the one-agent/zero/stopped/stale/dimmed states. Stopped agents get Open; working/idle zero state says Nothing needs you without rewriting stopped or connecting states.
- Share freshness wording with the testable state model and cover the working-roster and expired-snapshot regressions.
- Expand the screenshot gallery and verify a real card action instead of pinning fixture captions.

## Verification at 75df3fbd0cca8f55ab8a5bb1fcc7e53c75c4e486
- CI run https://github.com/jerryfane/herdrup/actions/runs/35078228919 succeeded: app/widget build, WidgetGalleryTests capture, Linux job.
- Eight distinct final screenshots downloaded and inspected against the original on-device mock.
- `swift test --filter AgentActivityStateTests`: 11 tests passed. Quiet-working regression reproduced failing before the condition fix.
- Independent read-only reviewer WidgetDesignAudit approved this exact head with no blockers; same model family, not the implementer. Review and evidence saved in workflow notes 167594 and 167619; subagent review record `/root/.omp/agent/sessions/-tmp/2026-09-05T08-15-33-989Z_01a070a3-2065-71ce-923e-5bbeb3587a4f/WidgetDesignAudit.md`.

## Explicit limits
No TestFlight deployment or on-device verification is claimed. The executing daemon build 5a244caa60b0c3a5742315c59d20ed81c05bc23e supplies neither question text nor a safe default answer/prompt identity; Open is the design's required fallback, not a substitute fake Approve. The existing controller's nil staleDate remains unchanged. Full interactive approval needs a separate real daemon contract.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5c6e45bc9b7fd-1`

**Decision:** `approved`
**Summary:** Reviewed PR #260 at exact head 75df3fbd0cca8f55ab8a5bb1fcc7e53c75c4e486 (git rev-parse HEAD matches; net diff vs merge-base 2476210 is 8 files, 117+/136-). No prior verdict exists for #260 in the frozen prior-verdicts set (as_of 2026-09-16T10:39:51Z; it covers 244-259), so the WidgetDesignAudit approval is not corroborated here and this is an independent pass. EVIDENCE static_only and not by choice: `which swift swiftc xcodebuild xcodegen` is empty in this seat, so CI 35080116106, capture 35078228919 and the 11 AgentActivityStateTests could not be re-run; I executed git rev-parse, the full git diff, targeted reads, and a Python WCAG recomputation of the palette. APPROVED: the two things the brief flagged as riskiest are both correct. (1) ASSET BUNDLING app/widget: project.yml is unchanged and already lists `- path: Shared` for BOTH the app (:74) and HerdrWidgets (:309); xcodegen classifies .xcassets as a resource by extension regardless of target type, so Shared/Assets.xcassets lands in the extension's own bundle and `Image("AppLogo")` (AgentLiveActivity.swift:269, no explicit bundle, so Bundle.main = the extension) resolves on device, not merely in the in-app gallery that the capture screenshots. Root Contents.json is present, the shared catalog carries no AppIcon set, HerdrWidgets does not set ASSETCATALOG_COMPILER_APPICON_NAME, and AppLogo does not collide with App/Assets.xcassets' AppIcon, so actool has nothing to fail on. The Package.swift exclude is required, not cosmetic: AgentActivityState is the only SwiftPM target rooted at Shared, and the new xcassets would otherwise be an unhandled file; excluding AgentActivityAttributes.swift is redundant with the explicit `sources` but harmless. (2) STATE SEMANTICS: the new presentationHeadline all-clear cannot reproduce the defect the deleted comment recorded. `status` is the LEAD's group (AgentList.swift:356-361) and stopped ranks LAST in activityLead (:292), so lead == .stopped only when no needsYou/unrecognised/wor
[truncated]

**Findings**
- **WidgetPalette.ground is now dead code, and a doc comment still cites it** (P3)
  - detail: The tint moved from WidgetPalette.ground to WidgetPalette.cardBottom (:15), and `ground` now has no consumer anywhere in either target (grepped HerdrWidgets/ and App/: only the definition matches). Swift does not warn on an unused static let, so CI stays green while the token rots. The ActivityAction doc comment at :353 still describes the absent Approve control as 'ink fill on a `ground` label', so the surviving reference points at a dead token.
  - id: f1
  - line: 435
- **The action title uses the raw headline while the card hedges it** (P3)
  - detail: Both stale surfaces now hedge the headline through presentationHeadline ('No recent update', or 'N may need you'), but the ActivityAction title is still built from raw `state.headline` (lock :279, island :157-160). On one stale card the headline refuses to name the agent while the button underneath says 'Open api-refactor' with full confidence. Passing presentationHeadline(isStale:) to the title, or dropping the name when stale, would make the card agree with itself. Unreachable in production today (see f6), which is why this is P3 rather than P2.
  - id: f2
  - line: 279
- **'No recent update' is a new user-visible string with neither a test nor a capture** (P3)
  - detail: The stale/zero branch is the only new wording in the PR that nothing pins: grep finds 'No recent update' at exactly one site and in no test. The two new tests cover the quiet and stale-with-demand paths only, and both stale gallery cards (AgentLiveActivity.swift:560, :566) reuse cases[0], which has needsYouCount 23, so every captured stale frame renders '23 may need you'. The zero-count stale card was never rendered. A gallery case with needsYouCount 0 plus isStale true, or one XCTAssertEqual, closes it.
  - file: Shared/AgentActivityState.swift
  - id: f3
  - line: 135
- **AppLogo ships the full 1024x1024 app icon into a 26pt slot in the widget's memory budget** (P3)
  - detail: The new PNG is byte-identical to App/Assets.xcassets/AppIcon.appiconset/icon-1024.png (cmp clean; 26277 bytes, 1024x1024, 8-bit colortype 2, no alpha), so the artwork provenance is right. But it is a single-scale image set rendered at 26x26pt: WidgetKit decodes a 1024x1024 bitmap (~4MB resident) and .interpolation(.high) downsamples it ~39x on every render, inside an app-extension render budget that is the tightest surface in the app. A 78px derivative (26pt @3x) or explicit @1x/@2x/@3x slots would cost nothing visually. Also note the app target now bundles the same payload twice (AppIcon set plus AppLogo set) -- 26KB, immaterial on its own.
  - file: Shared/Assets.xcassets/AppLogo.imageset/AppLogo.png
  - id: f4
- **The comment deletions drop hard-won evidence this repo has already paid for twice** (P3)
  - detail: The diff removes the recorded proof that the glass route is closed -- a material painted INSIDE a Live Activity does not sample the wallpaper because iOS composites its own surface behind the view, measured on builds 144-146 -- and replaces it with 'The owner chose an opaque vertical gradient after viewing the system surface.' That one-liner records the decision but not the evidence, so nothing in the file now stops a fourth attempt. Same pattern for the deleted measured contrast figures and for the 'NO HEADLINE OVERRIDE LIVES HERE' block (the latter is at least partly obsolete now that the rule moved to a tested file, and testQuietHeadlineDoesNotHideStoppedAgentsOrConnecting carries the intent). Given that this repo's comments elsewhere exist precisely because these defects recurred, the build-144-146 finding is the one worth keeping.
  - id: f5
  - line: 287
- **Every isStale branch touched by this PR is unreachable on device** (P3)
  - detail: staleDate is nil at both ActivityKit sites (:122 start, :182 update), so context.isStale is always false in production and the stale corrections in this PR -- 'No recent update', the stale action button, the suppressed hero, line(_:isStale:true) -- are exercised only by the gallery's synthetic isStale flag and the unit tests. Pre-existing and explicitly declared out of scope by the requester; recorded so the verdict is not read as on-device verification of the stale states.
  - id: f6
  - line: 122
- **The gallery receipt's container assertion got weaker than the one it replaced** (P3)
  - detail: app.scrollViews.firstMatch is satisfied by ANY scroll view, whereas the old assertions pinned a specific caption and a specific card's content text. The gallery already carries .accessibilityIdentifier("widget-gallery") (AgentLiveActivity.swift:575), so the container could be named. The new live-activity-open identifier assertion is the stronger half and is a real improvement -- it proves a card body rendered, not just a caption -- but pairing it with app.scrollViews["widget-gallery"] would keep both guarantees.
  - file: HerdrUITests/WidgetGalleryTests.swift
  - id: f7
  - line: 22

PR #261, opened 2026-09-16
Size the widget logo to its slot so the device renders it

jerryfane:
Owner reported build 148 on device (gram `gram-1a0a9e92f03-1394e1-dd`): the Lock Screen Live Activity draws a **flat grey square** where the app logo belongs. Release authority for this workstream: workflow note 167778; defect and root cause recorded in note 167935.

## Cause

The widget drew `Image("AppLogo")` in a **26pt** slot while that asset was the **1024x1024** app icon — md5 `a598520…`, byte-identical to `AppIcon.appiconset/icon-1024.png`. A Live Activity will not render an asset whose resolution exceeds its presentation: it substitutes a grey placeholder, silently, with no build error and no runtime log.

The asset *did* reach the extension — the archive log for run 35086883478 shows `actool` compiling `Shared/Assets.xcassets` and `Emplaced …/HerdrWidgets.appex/Assets.car` — so this is a resolution limit, not a missing asset.

This is review finding **f4** from `local-review-herdr-ios-review-opus-18d5c6e45bc9b7fd-1`, which I merged as nonblocking. That was my error.

## Fix

| asset | slot | sizes |
|---|---|---|
| `AppLogo` | app header, 64pt | 64 / 128 / 192 px |
| `AppLogoSmall` (new) | widget mark, 26pt | 26 / 52 / 78 px |

Both declare real 1x/2x/3x entries instead of one unscaled image.

## Why the green receipt missed it

The UI capture renders `LockScreenView` **inside the app target**, in-process, where the limit does not apply — so capture 35078228919 showed the logo correctly while the device showed grey. A passing screenshot was read as proof of a surface it structurally cannot observe.

So the new guard is arithmetic over the asset catalogue, the part that was actually wrong: widget rasters must fit their slot at 3x, must live in `Shared` (the only catalogue in the extension bundle), and must declare all three scales.

- **Fails on #260's head** (`75df3fbd`), verified in a worktree: `AppLogo is 1024px at its largest … whose 3x presentation is 78px`, plus `declares scales []`.
- **Passes here**; 13 tests in the target, on Linux, no simulator or device.

## Not verified

Device rendering of this build. That needs the owner's iPhone on the next TestFlight build, and it is the only thing that can confirm the grey square is gone.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5c922740f4e6d-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/herdrup PR #261 at exact head 2d8659ccc776efa61d4193884dd2a2121bfa07f3 (git rev-parse HEAD matches; diff e50bba3..2d8659c = 10 files, +214/-7). Prior verdicts read from GITMOOT_PRIOR_VERDICTS (as_of 2026-09-16T11:19:52Z): #260 at 75df3fbd approved/7 findings, which is where f4 was accepted nonblocking; no prior verdict for #261, so this is a first-round pass. EVIDENCE static_only AND NOT BY CHOICE: `which swift swiftc xcodebuild xcodegen` is empty in this seat, so the cited 13-test Linux run could not be re-executed. WHAT I DID EXECUTE: git rev-parse/log/diff/archive, a faithful Python port of widgetSlots()+both assertions run against BOTH this tree and the 75df3fbd tree, SHA-256 byte comparison of every PNG, IHDR dimension extraction over all 7 rasters, and a hand-rolled zlib PNG decoder (no PIL available) to compare artwork.

APPROVED. The fix is correct and the guard is real. Independent confirmation of the claimed pass/fail: my port FAILS on 75df3fbd (AppLogo 1024px > 78px allowed, and scales set empty, so both new tests fail there) and PASSES here (AppLogoSmall widest 78px <= 78px, scales exactly {1x,2x,3x}, no empty filenames). Pre-fix AppLogo.png was byte-identical to App/Assets.xcassets/AppIcon.appiconset/icon-1024.png (SHA-256 6278eb2bc7d6 on both); at this head it is not. Decoding all six rasters to a 16x16 luminance grid shows every one is the same mark (colortype 2, opaque, matching the 'carries its own dark ground' comment); mean absolute channel difference against icon-1024 is 0.08 (AppLogo@3x), 0.25 (AppLogo 1x), 3.07/3.45/8.48 (Small @3x/@2x/1x) out of 255 — monotonic downscale blur, so these are genuine resamples of the icon, not a wrong crop and not an alpha-stripped variant.

ANSWERS TO THE FIVE QUESTIONS.
(1) IS points*3 RIGHT, OR SHOULD IT BE TIGHTER? It is right and it CANNOT be tightened. AppLogoSmall@3x is 78px and the ceiling is 26*3 = 78, so the assertion passes on exact equality with zero slack; any tighter bound rejects
[truncated]
**Severity:** `P3`

**Findings**
- **Slot parser can silently bind an unframed Image to a sibling view's .frame inside the 400-char window** (P3)
  - detail: widgetSlots() searches the 400 characters after each Image literal for the FIRST .frame(width:N,height:N) and attributes it to that Image. The no-frame arm is loud (XCTFail), but an Image with no frame of its own silently inherits the next view's bound. Executed against a Python port of the parser: probe A (unframed Image("BrandMark").resizable().scaledToFit() in compactLeading, followed 316 chars later by the lock-screen Image's .frame(width: 26, height: 26)) produced TWO slots both measured at 26pt and ZERO XCTFails; probe B (Image("Hero").resizable().scaledToFill() then Rectangle().frame(width: 320, height: 320)) measured the hero at 320pt, licensing a 960px raster, again with zero failures. AgentLiveActivity.swift:61 already carries a bare .frame(width: 36, height: 36) on StatusMark in the minimal region, so the adjacency exists today. NOT a defect at this head: the widget has exactly one Image and its own .frame sits 96 chars downstream. Suggested hardening: bind the frame search to the same expression (stop at the first line whose indentation returns to or below the Image's, or at the next Image(/ next top-level view), and XCTFail rather than reuse when a frame has already been consumed by a previous slot.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - id: f1
  - lines: 88-104
- **Non-literal and labelled Image initialisers are invisible to the guard, and the vacuity check cannot detect it** (P3)
  - detail: The pattern Image\(\s*"([A-Za-z0-9_]+)"\s*\) matches only a bare string literal with no argument label and no extra arguments. Executed probe C fed four realistic raster forms — Image(decorative: "AppLogo"), Image(uiImage: UIImage(named: "AppLogo")!), Image(assetName) with a variable, and Image("AppLogo", bundle: .main) — each with a .frame(width: 26, height: 26), and the parser produced 0 slots and 0 failures. XCTAssertFalse(slots.isEmpty) does not protect against this because the one surviving literal keeps the array non-empty, so coverage can silently drop from 100% to a fraction while the test stays green. None of these forms exists in HerdrWidgets today (verified: exactly one Image( in the whole directory), so this is future-proofing. Suggested hardening: also match Image( with a label or trailing arguments and UIImage(named:) and XCTFail on any form the guard cannot resolve to a catalogue name, rather than skipping it.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - id: f2
  - lines: 85
- **Widget source scan is non-recursive, so a subdirectory refactor silently drops coverage** (P3)
  - detail: contentsOfDirectory(at: widgets) is a single-level listing filtered on pathExtension == "swift", so only HerdrWidgets/*.swift is scanned. Confirmed today's tree has no subdirectories under HerdrWidgets and both sources (AgentLiveActivity.swift, HerdrWidgetBundle.swift) are top-level, so coverage is currently complete. But moving any widget view into e.g. HerdrWidgets/Views/ removes it from the guard while XCTAssertFalse(sources.isEmpty) still passes on the remaining top-level file — the same 'guard outlived the thing it guarded' failure mode the test's own docstring is about. Suggested fix: enumerator(at:) for a recursive walk.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - id: f3
  - lines: 78-82
- **Per-scale pixel dimensions are never asserted; a 78/78/78 catalogue passes both tests** (P3)
  - detail: This is the substantive answer to 'should the ceiling be tighter': the ceiling itself cannot be (78px == 26*3 exactly, zero slack, and 3x is the densest iPhone scale — tightening rejects the correct @3x). What is missing is per-scale equality. testWidgetRastersDoNotOutResolveTheSlot compares only max(all PNGs in the imageset) <= points*3, and testWidgetImagesetsDeclareEveryScale reads only the scale keys and non-empty filenames from Contents.json — neither cross-checks a declared scale against the actual IHDR of the file behind it. Executed consequence: an imageset whose 1x, 2x and 3x files are all 78px yields widest == 78 <= 78 and scales == {1x,2x,3x}, so it passes both, yet a 1x device would then draw a 3x-density raster in a 26pt slot — the same defect class at 1/39th the magnitude. Under-resolution (an @3x accidentally shipped at 26px, i.e. a blurry mark) is equally unguarded. Suggested strengthening: assert pixels(scale) == points * scaleFactor for each declared entry, which subsumes the current ceiling.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - id: f4
  - lines: 120-162
- **Two variables changed at once and the causal mechanism remains unverified; only a device Lock Screen receipt can close the defect record** (P3)
  - detail: The comment and the test docstring both state as fact that 'a Live Activity refuses to render an asset whose resolution exceeds its presentation'. That is not an Apple-documented rule, and this change moved two variables simultaneously: resolution 1024px -> 78px AND a single unscaled catalogue entry -> explicit 1x/2x/3x. Both candidate mechanisms (over-resolution, missing scale declarations) are addressed, which is why the change is safe and strictly more correct than what shipped — but which one iOS was rejecting is undetermined. Consequence for the guard: points*3 is a per-slot RATIO, not a byte budget. A future 300pt slot passes at 900px (~3.2MB decoded RGBA) where the current slot is 78px (~24KB), so if the real mechanism is the extension's memory budget, the guard bounds the wrong quantity and would not fire. Per instruction I treated no in-app capture as evidence: LockScreenView is compiled into the app target (project.yml:82) and renders in-process, so the gallery structurally cannot observe this class, and `swift test` cannot either. Recommendation: merge on the code, but keep gram-1a0a9e92f03-1394e1-dd open until a Lock Screen photograph from a build carrying 2d8659c is attached, and soften the two comments from asserted mechanism to observed symptom plus applied remedy so a future reader does not derive a new guard from an unproven rule.
  - id: f5
  - lines: 269-281

**Tests Run**
- git rev-parse HEAD -> 2d8659ccc776efa61d4193884dd2a2121bfa07f3 (matches the requested exact head)
- git diff --stat e50bba3..2d8659c -> 10 files changed, 214 insertions(+), 7 deletions(-)
- which swift swiftc xcodebuild xcodegen -> empty (no Swift toolchain in this seat; the cited 13-test Linux run could NOT be re-executed)
- Python port of widgetSlots() + testWidgetRastersDoNotOutResolveTheSlotTheyAreDrawnIn + testWidgetImagesetsDeclareEveryScale, run at HEAD -> AppLogoSmall 26pt, widest 78px <= 78px PASS, files 26/52/78, scales {1x,2x,3x} PASS, 0 empty filenames, 0 XCTFails
- same Python port run against a `git archive 75df3fbd` tree -> AppLogo 26pt, FAIL 1024px > 78px, scales [] (both new tests fail there): the claimed failing-on-75df3fbd / passing-here property is independently confirmed
- SHA-256 over every PNG: pre-fix Shared/.../AppLogo.png == App/Assets.xcassets/AppIcon.appiconset/icon-1024.png (6278eb2bc7d6) TRUE; at HEAD FALSE
- IHDR dimension extraction over all 7 repo PNGs -> AppLogo 64/128/192, AppLogoSmall 26/52/78, icon-1024 1024
- hand-rolled zlib/unfilter PNG decoder (PIL unavailable) + 16x16 downsample: all 6 logo rasters colortype 2 (opaque RGB); mean|diff| vs icon-1024 = 0.08 AppLogo@3x, 0.25 AppLogo1x, 3.07 Small@3x, 3.45 Small@2x, 8.48 Small1x -> same artwork, pure downscale blur
- parser probe A (unframed Image + sibling .frame(26,26) 316 chars later) -> 2 slots both 26pt, 0 XCTFails (false attribution, silent pass)
- parser probe B (unframed Image + Rectangle().frame(320,320)) -> 1 slot at 320pt, allowed 960px, 0 XCTFails
- parser probe C (Image(decorative:), Image(uiImage:), Image(variable), Image("X", bundle:)) -> 0 slots, 0 XCTFails (4 rasters unmeasured)
- grep -rn 'UIImage|Image\(|uiImage|decorative|AppIcon|Canvas|\.resizable' HerdrWidgets/ -> 2 lines, both the AppLogoSmall chain at :278-279 (DynamicIsland raster-free)
- grep -rn AppLogo across *.swift/*.yml/*.md/*.json -> one live consumer App/HerdrApp.swift:622 (frame 64x64 at :624) plus comments only
- grep -c 'func test' Tests/AgentActivityStateTests/*.swift -> 11 + 2 = 13 (matches the claimed target count)
- grep 'swift test|--filter|only-testing' .github/workflows/ -> ci.yml:71 `swift test` unfiltered, swift:6.1 ubuntu container, pull_request trigger (guard executes on this PR)
- git ls-files | grep xcodeproj -> none committed (.gitignore:3); xcodegen generate at ci.yml:116 and testflight.yml:92, project.yml:309 `- path: Shared` (new imageset is regenerated into the appex Resources phase)

**Needs**
- A device Lock Screen photograph from a build carrying 2d8659c, as the only evidence that can close defect record gram-1a0a9e92f03-1394e1-dd. CI and the in-app gallery both structurally cannot observe this class (LockScreenView is compiled into the app target at project.yml:82 and renders in-process).

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5ca3830e95e2f-1`

**Decision:** `approved`
**Summary:** Second-round review of jerryfane/herdrup PR #261 at exact head 17f9848bdef6b005045868d1caa384e234a1b16b (git rev-parse HEAD matches; diff 2d8659c..17f9848 = WidgetAssetResolutionTests.swift only, +132/-63; whole PR vs e50bba3 = 10 files, +283/-7). Prior verdict read from GITMOOT_PRIOR_VERDICTS (as_of 2026-09-16T11:40:34Z): my own approval of #261 at 2d8659c, P3, 5 findings, static_only. EVIDENCE static_only AND NOT BY CHOICE: `which swift swiftc xcodebuild` is empty in this seat, so the cited 13-test run could not be re-executed (I did confirm the count: 13 `func test` across the two files in Tests/AgentActivityStateTests). WHAT I DID EXECUTE: git rev-parse/log/diff/show, a faithful Python port of widgetSlots()+both assertions at this exact head, 15 mutations of that port against copies of this tree, IHDR reads of all six PNGs, and visual inspection of the rasters.

EVERY ARM THE BRIEF CLAIMS IS CONFIRMED. Baseline: one slot, ('AppLogoSmall', 26pt, AgentLiveActivity.swift), zero failures. Mutations CAUGHT: 78/78/78 in every scale slot (WRONG SCALE AppLogoSmall.png 78x78 != 26); build-148 replay with the real 1024px raster restored from e50bba3 and scales [] (OVER-RES 1024>78, SCALES []); unframed raster in HerdrWidgets/Views/Extra.swift (recursive enumerator works); unframed sibling BEFORE the framed image; Image(decorative:); Image(uiImage: UIImage(named:)); Image("Name", bundle:). Legit forms not broken: multi-line Image(\n "AppLogoSmall"\n) passes, Image(systemName:) with no frame passes. `^` does anchor at the argument window (ICU anchoring bounds are on by default; if they were not, EVERY Image call would XCTFail, so this cannot fail silently).

ANSWERS TO WHAT YOU ASKED ME TO CHALLENGE. (1) bareLiteral admits nothing it should reject that I could find — but it REJECTS two legitimate forms loudly: a hyphenated/dotted asset name ([A-Za-z0-9_]+ only) and Image(_:bundle:), the correct initialiser for a non-main-bundle asset; a custom symbol from a .symbolset is al
[truncated]
**Severity:** `P3`

**Findings**
- **Silent sibling-borrow is closed, but by f4's exactness rather than by the window bound; a non-Image sibling still lends its frame** (P3)
  - continues_uid: jerryfane/herdrup#261-f1
  - detail: Checked at this head by running the Python port with mutated copies of AgentLiveActivity.swift. chainEnd is the next `Image(` call only (lines 109-113), so an unframed raster followed by a NON-Image view still binds that view's frame: Rectangle().frame(width: 300, height: 300) yields slot points=300, a VStack yields 512, and an Image whose own frame is non-literal (.frame(width: logoSide...)) plus a literal sibling yields 400. What removes the silence is testEveryDeclaredScaleCarriesARasterOfExactlyThatSize: each mis-bind now fails loudly (WRONG SCALE AppLogoSmall.png 26x26 != 300/512/400). The one surviving false pass is a mis-bind whose borrowed number equals the correct slot (probe H: Rectangle().frame(width: 26, height: 26) after an unframed Image -> PASSED), which is benign because the raster is then correctly sized anyway. The reverse arm the brief cites is real: an unframed raster placed BEFORE the framed one is CAUGHT (UNFRAMED AppLogoSmall).
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - line: 109
  - rationale: Recording the actual mechanism matters: if the exactness assertion is ever relaxed back to a bound, the sibling-borrow becomes silent over-resolution again, because the window bound alone does not prevent it.
  - state: answered
- **Every Image( is now classified and unknown initialiser forms fail loudly** (P3)
  - continues_uid: jerryfane/herdrup#261-f2
  - detail: Checked at this head with the ported classifier over mutated copies of the widget source. Image(decorative: "AppLogoSmall") -> UNMEASURABLE; Image(uiImage: UIImage(named: "AppLogo")!) -> UNMEASURABLE (\bImage does not match UIImage(, so the inner call is not even enumerated); Image("AppLogoSmall", bundle: .main) -> UNMEASURABLE. Image(systemName:) is skipped without demanding a frame, and a multi-line Image(\n "AppLogoSmall"\n) still parses (PASSED, slot 26pt). The `^` anchor does bind to the argument window rather than the file start: ICU anchoring bounds are on unless .withoutAnchoringBounds is passed, and the failure mode if they were not is every call XCTFailing, i.e. loud, not silent. The vacuity hole is therefore closed for every form I could construct. Residual false-rejects are filed separately.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - line: 77
  - rationale: The concern was invisible initialisers; each door I could open now reports rather than returning zero slots.
  - state: answered
- **Widget source scan is now recursive** (P3)
  - continues_uid: jerryfane/herdrup#261-f3
  - detail: Checked at this head: FileManager.enumerator(at:) replaces contentsOfDirectory, and the ported walk over a copy of the tree with a new HerdrWidgets/Views/Extra.swift containing an unframed Image("AppLogoSmall") reports UNFRAMED AppLogoSmall in Extra.swift, where the old single-level listing reported nothing. Sorting by path makes the order deterministic. The depth axis of this finding is closed; the sibling-source-root axis is filed as a new finding.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - line: 60
  - rationale: A subdirectory refactor can no longer drop coverage.
  - state: answered
- **Per-scale pixel dimensions are now asserted exactly; the 78/78/78 catalogue is caught** (P3)
  - continues_uid: jerryfane/herdrup#261-f4
  - detail: Checked at this head against a copy of the tree with AppLogoSmall@3x.png copied over the 1x and 2x slots: the ported assertion reports WRONG SCALE AppLogoSmall.png 78x78 != 26 and AppLogoSmall@2x.png 78x78 != 52, where the previous filename-non-empty check passed. Contents.json declarations are cross-checked against files on disk (existence, then exact points*factor), and the ceiling test separately measures every PNG on disk including one no scale declares, so an undeclared stray raster is still bounded. Baseline on the real tree: zero failures, slot 26pt, rasters 26/52/78 exactly. The reasoning in the docstring is correct: 78px is 26pt at 3x with no slack, so the ceiling cannot be tightened and per-scale exactness is the only remaining tightening available.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - line: 186
  - rationale: The 78/78/78 catalogue that passed a maximum-only check now fails at two scales.
  - state: answered
- **The production comment still asserts the iOS mechanism the test docstring just retracted, so the two artifacts in this PR now contradict each other** (P3)
  - continues_uid: jerryfane/herdrup#261-f5
  - detail: Checked at this head: the retraction landed in the test file only. WidgetAssetResolutionTests.swift:19-26 now says the exact mechanism -- whether iOS rejected the over-resolution, the single unscaled entry, or both -- is NOT established by this file and only a device receipt closes it. AgentLiveActivity.swift:271-273 still states it as fact: 'a Live Activity refuses to render an asset whose resolution exceeds its presentation, silently, with no build error.' git diff 2d8659c..17f9848 touches one file (the test), so this comment is untouched, and the brief's claim that 'the comment and docstring no longer assert the iOS mechanism' holds for the docstring only. The rest of that comment is accurate and valuable (the in-app gallery receipt genuinely cannot observe the Lock Screen surface; the 26/52/78 and 64pt facts check out). Nonblocking, and the underlying causal question stays open until a device Lock Screen receipt lands against the defect record.
  - line: 271
  - rationale: A future reader takes the load-bearing comment next to the code as the repo's position, and it currently asserts a mechanism the same PR's test explicitly declines to claim.
  - state: open
- **bareLiteral rejects legitimate asset-name and initialiser forms, and any comment or string containing Image("X") in a widget source fails the guard** (P3)
  - detail: Executed against mutated copies of AgentLiveActivity.swift with the ported classifier. (1) [A-Za-z0-9_]+ rejects a hyphenated or dotted asset name: Image("app-logo-small") -> UNMEASURABLE, although that is a perfectly legal asset-catalogue name. (2) Image("AppLogoSmall", bundle: .main) -> UNMEASURABLE, and Image(_:bundle:) is the correct initialiser when an asset resolves from a bundle other than main -- exactly the shape a widget reaching into a framework catalogue would use. (3) A custom SF symbol is loaded as Image("MySymbol") and lives in a .symbolset, so the guard would demand Shared/Assets.xcassets/MySymbol.imageset and fail. (4) A comment or string literal containing the form is parsed as a call: adding '// do not ship Image("AppLogo") into this slot again' above line 278 yields UNFRAMED AppLogo. Every case is a loud failure, never a silent pass, so none of it is a correctness hole -- but note the trap in (4): the file documenting this defect class must never quote the defective line, and the test's own docstring (line 6) does exactly that, which is only safe because Tests/ is not scanned.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - line: 78
  - rationale: A guard that rejects legal code trains the next author to edit the guard rather than the code, which is how the guard loses its teeth.
- **Shared/ is compiled into the widget target but never scanned, so a raster reaching the Lock Screen from a shared view is invisible and the vacuity check cannot notice** (P3)
  - detail: project.yml:307-309 gives the HerdrWidgets extension two source roots, `HerdrWidgets` and `Shared`; widgetSources() walks only repoRoot()/HerdrWidgets. Executed: a copy of the tree with Shared/LogoBadge.swift containing an unframed Image("AppLogo") (a 192px raster at 3x) yields zero failures -- PASSED -- because HerdrWidgets sources are non-empty, so XCTAssertFalse(sources.isEmpty) is satisfied and the walk never sees the new file. Shared currently holds only AgentActivityAttributes.swift and AgentActivityState.swift with no Image( calls, so there is no live defect; the exposure is that the obvious place to put a view both the app and the widget draw is the one place the guard does not look. The fix is to walk the widget target's source roots (HerdrWidgets and Shared) rather than one directory.
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - line: 58
  - rationale: This is the same class as the non-recursive scan just fixed, on a different axis: coverage is directory-scoped where the defect is target-scoped.
- **Non-square slots are unsatisfiable under the exact assertion, and the ceiling is loose on the short axis** (P3)
  - detail: points = max(width, height) at line 123-124 collapses the slot to its longest side, and the exactness assertion compares [size.width, size.height] to [expected, expected]. Executed with .frame(width: 26, height: 40): the port reports WRONG SCALE for all three rasters (26x26 != 40, 52x52 != 80, 78x78 != 120), so no correctly sized non-square asset can pass; meanwhile the ceiling test allows max(w,h) <= 40*3 = 120px, which is 4.6x the true 26pt width, so the short axis is under-constrained. Both directions are loud rather than silent, and the only slot in the tree today is square (26x26 with square rasters), so nothing is wrong at this head. The shape that generalises is per-axis: assert size.width == frameWidth*factor and size.height == frameHeight*factor, carrying both frame numbers in Slot instead of max().
  - file: Tests/AgentActivityStateTests/WidgetAssetResolutionTests.swift
  - line: 123
  - rationale: The guard hard-codes squareness in a place that reads as generic, so the first non-square widget raster will look like a test bug rather than a design constraint.
- **The AppLogo 64/128/192 half of the catalogue change is covered by no test at all** (P3)
  - detail: The guard derives its slots from HerdrWidgets sources, and AppLogo is not drawn there: grep over the repo gives exactly one live consumer, App/HerdrApp.swift:622, plus two mentions inside the new comment. So neither test observes this imageset, and I verified it by hand instead: IHDR reads give 64/128/192 square, App/HerdrApp.swift:619-622 frames it at .frame(width: 64, height: 64), which is exactly points*scale at all three scales; there is no competing AppLogo.imageset in App/Assets.xcassets (only AppIcon.appiconset) so no shadowing; and visual inspection of the 64px raster against the 1024px original restored from e50bba3 confirms the same lamb artwork, correctly downscaled rather than cropped or flattened. The residue is that this pairing is unguarded: raising the app header from 64pt to, say, 96pt would draw a 64px raster at 1x with nothing failing, which is a milder form of the defect this PR exists to prevent. Extending the walk to the app target's own sources would cover it.
  - file: Shared/Assets.xcassets/AppLogo.imageset/Contents.json
  - line: 1
  - rationale: The PR changed two imagesets and the guard pins one; the unguarded one is the asset that was 1024px an hour ago.

**Tests Run**
- git rev-parse HEAD -> 17f9848bdef6b005045868d1caa384e234a1b16b; git diff --stat 2d8659c..17f9848 -> 1 file, +132/-63; git diff --stat e50bba3..17f9848 -> 10 files, +283/-7
- Python port of widgetSlots()+testWidgetRastersDoNotOutResolveTheSlotTheyAreDrawnIn+testEveryDeclaredScaleCarriesARasterOfExactlyThatSize at this head -> 1 slot (AppLogoSmall, 26pt, AgentLiveActivity.swift), 0 failures
- Mutation: AppLogoSmall@3x copied into the 1x and 2x slots -> CAUGHT (WRONG SCALE 78x78 != 26, 78x78 != 52)
- Mutation: build-148 replay, real 1024px AppLogo.png restored from e50bba3 with scales [] and Image("AppLogo") in the slot -> CAUGHT (OVER-RES 1024>78, SCALES [])
- Mutation: HerdrWidgets/Views/Extra.swift with an unframed Image("AppLogoSmall") -> CAUGHT (UNFRAMED, recursive enumerator)
- Mutation: unframed Image("AppLogoSmall") placed before the framed one -> CAUGHT (UNFRAMED)
- Mutations: unframed raster followed by Rectangle().frame(width:300), by VStack{}.frame(width:512), and non-literal own frame plus literal sibling frame(400) -> each mis-binds the sibling bound but is CAUGHT by the exactness assertion; equal-valued sibling bound (26) -> PASSED (benign)
- Mutations: Image(decorative:), Image(uiImage: UIImage(named:)), Image("Name", bundle:), Image("app-logo-small") -> all UNMEASURABLE (loud); multi-line Image(\n "AppLogoSmall"\n) and Image(systemName:) with no frame -> PASSED (no false failure)
- Mutation: unframed Image("AppLogo") in Shared/LogoBadge.swift (widget target source root per project.yml:309) -> PASSED, i.e. NOT caught
- Mutation: .frame(width: 26, height: 40) -> CAUGHT at all three scales (non-square slot unsatisfiable)
- Mutation: widget-source comment containing Image("AppLogo") -> CAUGHT (spurious UNFRAMED AppLogo)
- IHDR reads: AppLogo 64/128/192 square, AppLogoSmall 26/52/78 square; e50bba3 AppLogo.png was 1024x1024/26277B
- Visual inspection of AppLogoSmall.png (26px), AppLogoSmall@3x.png (78px), AppLogo.png (64px) and the 1024px original -> same lamb artwork, no placeholder or grey square
- grep 'func test' Tests/AgentActivityStateTests/ -> 13 (matches the claimed baseline count); swift/swiftc/xcodebuild absent, so the suite itself was NOT executed here

Issue #262, opened 2026-09-16
Widget follow-ups from herdrup#260 and #261 reviews (one product decision, five P3)


jerryfane:
Non-blocking leftovers from the exact-head reviews of #260 (`75df3fbd`, 7 findings) and #261 (`2d8659c` then `17f9848`, 5 findings, 4 answered in-PR). Filing so they are tracked rather than rediscovered. Authorized by the owner in session; record: workflow note 168422.

Ordered by what actually costs something, not by review order.

## 1. Three designed states cannot reach the device at all

`LiveActivityController` passes `staleDate: nil` at **both** push sites — [`:122`](../blob/main/App/LiveActivityController.swift#L122) (`start`) and [`:182`](../blob/main/App/LiveActivityController.swift#L182) (`update`) — deliberately, because nothing re-pushes on a cadence, and a stale window that lies is worse than none.

The consequence is not documented anywhere: **every `isStale` branch in the widget is unreachable in production.** That is not one line of dead code, it is three states we designed, drew in the spec, and unit-tested:

- the hollow mark (`StatusMark`, `isStale || isUnconfirmed`)
- `"may need you"` wording (`presentationHeadline`)
- `"No recent update"` / `"no update from <host>"` (`lockDetail`)

`AgentActivityStateTests` proves the *strings and transitions* are right; nothing proves they are ever *shown*, and today they are not. Two honest options, and the choice is a product one:

- **Make them reachable.** Give ActivityKit a real `staleDate` derived from the push cadence the daemon actually has, so the card degrades on its own when the machine goes quiet.
- **Delete them.** Drop the stale branches, the wording, and their tests, and record why the surface cannot express staleness.

Do not leave the third option, which is where we are: code and tests for a state the user cannot see. Note the reviewer flagged the same wording (`"No recent update"`) as having no test *and* no capture — both are symptoms of this, so fix or remove it here rather than separately.

## 2. `WidgetPalette.ground` is dead, and a doc comment still cites it

`ground` (`#13162A`) has no reader since the card began painting `cardGradient`. `ActivityAction`'s doc comment still describes the absent `Approve` sibling as *"an ink fill on a `ground` label"*, so deleting the token without reading that comment would strand it. Remove both, or keep the token and say what still uses it.

## 3. The gallery receipt got weaker in #260, and the pieces to fix it already exist

[...]

PR #264, opened 2026-09-16
Delete the widget stale states nothing can reach

jerryfane:
Closes the first item of #262, which asked for a decision: wire a real `staleDate`, or delete the states that need one. Owner instruction this session: do the recommended thing.

## It was measurable, not a matter of taste

`context.isStale` only becomes true when something supplies a stale date. **Nothing does, at either end:**

| end | evidence |
|---|---|
| client | `LiveActivityController` passes `staleDate: nil` at both push sites, deliberately — a 90s window once marked a **healthy** roster stale and claimed the host was unreachable while the app was connected in the foreground |
| daemon | `src/push/apns.rs` builds the payload with exactly `event`, `timestamp`, `content-state`. There is no `stale-date` key anywhere in the server |

So three designed states — the hedged `N may need you` headline, the hollow `needsYou` mark, and the `last update / no update from <host>` line — rendered in tests and in the DEBUG gallery, and on no device.

## What this removes

- `presentationHeadline(isStale:)` → `presentationHeadline`
- `AgentActivitySummary.line(_:isStale:)` → `line(_:)`, now nothing passes true
- `isStale` threading through `LockScreenView`, `ExpandedHero`, `ExpandedBody`, `StatusMark`
- the gallery's two stale cards

The mark still goes hollow for `markIsUnconfirmed`, which **is** reachable — that's the surviving doubt signal.

The test pinning the stale headline goes with the code it covered. A test asserting unreachable behaviour reports coverage it doesn't have — the same failure mode as the capture that "proved" build 148's logo was fine.

## If a liveness signal ever lands

A daemon heartbeat, or a push carrying `stale-date`: restore these **with** it, not before. Both the model and the controller now say so where the next person will look.

## Verification

Local: 521 tests, 1 skipped, 1 failure — `LiveRecoveryExecutorTests.testExecutorColdStartsAgainstTheLiveServer`, pre-existing and failing identically on `main`.

Not verified locally: the widget target doesn't compile on Linux, so the deletions compile-check in CI.


jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5eb2c2de34b91-1`

**Summary:** Reviewed jerryfane/herdrup PR #264 at exact head df6a2e2c15bbec51561e92e65dd67f44e10a6926 (git log HEAD == df6a2e2, single commit on top of c606de3; 4 files, +70/-102). No prior verdict exists for #264 in GITMOOT_PRIOR_VERDICTS (frozen 2026-09-16T21:43:29Z; it covers #244-#261 only, including the three #248 verdicts that produced the staleDate finding this PR closes). EVIDENCE IS static_only AND NOT BY CHOICE: `swift` is not installed in this seat (`command not found`), the widget target is iOS-only anyway, and /root/herdr is permission-denied (`ls: cannot open directory '/root/herdr': Permission denied`), so I could not independently confirm the server half of the premise. What I did run: git show/diff against HEAD~1, repo-wide greps, and I read Apple's ActivityKit documentation for staleDate and ActivityState.stale directly.

THE DECISION IS RIGHT. Delete, do not wire. I challenged all five of your points and the case holds.

(1) ANY path that sets a stale date — client side is exhaustively enumerated and clean. `ActivityContent(...)` appears exactly twice in the whole repo, LiveActivityController.swift:126 (`Activity.request`) and :186 (`activity.update`), both `staleDate: nil`. There is no third construction site: `Activity<AgentActivityAttributes>` appears only at :19, :40, :50, :95, :199, and HerdrWidgets contains only AgentLiveActivity.swift + HerdrWidgetBundle.swift (which merely declares `AgentLiveActivity()`), so no second target and no notification-service extension mints or updates an activity. Push-token rotation does not touch staleness: `pushTokenUpdates` (:144) yields token bytes only, and nothing in the staleDate/ActivityState.stale documentation ties token rotation to the stale transition. There is no ActivityKit default — `staleDate` is `Date?` and Apple's page says the state changes 'when time reaches the configured stale date'; absent a configured date there is no transition. THE ONE ROUTE YOU CANNOT CLOSE FROM THE CLIENT, and the one I could no
[truncated]
**Severity:** `P2`

**Findings**
- **Four accidental trailing commas + three mis-indented closers left by the deletion; raises this file's compiler floor to Swift 6.1** (P2)
  - detail: Removing the last argument left the comma behind at four call sites: `LockScreenView(hostLabel:..., state: context.state,\n            )` (:10-11), `ExpandedBody(hostLabel:..., state: context.state,\n                            )` (:40-41), and both `StatusMark(..., isUnconfirmed: ...,\n                    )` calls (:47-48, :55-56). Three of the four closing parens are also mis-indented (:41 sits at column 29, :48 and :56 at column 21). A repo-wide grep for `,\s*\n\s*\)` across App, Shared, HerdrWidgets and Sources returns ONLY these four hits, so this is not an existing convention — it is an artifact. Trailing commas in call argument lists are legal only from Swift 6.1 (SE-0439); before that they are a parse error. The macOS job picks `ls -d /Applications/Xcode*.app | sort -V | tail -1` on macos-latest (.github/workflows/ci.yml:120-126) and the repo already assumes Xcode 26 (project.yml:58-61, MetalToolchain download at ci.yml:128-134), so CI should parse it — I am NOT claiming a build break, and I could not compile to prove it either way (no swift in this seat). But you told me CI is the first compile of these deletions, this is the only file in the repo carrying the construct, and it silently raises that file's minimum toolchain for no reason. Four-line cleanup: drop the commas, re-indent the closers.
  - files:
    - HerdrWidgets/AgentLiveActivity.swift:8-11
    - HerdrWidgets/AgentLiveActivity.swift:38-41
    - HerdrWidgets/AgentLiveActivity.swift:44-48
    - HerdrWidgets/AgentLiveActivity.swift:52-56
  - id: f1
- **updatedAt is now a write-only field and its doc comment still advertises the deleted stale copy** (P3)
  - detail: This is the unused field you asked about, and it is real. After this PR nothing READS `updatedAt`: the only two readers were ExpandedBody.age's 'last update' branch and LockScreenView.lockDetail's 'last update / no update from <host>' line, both deleted. It is still written (LiveActivityController.swift:235 `updatedAt: Date().timeIntervalSince1970`), still encoded and `decodeIfPresent`-decoded (Shared:95, :115), and still asserted in the round-trip tests (Tests/AgentActivityStateTests:32, :100). Its doc comment is now false: ':64 When this snapshot was produced (Unix seconds), for honest stale-state copy' — there is no stale-state copy left — and :72-73 still calls these the 'freshness fields'. That is precisely the 'kept as scenery' the commit message refuses. Either delete the field (the hand-written `init(from:)` ignores unknown keys, so a daemon payload still carrying the key stays decodable) or keep it as a deliberate payload-only field and say so in the comment. Keeping it silently with a comment pointing at deleted UI is the worst of the three.
  - files:
    - Shared/AgentActivityState.swift:64-65
    - Shared/AgentActivityState.swift:72-73
    - App/LiveActivityController.swift:235
  - id: f2
- **ExpandedBody.hostLabel is unused (pre-existing, but this PR is the sweep that should have caught it)** (P3)
  - detail: `ExpandedBody` declares `let hostLabel: String` (:140) and its body never reads it: `body` uses `content` and `state.headline`, `content` uses `headline`/`state.question`/`age`, and `age` now uses only state counts. I checked HEAD~1 — the deleted stale branch there used `state.updatedAt`, never hostLabel — so this is PRE-EXISTING dead weight, not introduced by you. Swift does not warn on unused stored properties, so it will sit there forever while both call sites keep feeding it (`context.attributes.hostLabel` at :39 and the gallery's "hetzner-ts" at :586). Drop the property and the two arguments. Everything else in the sweep is clean: `tertiaryInk` is still read (:317), `@ViewBuilder` was correctly dropped from `lockDetail` (now a single expression) and correctly kept on `age` (still branching), the gallery's `card`/`island` signatures lost `isStale` with no orphaned default, `Self.cases[0]` is still used by the surviving always-on card, and the gallery's header comment was updated to drop the STALE card. WidgetGalleryTests only scrolls and attaches screenshots — it asserts no card count — so removing the two stale cards breaks no receipt.
  - files:
    - HerdrWidgets/AgentLiveActivity.swift:140
    - HerdrWidgets/AgentLiveActivity.swift:39
    - HerdrWidgets/AgentLiveActivity.swift:586
  - id: f3
- **The retained .stale branch is right, but its new comment asserts the opposite of the premise the deletion rests on** (P3)
  - detail: Keeping `.stale` in `liveActivity` (:43) and `isLive` (:52) is the correct call and I would not change the code: it is a dedup guard, admitting an unreachable case costs nothing, and treating it as death would re-open the duplicate-banner bug recorded at :23-29. So no, the widget rendering nothing stale-specific does not make this inconsistent — the branch is about identity, not pixels. The COMMENT is the problem: ':38 This branch still admits `.stale` because ActivityKit may set that state for reasons of its own.' Apple's ActivityState.stale page says the transition happens 'At the specified date' of a configured staleDate, and ActivityContent.staleDate says the same from the other side — there is no documented 'reasons of its own'. If that sentence were true, `context.isStale` could also become true on its own and the entire deletion would be unjustified. The comment undercuts the file it is defending. Say what is actually true: unreachable while staleDate is nil at both push sites; retained so that adding a staleDate later cannot resurrect the duplicate banner.
  - files:
    - App/LiveActivityController.swift:35-39
    - App/LiveActivityController.swift:43
    - App/LiveActivityController.swift:52
  - id: f4
- **Wired-vs-deleted: deleted is correct; do not take this back to the owner** (P3)
  - detail: You asked to be told plainly if this should have been wired. It should not. A stale window only carries information if something re-pushes inside it, and the client has no cadence to violate: the sole `update()` caller is `.onChange(of: fullList)` (HerdrApp.swift:2426 region), AgentList is Equatable, and no timer or heartbeat re-pushes anywhere — which is exactly the mechanism that made the 90s window call a healthy blocked roster 'unreachable' while the app was connected in the foreground. Wiring a staleDate therefore requires a liveness signal FIRST (daemon heartbeat push, or a client-side cadence), and the observed defect came from shipping the UI before the signal. Deleting the rendering and leaving the restoration note at Shared:132-140 and LiveActivityController:116-121/:184-185 is the correct ordering. The one thing I'd hold you to: when the signal lands, restore these from this commit rather than rewriting them, and add the test back WITH a reachable fixture.
  - files:
    - App/LiveActivityController.swift:116-126
    - App/LiveActivityController.swift:184-186
  - id: f5

**Needs**
- Independent confirmation that /root/herdr/src/push/apns.rs (and every other APNs payload builder in the daemon) emits no `stale-date` key in the `aps` dictionary — /root/herdr is permission-denied in this seat, so the server half of the premise rests on the author's inspection, not on mine.
- A compile of HerdrWidgets/AgentLiveActivity.swift (CI's macOS job) to close f1 either way — no Swift toolchain exists in this seat and the target is iOS-only.

jerryfane:
> Runtime: `omp`
> Job: `local-review-herdr-ios-review-opus-18d5ebf9bcc67509-1`

**Decision:** `approved`
**Summary:** At c909577 the sole P2 is discharged: all four call sites (LockScreenView, ExpandedBody, both StatusMark) lost the trailing comma and the three closers are aligned with their openers (measured 12/12, 20/20, 16/16, plus untouched 24/24), so the file no longer requires Swift 6.1 / SE-0439. Verified by `git show c909577` (7+/7-, one file, comma+indent only), a repo-wide grep over **/*.swift for `,[ \t]*\n[ \t]*\)` (no matches; the only hits anywhere are Python/YAML where trailing commas are always legal), and a widened `,[ \t]*(//.*)?\n[ \t]*[\)\>\}]` over the four touched files (no matches). The removal is textual, not semantic: each call's argument order still matches its declaration and no argument was added, dropped or reordered. No closer is attached to the wrong call. No other version-gated syntax appears in the diff; the array-literal commas at AgentLiveActivity.swift:512/520 are pre-existing and legal in every version, and dropping @ViewBuilder from lockDetail is sound for a single-expression concrete body. The stale-state deletion itself is unchanged and semantically faithful to its premise (isStale||isUnconfirmed -> isUnconfirmed; the two `|| isStale` action gates; the fleet line that was already inside the else), with every caller migrated - zero `isStale:` labels remain in App/, HerdrWidgets/, Shared/ or Tests/AgentActivityStateTests/. The three carried P3s are all still present at this head and none of them should block: each is a comment/dead-field cleanup, not a behaviour defect. No compile or test run: this box has no swift toolchain, so CI remains the first real compile of the widget target.
**Severity:** `P3`

**Findings**
- **Trailing commas and mis-indented closers are gone; compiler floor restored** (P2)
  - continues_uid: jerryfane/herdrup#264-f1
  - detail: Fixed at all four sites. Ran `git show c909577` (single file, 7 insertions / 7 deletions, the changed lines differ only by the removed `,` and the closer's leading whitespace), then a repo-wide grep over **/*.swift for `,[ \t]*\n[ \t]*\)` which returns no matches, and a widened `,[ \t]*(//.*)?\n[ \t]*[\)\>\}]` restricted to the four touched files, also no matches - so no array/generic/tuple variant survives either. `awk` on the leading whitespace confirms opener/closer alignment: LockScreenView 8/11 both indent 12, ExpandedBody 38/41 both 20, StatusMark 44/48 and 52/56 both 16, and the pre-existing 217/221 both 24. Each `)` is the line immediately after that call's final argument with no intervening token, and DynamicIslandExpandedRegion(.bottom)'s brace still closes on 42, so no closer moved between calls. Argument lists are unchanged semantically: LockScreenView(hostLabel:state:) still matches the declarations at 206/207, ExpandedBody(hostLabel:state:) at 140/141, StatusMark(status:diameter:isUnconfirmed:) at 350/351/352 with its defaults intact. The array-literal trailing commas at 512 and 520 are pre-existing and legal in every Swift version, so nothing else in this file needs 6.1.
  - line: 8
  - rationale: The concern was a silently raised language-version floor, not formatting; a textual grep of the exact shape plus an indentation measurement is sufficient to discharge it without a compiler.
  - state: answered
- **updatedAt is still write-only and its doc comment still advertises the deleted stale copy** (P3)
  - continues_uid: jerryfane/herdrup#264-f2
  - detail: Unchanged at this head; the head commit touched only HerdrWidgets/AgentLiveActivity.swift. Grepped `updatedAt` across the whole repo: it is written at App/LiveActivityController.swift:235 (`updatedAt: Date().timeIntervalSince1970`), carried by the memberwise init (103/115) and decoded at 95, and asserted in tests at Tests/AgentActivityStateTests/AgentActivityStateTests.swift:32 and 100 - but there is no reader anywhere. The only two readers were the `last update` lines this PR deleted from ExpandedBody.age and LockScreenView.lockDetail. Its doc comment at line 64 still reads "for honest stale-state copy", naming a consumer that no longer exists, so the next reader will conclude the widget renders it. Either drop the field (it is optional and decodeIfPresent, so removing it is payload-compatible in the same way it was added) or restate the comment as forward-looking metadata kept for a future liveness signal.
  - file: Shared/AgentActivityState.swift
  - line: 64
  - rationale: A stored-and-encoded field with no reader plus a comment pointing at deleted code is exactly the scenery this PR set out to remove; it does not affect behaviour, so it cannot block.
  - state: open
- **ExpandedBody.hostLabel is now provably unused and two call sites still pass it** (P3)
  - continues_uid: jerryfane/herdrup#264-f3
  - detail: Still open at this head. Read the whole struct (139-203): `hostLabel` is declared at 140 and never referenced in `body`, `content`, `headline` or `age` - the only interpolations there are state.workingCount / state.totalCount at 196. Its last reader was the deleted `no update from \(hostLabel)` branch. It is still passed at AgentLiveActivity.swift:39 (`hostLabel: context.attributes.hostLabel`) and 586 (`hostLabel: "hetzner-ts"`), and Swift emits no warning for an unused stored property, so nothing will surface it. The host label reaches the island through ExpandedHost(hostLabel:) in the trailing region (34/582), so removing the parameter from ExpandedBody loses no information. Note LockScreenView.hostLabel is genuinely still used at 315 - only ExpandedBody's is dead.
  - line: 140
  - rationale: Dead parameter kept alive by two call sites; harmless at runtime, but this PR is the sweep that should have removed it.
  - state: open
- **The .stale branch's new comment contradicts the premise the deletion rests on** (P3)
  - continues_uid: jerryfane/herdrup#264-f4
  - detail: Unchanged at this head. Lines 35-39 assert `context.isStale` "was always false" because staleDate is nil at both ends, then justify keeping `case .active, .stale` with "ActivityKit may set that state for reasons of its own". Those two clauses cannot both hold: ActivityState.stale and ActivityContext.isStale are the same condition surfaced twice, so if ActivityKit can mint .stale unprompted then isStale can be true unprompted and the deleted stale UI was reachable after all; if it cannot, the stated reason for the branch is wrong. The branch itself is right and must stay - it is load-bearing for the swiped-banner defect documented at 23-29 and mirrored in isLive at 50-55 - so this is purely the justification. Say the honest thing: the branch is defensive against an ActivityKit state we do not drive, and it costs nothing, rather than positing a mechanism that would falsify the deletion.
  - line: 38
  - rationale: A comment that undercuts its own PR's premise will be read as evidence against the deletion in the next round; the code is correct, so this cannot block.
  - state: open
- **Wired-vs-deleted re-confirmed: delete remains the right call at this head** (P3)
  - continues_uid: jerryfane/herdrup#264-f5
  - detail: Re-checked rather than assumed. The head commit changes no logic, and the deletion is semantically faithful to its premise at every site: StatusMark's needsYou case went `isStale || isUnconfirmed` -> `isUnconfirmed` (361), both action gates dropped `|| isStale` (146, 266), the `question` and needsYou guards dropped `, !isStale` (164, 227, 232, 300), and age's fleet-count Text at 196 was already inside the deleted `else`, so every non-stale path is byte-identical in behaviour. Every caller migrated: grep for `isStale` across App/, HerdrWidgets/, Shared/ and Tests/AgentActivityStateTests/ finds no labelled argument left, only prose and the unrelated HerdrKit StreamLiveness.isStale(now:timeout:). staleDate is still nil at both ActivityContent sites (126, 186) with the reason recorded at 116-123. No new argument for wiring it appeared at this head.
  - line: 361
  - rationale: The prior round's conclusion holds unchanged; recording the re-check so the next round does not reopen it.
  - state: answered
- **The deletion's load-bearing premise cites src/push/apns.rs, which does not exist in this repository** (P3)
  - detail: The comment at 136-138 rests the whole deletion on "the daemon's APNs payload carries only `event`, `timestamp` and `content-state` (`src/push/apns.rs`), never `stale-date`". That path is absent here - `ls src/push/apns.rs` fails, and a repo-wide grep for `stale-date|stale_date|staleDate` returns only App/LiveActivityController.swift (36, 37, 116, 126, 184, 186) and this comment. So the in-tree half of the premise (the client passes staleDate: nil) is verifiable and verified; the daemon half is a cross-repo claim with no repository named and no way for any future round, or any reader of this file, to check or re-check it. Name the daemon repo and pin the claim to a commit or a contract test, or move it to the docs that already describe the push payload, so the justification for deleting user-visible states does not depend on an unresolvable pointer.
  - file: Shared/AgentActivityState.swift
  - line: 137
  - rationale: Recorded against a locator in the file that makes the claim; the premise is almost certainly true, but its sole citation cannot be discharged from this repository, which is a documentation defect rather than a behavioural one.

PR #268, opened 2026-09-17
Make the Gram refresh do something, and stop the header drifting mid-page

jerryfane:
Two owner-reported bugs on Gram.

## 1. The refresh icon did nothing — two real causes

It was wired straight to `load(initial:false)`, which opens with `guard !isLoading else { return }`. The ambient **6s poll** calls the same function, so a tap landing inside one was **silently discarded**. And on the common case — an unchanged store, answered by digest — `load` returns without altering a pixel, so even a *successful* refresh was indistinguishable from a dead button.

Manual refreshes now go through `manualRefresh()`:

| before | after |
|---|---|
| tap during poll → dropped | remembered, run when the load finishes |
| no visual change on success | spinner while working |
| unchanged store → silence | reports `Up to date` or `N new messages` |

The sidebar affordance routes through the same path, so it can't be swallowed either.

## 2. On macOS the search row drifted to the middle of the page

The page never claimed the offered height and never pinned to top, and **both hosts centre by default** — the split view's detail `HStack(spacing: 0)` (`HerdrApp.swift:1857`) and the compact `ZStack {}` (`:2312`). Full height was *incidental*, coming only from `content`'s greedy ScrollViews — and a greedy child collapses to its ideal height the moment the enclosing layout measures with an **indefinite height proposal**, which is what a Designed-for-iPad Mac split view does while re-measuring columns (window resize, ⌘K rail toggle, divider drag, section remount). The undersized page was then centred, putting its first child mid-window. That's the "for no reason" intermittency.

One modifier fixes it. Note `gramSidebar` already carries exactly this against the same kind of centring parent (`HerdrApp.swift:2047`) — so this is the **second instance of one bug**.

A diagnosis pass falsified the obvious hypothesis first: all eight `content` branches *are* greedy, so no phase/search/message-count state flips it. The missing anchor is the whole cause.

## Receipt

The existing Gram tests assert the search control exists and is hittable — true in the broken layout too. The new test pins its **position** as a fraction of the window, so a vertically-centred page fails on any device or window size.

## Not verified

macOS itself. There's no Mac in CI and no simulator here, so the anchor fix is code-grounded reasoning plus a simulator position assertion; the "sometimes" case needs the owner's Mac.


jerryfane:
> Runtime: `omp`
> Job: `local-review-review-router-18d60a640901c049-1`

**Summary:** Reviewed PR #268 at exact head 347bfa97ae987d4107761be8db6b467c2f228a8a (base 028d9c7, diff = App/GramView.swift +76/-6 and HerdrUITests/GramTests.swift +37). No prior verdict exists for PR #268 in GITMOOT_PRIOR_VERDICTS (75 entries, as_of 2026-09-17T07:16:38Z, none with pull_request==268), so there are no prior findings to answer. EVIDENCE MODE IS static_only AND THAT IS NOT A CHOICE: this seat has no Swift toolchain at all — `swift`, `swiftc` and `xcodebuild` are all absent (verified, see tests_run), /root/.gitmoot/toolchains contains only Go, and the app target is iOS-only regardless. I did execute what could be executed: a faithful MainActor-serialized model of pollLoop/manualRefresh/load to settle the re-entrancy and note-lifetime questions, and the exact algebra of the new position assertion. BUG ONE (dead refresh button) is real and the queue-and-honour mechanism is sound: no loop, no orphaned flag. BUG TWO (header drifting mid-page) is where this PR does not hold up. The 14-line comment at GramView.swift:317-330 names HerdrApp.swift:1857 and :2312 as GramView's centring parents; neither is its parent. On iPad/Mac regular width GramView is instantiated at HerdrApp.swift:1910 inside detailColumn's `ZStack(alignment: .topLeading)` (:1899), already top-anchored and already full-size via `Palette.groundMachine.ignoresSafeArea()` (:1900); in compact it is a TabView page (:2323), which gets a definite page-sized proposal, with the ZStack at :2312 being the TabView's parent, not GramView's. And the blamed mechanism defeats the chosen fix: a flexible frame with `maxHeight: .infinity` and no `idealHeight` cannot materialise a definite height from a nil proposal — it forwards nil and reports the child's ideal, clamping to .infinity being a no-op — so under the indefinite proposal the comment blames, the new modifier collapses with the content; under a definite proposal the content is already greedy (the PR's own diagnosis says all eight branches are). The receipt canno
[truncated]
**Severity:** `P1`

**Findings**
- **The top-anchor fix names parents that do not host GramView, and its mechanism cannot work in the case it blames — bug TWO is likely not fixed** (Reproduce on the Designed-for-iPad Mac build and identify the parent that actually undersizes or re-centres the page (instrument with a GeometryReader/`onGeometryChange` logging the proposal GramView receives during a column re-measure, a ⌘K rail toggle and a divider drag). If the frame is wanted as belt-and-braces, keep it but replace the causal paragraph with what was actually observed, and drop the false references to :1857/:2312 (and while there, :487's stale reference to :1885, which is `.tint(Palette.brand)`). If the collapse really is an indefinite-proposal collapse, the fix has to supply a definite height (a GeometryReader/`containerRelativeFrame` height, or an `idealHeight`), not `maxHeight: .infinity`.)
  - detail: Three independent problems with the same conclusion. (1) WRONG PARENT, iPad/Mac: the HStack(spacing: 0) at HerdrApp.swift:1857 contains `sidebarRail` and `detailColumn` (:1865-1866). GramView is created at :1910 INSIDE `detailColumn`, whose body is `ZStack(alignment: .topLeading)` (:1899) made full-size by `Palette.groundMachine.ignoresSafeArea()` (:1900) and carrying no trailing frame/alignment modifier (verified through :1941). An undersized GramView was therefore ALREADY pinned to that ZStack's top-leading corner — it could not be vertically centred there, so the cited cause cannot produce the reported symptom on the path the owner reported it on. (2) WRONG PARENT, compact: the ZStack at :2312 wraps the TabView (:2316); GramView is a TabView PAGE (:2323) and receives a definite page-sized proposal. Not a centring parent either. (3) MECHANISM CONTRADICTS THE FIX: the comment blames a greedy child collapsing to its ideal under an INDEFINITE height proposal. A flexible frame with `maxHeight: .infinity` and no `idealHeight` behaves identically under that proposal — it forwards nil to its child and reports the child's size clamped to [min,max], and clamping to .infinity is a no-op; no definite height can be produced from a nil proposal. Under a definite proposal the content is already greedy (the PR's own falsification pass establishes all eight branches are), so the modifier is inert there too. Corroboration inside the file: `centered` (:940-942) already applies `frame(maxWidth:.infinity,maxHeight:.infinity)` to the empty/error branches, which did not prevent the report. The only real residual effects of the new modifier are horizontal (full-width claim in the top-leading ZStack) and the extent painted by `.background(Palette.ground…)` at :332. Net: either the diagnosis or the fix is wrong, the true cause of the macOS drift remains unidentified, and the 14-line comment will actively mislead the next reader — in a codebase where these comments are treated as l
[truncated]
  - id: P1-anchor-wrong-parent
  - locator: App/GramView.swift:317-331 (new comment + `.frame(maxWidth:.infinity,maxHeight:.infinity,alignment:.top)`); HerdrApp.swift:1857,1865-1866,1899-1900,1910 (iPad host); HerdrApp.swift:2312,2316,2323 (compact host); GramView.swift:940-942 (`centered` already carries the same modifier)
  - severity: P1
- **testTheSearchRowStaysAtTheTopOfThePage passes on the pre-fix code: wrong host, wrong platform, and a threshold a centred page clears** (Assert absolutely and two-sidedly instead of fractionally: (1) `search.frame.minY - window.minY <= ~88pt` (header band incl. safe area) AND (2) `window.maxY - app.textViews["gram-composer-input"].frame.maxY` inside a small bottom band — together these prove the page SPANS the window, which fails a centred page at any content height, tall or short. (3) Host the receipt where the bug lives: add a DEBUG mock mode that reproduces the offending container (a centring parent that proposes an indefinite height, e.g. the page wrapped in `.fixedSize(horizontal:false, vertical:true)` inside a centring ZStack) so the test FAILS on the pre-fix code — a receipt that passes both before and after pins nothing. (4) Since the report is Mac-only and CI has no Mac, state in the PR that the anchor claim is unverified rather than presenting this test as its receipt.)
  - detail: Three reasons the receipt cannot discriminate. (a) WRONG HOST: `HERDR_SCREENSHOT_MOCK=gram` routes to GramScreenshotHarness, whose body is GramView directly, rendered as the root of RootView's Group — a window root receives a definite full-window proposal, so the greedy content fills and the search row is at the top with or without the new modifier. Neither cited host (split-view detail, compact TabView) is in the tree under test. (b) WRONG PLATFORM: CI runs HerdrUITests on an iPhone simulator (ci.yml:313; the only iPad legs are TerminalResizeTests and TerminalControlTests, :321/:328) and there is no macOS destination in the workflow — so the receipt exercises phoneBody on iOS, where the drift was never reported. The test's own `app.staticTexts["Gram"]` gate confirms this: only phoneBody's `header` renders that title, iPadBody has no title. (c) THRESHOLD: for a vertically centred page of collapsed height H in a window of height W, the row's midY fraction is ((W-H)/2 + h/2)/W with h≈44pt, which falls below 0.25 as soon as H > 0.5W + h — computed: >418pt at W=748, >470pt at W=852, >556pt at W=1024, >727pt at W=1366, i.e. 53-56% of the window — and goes NEGATIVE once H > W (W=852,H=1600 → -0.413). The `gram` fixture renders five messages plus the setup card in a ScrollView whose ideal height is its content height, so a still-centred page can sit comfortably inside the passing region. So the assertion answers 'can 0.25 pass on a still-centred page at any window size' with yes, at every size tested.
  - id: P1-receipt-cannot-fail
  - locator: HerdrUITests/GramTests.swift:176-197; App/HerdrApp.swift:7628-7637 (GramScreenshotHarness), :229-239 (mock is the RootView root), :316-422 (mockView `.gram`); .github/workflows/ci.yml:313,321,328
  - severity: P1
- **The manual acknowledgement is written to the one slot the poll clears, against the file's own twice-documented rule; measured as low as 0.16s on screen** (Give the manual outcome its own `@State` slot with an explicit lifetime (set on completion, cleared by a short cancellable timer or by the next user action), and render it with precedence above or beside the banner rather than beneath a sticky `sendError`.)
  - detail: `load` clears `refreshNote` on every success, including every 6s poll. `manualRefresh` puts 'Up to date' / 'N new messages' in that same slot, so the acknowledgement survives only until the next successful poll, whose phase is unrelated to the tap. Executed model (1000 taps at uniformly random phases of the 6s cycle, 0.15s RTT): median lifetime 3.24s, minimum 0.161s, 12.6% of notes cleared in under 1s, 4.6% in under 0.5s. In those cases the reader gets a ~150ms spinner and a ~200ms note for an unchanged store — the same 'dead button' impression the PR set out to remove, about one refresh in eight. The file already states the correct pattern twice: `markAllNote` and `pageError` were given their own slots with the explicit reason 'the 6-second poll clears refreshNote on every success'. Additionally `bannerView` (:588) reads `sendError ?? markAllNote ?? refreshNote`, and `sendError` is only cleared when the NEXT send starts (:1530), so after any failed send the manual acknowledgement is never shown at all.
  - id: P2-refreshnote-erased-by-poll
  - locator: App/GramView.swift:1297-1302 (write) vs :1352 (`refreshNote = nil` on every successful load) and :588 (banner precedence); the rule stated at :103-105 (`pageError`) and :127-129 (`markAllNote`)
  - severity: P2
- **'Up to date' / 'N new messages' is derived from a list-length delta, which is wrong at the refresh ceiling, on trims/deletes, and for the reader's own post** (Have `load` return (or record) the `changed` flag and the set of newly-present ids, and base the wording on new ids not authored by the owner rather than on a length difference.)
  - detail: `arrived = inboxStore.inbox.messages.count - before`. Three concrete wrong answers: (1) At the ceiling, `limit = min(max(pageSize, count), 500)`, so a reader holding 500 messages re-fetches the newest 500 — N new arrivals push N old ones out, the count is unchanged, and the reader is told 'Up to date' while the list just changed under them. (2) Any server-side trim/delete concurrent with arrivals offsets the delta to zero or negative, again reporting 'Up to date' (negative correctly falls through to 'Up to date', but so does a real arrival that was offset). (3) When the server first reflects the reader's own optimistic post, the server list grows by one and the reader is told '1 new message' about their own text. `load` already has the exact answer on the line above — `apply(answer)` returns `changed`, and the id set is in hand at :1362.
  - id: P2-count-delta-lies
  - locator: App/GramView.swift:1294,1298-1301; ceiling at :1346-1349 and :1400 (`maxRefreshLimit = 500`); the truth is already computed at :1350 (`inboxStore.inbox.apply(answer)` returns `changed`)
  - severity: P2
- **The exact case the PR exists for — a tap landing inside a poll — still shows no acknowledgement while it waits** (Set the working flag when queuing (and clear it in the queued run's `defer`) so the acknowledgement is immediate on every path.)
  - detail: When `isLoading` is true the function sets `refreshQueued` and returns before touching `manualRefreshing`, so the button keeps showing the static arrow for the remainder of the in-flight poll and only then begins to spin. Executed model: 20/1000 uniformly-phased taps (≈RTT/POLL) got no immediate spinner at a 0.15s RTT; the share grows linearly with round-trip time, so on a slow SSH link a substantial fraction of taps look ignored at the instant they are made — which is the complaint.
  - id: P3-queued-tap-silent
  - locator: App/GramView.swift:1287-1291 (early return precedes `manualRefreshing = true` at :1292); control at :650-658
  - severity: P3
- **Pull-to-refresh is still on the swallowed path — the same bug, left unmigrated in the same file** (Route it through `manualRefresh` as well.)
  - detail: `.refreshable` is a reader-initiated refresh. It still calls `load` directly, so it is still discarded by `guard !isLoading` (:1307) whenever it lands inside a poll, and it never reports an outcome. The PR migrated the header button (:646) and the sidebar token (:346) on exactly this reasoning; this third affordance is the same class of bug in the same file. The system refresh control supplies its own spinner, so the visible defect is narrower — silently doing nothing, plus no 'Up to date' — but it is the same defect.
  - id: P3-refreshable-not-migrated
  - locator: App/GramView.swift:936 `.refreshable { await load(initial: false) }`
  - severity: P3
- **A hung RPC leaves the refresh button permanently disabled and spinning; there is no request deadline anywhere in the path** (Bound the manual path (`withTimeoutAfter`/`Task.sleep` race, or a deadline in `call`) and clear the flag on expiry, or drop `.disabled` and let a second tap coalesce through `refreshQueued`.)
  - detail: Answering the cancellation question directly: cancellation CANNOT strand the flag. The `Task {}` at :646 and :346 is unstructured and not bound to view lifetime, so nothing cancels it on disappearance, and `await load(initial:)` is non-throwing, so a cancelled task still returns normally and the `defer` always runs. The real stick is a never-returning round trip: `HerdrClient.call` awaits `transport.roundTrip` with no application-level deadline and `CitadelTransport.roundTrip` adds none, so a wedged SSH channel leaves `manualRefreshing == true` forever. `.disabled(manualRefreshing)` then makes the control permanently untappable — strictly worse than before, when a tap during a wedged load did nothing but the control stayed live. On iPad the section switch remounts GramView and resets @State; on the phone the Gram tab stays mounted, so the dead button persists for the session. Note the pre-existing `isLoading` stick has the same root cause, which is why this is P3 and not higher.
  - id: P3-manualrefreshing-stick
  - locator: App/GramView.swift:1292-1293 (`defer`), :658 (`.disabled(manualRefreshing)`); Sources/HerdrKit/HerdrClient.swift:59-65 (`call` → `transport.roundTrip`, no deadline); Sources/HerdrKit/CitadelTransport.swift:323-331 (no timeout)
  - severity: P3
- **Four scrutiny items checked and CLEARED — no change requested for these** (No action. Recorded so these are not re-litigated.)
  - detail: (1) RECURSION/LOOP: safe. `refreshQueued` is set only when `isLoading` is true, and `load`'s defer clears it, sets `isLoading = false` and spawns the retry inside ONE synchronous MainActor region (View is @MainActor; the file records that inference at :2030), so no interleaving window exists and no flag can be orphaned. Each re-entry consumes a real tap, so the chain cannot self-sustain. Executed model: 50 taps crammed into 0.5s collapse into 4 extra loads, max depth 4, terminating with both flags false; 1000 single-tap runs → max depth 1; 200 tap-storm runs (200 taps/60s) → `refreshQueued` never left set, loads never exceeded taps+polls. (2) GAP UNDER A SHORT LIST: none. `.background(Palette.ground.ignoresSafeArea())` sits OUTSIDE the new frame (:332), so it paints the claimed frame, and `Palette.groundMachine` (HerdrApp:1900) is behind it in the iPad host regardless. (3) COMPOSER `layoutPriority(1)` CONTRACT: unchanged. When the stack's minimum exceeds the proposal, the flexible frame reports the child's overflowing height (max .infinity clamps nothing), so frame height == child height and `alignment: .top` is inert — the overflow behaviour described at :482-489 is exactly what it was. (4) HIT TARGET / ACCESSIBILITY IDENTITY: no existing gram receipt touches this control — GramTests queries `gram-search`, `gram-search-field`, `gram-composer-input`, `gram-send-button` and static texts, and nothing in HerdrUITests references `arrow.clockwise` or 'Refresh' as an element. `.accessibilityLabel("Refresh")` matches the sibling Read-all button (:642). ProgressView(.small) ≈16pt vs the 15pt-semibold Image ≈17pt inside identical button padding, so the hit rect is effectively unchanged; the only future hazard is `.disabled` making the control non-hittable while spinning, which a 'tap refresh twice' receipt would have to account for.
  - id: SCRUTINY-CLEARED
  - locator: App/GramView.swift:1286-1316 (queue/defer), :332 (background), :482-489 + :502-511 (composer priority), :644-659 + HerdrUITests/GramTests.swift:13-159 (control identity)
  - severity: P3

**Tests Run**
- `swift build` — NOT RUN, `error: command not found: swift` (no Swift toolchain in this seat; /root/.gitmoot/toolchains contains only go1.26.4)
- `swiftc -parse App/GramView.swift` — NOT RUN, `error: command not found: swiftc`
- `xcodebuild -version` — NOT RUN, `error: command not found: xcodebuild` (and the app target is iOS-only; the workflow's UI-test legs need a macOS runner, .github/workflows/ci.yml:105)
- `command -v swift swiftc xcrun xcodebuild sourcekit-lsp` → rc=1, no match: build/vet/test of the changed production paths is impossible here, hence evidence=static_only
- `git diff --stat 028d9c7 347bfa9` → App/GramView.swift +76/-6, HerdrUITests/GramTests.swift +37/-0; `git rev-parse HEAD` → 347bfa97ae987d4107761be8db6b467c2f228a8a (exact head confirmed)
- EXECUTED (python, eval kernel) — MainActor-serialized model of pollLoop/manualRefresh/load transcribed from GramView.swift:1267-1316: 1000 single taps at uniformly random phases of the 6s cycle → max re-entrancy depth 1, `refreshQueued` never left set, 20 taps got no spinner while queued; note lifetime min 0.161s / p10 0.824s / median 3.240s / max 6.300s, 12.6% under 1s, 4.6% under 0.5s
- EXECUTED (python) — tap-storm, 200 runs x 200 taps/60s → max depth 11 (each level consuming a real tap), `refreshQueued` stuck in 0/200 runs; 50 taps in 0.5s then silence → 4 extra loads, depth 4, terminates with both flags false (no self-sustaining loop)
- EXECUTED (python) — algebra of the new assertion `((search.midY - window.minY)/window.height) < 0.25` for a centred page: passes while still centred once collapsed height exceeds 418pt@748, 470pt@852, 556pt@1024, 634pt@1180, 727pt@1366, 844pt@1600 (53-56% of window); fraction goes negative (-0.002 / -0.413 / -0.582) when the centred page is taller than the window
- READ-ONLY code checks used as evidence: grep of `load(initial` callers in App/GramView.swift (:830 Retry, :936 .refreshable, :1268/:1273 pollLoop, :1295 manualRefresh); grep of `GramView(` call sites in App/HerdrApp.swift (:1910 detailColumn ZStack(.topLeading), :2323 TabView page, :7635 GramScreenshotHarness); grep of HERDR_SCREENSHOT_MOCK across HerdrUITests; grep of `arrow.clockwise|Refresh` across HerdrUITests (no element references); grep of `destination|HerdrUITests` in .github/workflows/ci.yml (iPhone-simulator-only suite at :313, no macOS destination)
- Prior-verdict check: parsed $GITMOOT_PRIOR_VERDICTS (75 verdicts, as_of 2026-09-17T07:16:38Z) — zero entries with pull_request==268, so no prior findings exist to answer for this PR

**Needs**
- A macOS (Designed for iPad) run of the Gram page — or an instrumented log of the height proposal GramView receives during a split-view column re-measure — to establish the real cause of the drift. This seat has no Swift toolchain and CI has no macOS destination, so the anchor claim cannot be verified by anyone on the current infrastructure.

jerryfane:
> Runtime: `omp`
> Job: `local-review-review-router-18d60b3577a087e0-1`

**Summary:** Reviewed jerryfane/herdrup#268 at exact head b5c70f0212faba7ee64d4778fb1cbeec9ba14dda (git rev-parse matches; delta vs the prior reviewed head 347bfa9 is exactly App/GramView.swift +85/-37-net and HerdrUITests/GramTests.swift -37, i.e. +49/-73). Both P1s are genuinely gone by withdrawal, not by patching: the `.frame(maxWidth:.infinity,maxHeight:.infinity,alignment:.top)` and its 14-line causal comment are absent from GramView.swift (grep for `alignment: .top)` returns nothing) and testTheSearchRowStaysAtTheTopOfThePage is absent from GramTests.swift. Both P2s are really fixed and I verified the mechanisms: the acknowledgement now lives in its own `refreshAck` slot with a token+text-guarded 3s expiry, and the list-length delta is gone in favour of a `LoadOutcome` derived from `apply()`'s `changed` flag and `answer.isUnchanged`, which is the correct source of truth (GramInbox.swift:105-132 — nil messages -> false/unchanged, `fresh != messages` -> changed), so 'Up to date' now fires only when the digest genuinely matched and the 500-ceiling / concurrent-trim / own-optimistic-post cases all report 'Refreshed'.

BUT THIS HEAD DOES NOT COMPILE. `LoadOutcome` is referenced as load's return type at App/GramView.swift:1316 and switched over at :1286-1290, and it is DECLARED NOWHERE. grep -rn over all 229 tracked .swift files returns exactly one occurrence — the use site. The only non-Apple module GramView imports is HerdrKit, and the built HerdrKit.swiftmodule contains 0 occurrences of the string 'LoadOutcome' (control: 'GramInbox' -> 13). `swiftc -typecheck` on a 7-line file mirroring :1315-1318 returns `error: cannot find type 'LoadOutcome' in scope`. The owner's two local green checks cannot see this: `swift build` never compiles App/ (Package.swift declares HerdrKit + test targets only; App/ is exclusively in the Xcode `Herdr` target per project.yml:66-74) and `swiftc -parse` is syntax-only — I ran it on both changed files and both exit 0. CI's `xcodebuild build` (ci.yml
[truncated]
**Severity:** `P0`

**Findings**
- **`LoadOutcome` is never declared — the app target cannot compile at this head, and neither of the owner's two green checks can see it** (Declare the enum next to `LoadPhase` (App/GramView.swift:274) — `private enum LoadOutcome { case unchanged, updated, failed }` — and verify with something that type-checks App/: `xcodebuild build -project Herdr.xcodeproj -scheme Herdr -destination 'platform=iOS Simulator,...'`. `swift build` plus `swiftc -parse` is not a sufficient gate for any change under App/.)
  - detail: The new return type does not exist. `grep -rn "LoadOutcome" --include=*.swift .` over all 229 tracked Swift files returns exactly ONE line — App/GramView.swift:1316, the use site. There is no `enum LoadOutcome` anywhere: the only `*Outcome` declarations in the repo are `SessionRecovery.AdoptOutcome` (Sources/HerdrKit/SessionRecovery.swift:112) and `ClaudeLoginOutcome` (Sources/HerdrKit/AccountLogin.swift:57), neither of which has `.unchanged`/`.updated`/`.failed` and neither of which is named `LoadOutcome`. GramView.swift's imports are CoreTransferable, HerdrKit, PhotosUI, QuickLook, SwiftUI, UIKit, UniformTypeIdentifiers; the only non-Apple one is HerdrKit, and `strings .build/x86_64-unknown-linux-gnu/debug/Modules/HerdrKit.swiftmodule | grep -c LoadOutcome` = 0 against a control of `GramInbox` = 13, so HerdrKit does not export it either. Executed proof of the resulting diagnostic: `swiftc -typecheck` on a 7-line file mirroring :1315-1318 verbatim (`@discardableResult private func load(initial: Bool) async -> LoadOutcome { return .unchanged }`) emits `error: cannot find type 'LoadOutcome' in scope` and exits 1. WHY THE PR'S OWN EVIDENCE MISSES IT: `swift build` is green (I ran it — 'Build complete! (107.12s)') because Package.swift declares HerdrKit and the test targets only, and App/ belongs exclusively to the Xcode application target (project.yml:66-74) — SwiftPM never compiles this file. `swiftc -parse` is green (I ran it on BOTH changed files, exit 0/0) because parsing is syntax-only and an undeclared type is a type-check error, not a parse error. So 'swift build clean, swiftc -parse clean' is exactly the pair of checks that cannot observe this class of defect, and CI's `xcodebuild build -scheme Herdr` at ci.yml:181 is the first thing that will.
  - file: App/GramView.swift
  - line: 1316
  - locator: App/GramView.swift:1316 (`private func load(initial: Bool) async -> LoadOutcome`) and :1286-1290 (`switch await load(initial: false)` over `.unchanged`/`.updated`/`.failed`); Package.swift targets (HerdrKit + test targets only); project.yml:66-74 (App/ is in the Xcode `Herdr` target); .github/workflows/ci.yml:181 (`xcodebuild build`)
  - severity: P0
- **RESOLVED by withdrawal — the anchor modifier and its causal comment are gone; one stale line reference survives nearby** (No action required for the withdrawal. Optionally repoint :476 at HerdrApp.swift:1899 (the `ZStack(alignment: .topLeading)`) so the citation matches the mechanism it asserts.)
  - continues_uid: jerryfane/herdrup#268-f1
  - detail: Checked at this head. The `.frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)` and the whole 14-line paragraph blaming HerdrApp.swift:1857/:2312 are deleted; body now goes Group -> `.background(Palette.ground.ignoresSafeArea())`. I re-verified the host enumeration that made the diagnosis wrong: `grep -rn 'GramView(' App/*.swift` finds exactly three constructions — HerdrApp.swift:1910 (inside `detailColumn`, whose body is `ZStack(alignment: .topLeading)` made full-size by `Palette.groundMachine.ignoresSafeArea()` at :1899-1900), HerdrApp.swift:2323 (a TabView page), and HerdrApp.swift:7635 (GramScreenshotHarness) — so no centring parent hosts the page and the withdrawal is correct. ONE RESIDUAL, unchanged from the prior head and still wrong: the comment at :476 says 'On Mac the split-view host anchors that overflow at the top (HerdrApp.swift:1885)'; HerdrApp.swift:1885 is `.tint(Palette.brand)`. The substantive claim is true (detailColumn's ZStack does anchor top-leading) but the citation points at the wrong line, and this file's line-citations are treated as load-bearing.
  - file: App/GramView.swift
  - line: 321
  - locator: App/GramView.swift:311-336 (body; `grep -n 'alignment: .top)' App/GramView.swift` returns no match, :321 is now `.background(Palette.ground.ignoresSafeArea())` directly after the Group); residual at App/GramView.swift:476 vs App/HerdrApp.swift:1885
  - severity: P1
  - state: answered
- **RESOLVED by removal — testTheSearchRowStaysAtTheTopOfThePage is gone, and the mock host it used is still exercised only by the existing receipts** (No action.)
  - continues_uid: jerryfane/herdrup#268-f2
  - detail: Checked at this head. The test and its 11-line comment are deleted; the file's remaining tests (:13-159) all still use `HERDR_SCREENSHOT_MOCK=gram` and assert existence/hittability/text, which is what that harness can honestly support. Nothing dangles: GramScreenshotHarness (HerdrApp.swift:7628) is still reached from the mockView `.gram` branch at :421, so no dead code was left behind and no other test referenced the removed one. `swiftc -parse HerdrUITests/GramTests.swift` exits 0. The reason the receipt could not discriminate (window-root definite proposal + no macOS destination in ci.yml) is accepted by the owner and needs no further action; the PR correctly no longer presents any receipt for the anchor claim.
  - file: HerdrUITests/GramTests.swift
  - line: 159
  - locator: HerdrUITests/GramTests.swift (file now ends at :162; `grep -c testTheSearchRowStaysAtTheTopOfThePage` = 0); App/HerdrApp.swift:7628 GramScreenshotHarness still referenced from :421
  - severity: P1
  - state: answered
- **Poll-erasure half FIXED and verified; the residual is precedence — refreshAck still sits beneath two slots that are sticky until the next send / next read-all** (Either render the acknowledgement as its own row beside the banner rather than in the same `??` chain, or give `sendError`/`markAllNote` bounded lifetimes of their own so a stale failure cannot mask a live acknowledgement indefinitely.)
  - continues_uid: jerryfane/herdrup#268-f3
  - detail: The measured defect is genuinely gone: the acknowledgement no longer goes in `refreshNote` (which :1363 still clears on every poll success), it has its own @State slot, and the 3s expiry is token-guarded. I re-ran the note-lifetime model at this head and the 0.16s-minimum / 3.2s-median erasure no longer occurs; refreshAck's lifetime is now exactly 3.0s for terminal outcomes and 'until the queued run completes' for the queued one. WHAT IS NOT FIXED is the second half of the original finding, which asked for precedence above a sticky slot. `bannerView` still reads `sendError ?? markAllNote ?? refreshAck ?? refreshNote`. `sendError` is written on every send failure and cleared ONLY at :1543 when the next send starts, and `markAllNote` is written on a partial read-all (:2343) and cleared ONLY at :2285 when the next read-all starts. So once a send has failed or a read-all has partially failed, EVERY subsequent manual-refresh acknowledgement — including the new immediate 'Refreshing…' that f5 exists to deliver — renders nothing at all, and the button is back to being indistinguishable from broken. This is narrower than the original P2 (it needs a prior failure) so it does not deserve P2, but it is the same 'the reader gets no acknowledgement' outcome on a reachable path.
  - file: App/GramView.swift
  - line: 577
  - locator: App/GramView.swift:100-103 (`refreshAck` + `refreshAckToken`), :1304-1313 (`acknowledge`), :577 (`sendError ?? markAllNote ?? refreshAck ?? refreshNote`); sendError cleared only at :1543 (start of the next send); markAllNote cleared only at :2285 (start of the next read-all) and set at :2343
  - severity: P3
  - state: answered
- **RESOLVED — the length delta is gone and the replacement is derived from the right source; I checked all three wrong answers against GramInbox.apply** (No action.)
  - continues_uid: jerryfane/herdrup#268-f4
  - detail: Checked at this head. `arrived = inboxStore.inbox.messages.count - before` and the 'N new message(s)' wording are deleted; the outcome comes from `apply(answer)`'s `changed` plus `answer.isUnchanged`, which is the truth I pointed at. I verified the mapping against apply's actual semantics: a digest-matched answer carries nil messages and returns false (GramInbox.swift:105-117) -> `.unchanged` -> 'Up to date'; any answer carrying messages sets `changed = fresh != messages` (:119) and makes `answer.isUnchanged` false, so `guard changed || !answer.isUnchanged` passes -> `.updated` -> 'Refreshed'. All three original wrong answers are therefore gone: (1) at the 500 ceiling a reader holding 500 who receives N arrivals gets `fresh != messages` -> 'Refreshed', not a false 'Up to date'; (2) a concurrent trim/delete likewise changes the list -> 'Refreshed'; (3) the reader's own optimistic post now produces 'Refreshed' rather than a claim of '1 new message'. One benign imprecision remains and is not worth changing: after any local markRead/remove the digest is cleared, so `conditionalDigest` is nil, the answer always ships a list, and the next manual refresh says 'Refreshed' even when nothing moved — 'Refreshed' claims only that the list was re-fetched, which is true, so it is not the lie the old wording was.
  - file: App/GramView.swift
  - line: 1286
  - locator: App/GramView.swift:1286-1290 (outcome mapping), :1367 (`guard changed || !answer.isUnchanged else { return .unchanged }`), :1378 (`return .updated`), :1386/:1398 (`return .failed`); Sources/HerdrKit/GramInbox.swift:89-132 (`apply`), :41 (`conditionalDigest`)
  - severity: P2
  - state: answered
- **FIXED — a tap landing inside a poll now acknowledges immediately; the acknowledgement is a banner rather than the spinner, and inherits the masking residual in f3** (No action beyond the f3 precedence fix.)
  - continues_uid: jerryfane/herdrup#268-f5
  - detail: Checked at this head, including the lifetime question the owner asked. The queued branch now sets the acknowledgement BEFORE returning, so the tap is never silent on that path. My executed model confirms 'Refreshing…' spans exactly the queued wait and is then replaced by the terminal note: at rtt 0.15s it holds t=0.10-0.25 then 'Up to date' t=0.30-3.30; at rtt 1.0s it holds t=0.55-1.95 then 'Up to date' t=2.00-5.00; at rtt 5.0s it holds t=2.55-9.95 then 'Up to date' t=10.00-12.95. Two honest caveats rather than a re-open: the refresh GLYPH still does not change on the queued path (`manualRefreshing` is not set there), so the acknowledgement is only the banner text; and because `refreshAck` sits third in the `??` chain at :577, the immediate 'Refreshing…' is invisible after any failed send or partial read-all — that is the f3 residual, tracked there, not double-counted here.
  - file: App/GramView.swift
  - line: 1281
  - locator: App/GramView.swift:1277-1283 (`refreshQueued = true; acknowledge("Refreshing…", expiring: false); return`), :639-647 (the control still keys its spinner on `manualRefreshing`, which the queued path deliberately does not set)
  - severity: P3
  - state: answered
- **FIXED for pull-to-refresh; the `.unavailable` Retry button is the same class of bug and is still on the swallowed path** (Route :819 through `manualRefresh()` as well; `initial: true` is not load-bearing there because :1344-1345 already handle the empty/warm-cache phase transitions.)
  - continues_uid: jerryfane/herdrup#268-f6
  - detail: Checked at this head. `.refreshable` now routes through `manualRefresh`, so a pull that lands inside a poll is queued and acknowledged instead of being discarded by `guard !isLoading`; the census above confirms the three reader-initiated affordances the PR set out to fix (sidebar token :[REDACTED], header button :635, pull :927) are all on the acknowledged path. ONE reader-initiated affordance was missed: the Retry button in the `.unavailable` phase at :819 still calls `load(initial: true)` directly, so a tap landing inside the ambient poll is still swallowed by the guard at :1318 and reports nothing. It matters less than the others — it only exists when the list is empty and the poll has just failed, and the failure backoff stretches the poll to 30s (:1262) so the overlap window is proportionally smaller — but it is the same defect the PR migrated three other affordances off, in the same file. Also note that on the queued path `.refreshable`'s system spinner retracts immediately (manualRefresh returns without awaiting the work), which is not a regression — the old `load` early-returned identically — but it does mean the pull gesture's own progress indicator no longer tracks the actual refresh.
  - file: App/GramView.swift
  - line: 927
  - locator: App/GramView.swift:927 (`.refreshable { await manualRefresh() }`); residual at :819 (`Button { Task { await load(initial: true) } }` in the `.unavailable` phase). Full call-site census: `load(initial:` at :819, :1259, :1264, :1286, :1316; `manualRefresh()` at :335, :635, :927, :1325
  - severity: P3
  - state: answered
- **STILL OPEN and now with a second stuck slot: a hung round trip leaves 'Refreshing…' on screen permanently, and if it is the queued run that hangs the control is also permanently disabled** (Bound the manual path (a `Task.sleep` race or a deadline in `HerdrClient.call`) and on expiry clear both `manualRefreshing` and `refreshAck`; alternatively give 'Refreshing…' a generous expiry (or a watchdog) rather than `expiring: false`, so no code path can leave a permanent 'work in progress' claim, and drop `.disabled` so a second tap can coalesce through `refreshQueued`.)
  - continues_uid: jerryfane/herdrup#268-f7
  - detail: Checked at this head; no deadline was added anywhere in the path. I re-confirmed cancellation is NOT the hazard: `load` is non-throwing so its defer always runs, a cancelled `gramList` lands in the generic catch and returns `.failed` -> `acknowledge(nil)`, and the expiry `Task {}` does not inherit cancellation from its creator, so no flag and no note is stranded by teardown. The hang is. Executed model with load #2 never returning: tap-while-the-hung-load-is-running -> isLoading=true, refreshQueued=true, refreshAck='Refreshing…' forever; tap-arriving-later -> identical; and hang on the QUEUED run itself -> manualRefreshing=true AND refreshAck='Refreshing…' forever, which via :647 leaves the control permanently untappable while a banner insists work is in progress. This head makes the symptom worse rather than better in one specific way: `acknowledge("Refreshing…", expiring: false)` is the single acknowledgement deliberately exempted from the 3s expiry, so unlike every other note it has no escape hatch and depends entirely on load's defer firing. On iPad the section switch remounts GramView and resets @State; on the phone the Gram tab stays mounted, so it persists for the session. Still P3 because the pre-existing `isLoading` stick has the same root cause and the trigger is a wedged SSH channel, not ordinary input.
  - file: App/GramView.swift
  - line: 1281
  - locator: App/GramView.swift:1284-1285 (`manualRefreshing = true` + `defer`), :1281 (`acknowledge("Refreshing…", expiring: false)` — the one acknowledgement with NO expiry), :647 (`.disabled(manualRefreshing)`); Sources/HerdrKit/HerdrClient.swift:59-65 (`call` -> `transport.roundTrip`, no deadline); Sources/HerdrKit/CitadelTransport.swift:323-331 (no timeout)
  - severity: P3
  - state: answered
- **Re-verified at this head, still CLEARED — plus the new token guard checked exhaustively and found sound** (No action. If the wording set ever grows, bump the token on every `acknowledge` call (including the nil and non-expiring ones) so correctness stops depending on distinct strings.)
  - continues_uid: jerryfane/herdrup#268-f8
  - detail: All four original items re-checked at this head and none regressed. (1) RECURSION/LOOP: still safe, and re-measured against the new code — 400 seeds x 50 taps crammed into 0.5s across 60 polls gave max chain depth 4, `refreshQueued` and `manualRefreshing` never left set, `refreshAck` nil at t=400 in every seed, and loads never exceeded taps+polls (max 64 against 110 scheduled events). The whole clear-set-spawn sequence is still one synchronous MainActor region (`View` is @MainActor, asserted at :2043) so no flag can be orphaned, and each hop consumes a real tap so the chain cannot self-sustain. (2) GAP UNDER A SHORT LIST: moot — the frame is reverted, `.background(Palette.ground.ignoresSafeArea())` at :321 now applies directly to the Group. (3) COMPOSER `layoutPriority(1)`: untouched by this head (:478, :499 unchanged) and with the frame gone the contract is literally what it was before the PR. (4) CONTROL IDENTITY: unchanged — `.accessibilityLabel("Refresh")` at :648, spinner-vs-glyph swap at :639-645, and no receipt in HerdrUITests queries it. NEW ITEM, also cleared: the token-guarded expiry cannot clear a newer note. I drove `acknowledge` through all 1600 combinations of three consecutive outcomes from {unchanged, updated, failed, queued} at inter-event delays {0.5, 1.0, 2.9, 3.1, 5.9}s and found 0 violations of 'the most recent note is displayed immediately, survives to its own deadline, and is cleared at it'. Note WHY it holds: the token alone is insufficient because `acknowledge(nil)` and the non-expiring 'Refreshing…' deliberately do not bump it, so the `refreshAck == text` comparison is load-bearing — it holds only because 'Refreshing…' can never equal 'Up to date' or 'Refreshed'. That is a real invariant resting on string inequality; if a future outcome string ever collides with a pending one, a stale timer will clear a newer note.
  - file: App/GramView.swift
  - line: 1320
  - locator: App/GramView.swift:1320-1327 (load's defer / queue), :1304-1313 (`acknowledge` + token), :321 (background outside the removed frame), :478/:499 (composer `layoutPriority(1)`), :639-648 (control identity); HerdrUITests/GramTests.swift:13-159
  - severity: P3
  - state: answered

**Tests Run**
- git rev-parse HEAD -> b5c70f0212faba7ee64d4778fb1cbeec9ba14dda (matches the requested head); git diff --stat 347bfa9 b5c70f0 -> App/GramView.swift 85 changed, HerdrUITests/GramTests.swift 37 deleted, 2 files / +49 / -73
- grep -rn 'LoadOutcome' --include=*.swift . -> 1 hit total (App/GramView.swift:1316, the use site); git ls-files | grep -c '\.swift$' -> 229; grep -rn 'Outcome' --include=*.swift . -> only AdoptOutcome (SessionRecovery.swift:112) and ClaudeLoginOutcome (AccountLogin.swift:57). NO DECLARATION EXISTS
- docker run swift:6.1 swiftc -parse App/GramView.swift -> exit 0; swiftc -parse HerdrUITests/GramTests.swift -> exit 0 (confirms the owner's parse-clean claim is blind to the type error)
- docker run swift:6.1 swiftc -typecheck <7-line file mirroring GramView.swift:1315-1318> -> 'error: cannot find type LoadOutcome in scope', exit 1
- strings .build/x86_64-unknown-linux-gnu/debug/Modules/HerdrKit.swiftmodule | grep -c LoadOutcome -> 0 (control: grep -c GramInbox -> 13), so HerdrKit does not export it
- docker run swift:6.1 swift build -> 'Build complete! (107.12s)', exit 0 (and does NOT compile App/: Package.swift declares HerdrKit + test targets only; project.yml:66-74 puts App/ in the Xcode Herdr target)
- docker run swift:6.1 swift test (in-place .build) -> 537 executed, 15 skipped, 0 failures, exit 0. A first run with --scratch-path relocated showed 1 failure (PaneSnapshotAccessTests.testPaneSnapshotExposesNoInitialiserOutsideTheModule, NSFilePath=/w/.build 'file doesn't exist') caused solely by my scratch-path relocation; the in-place re-run is green
- Executed Python model of pollLoop/manualRefresh/load/acknowledge at this exact head, Q1 recursion: 400 seeds x 50 taps/0.5s x 60 polls -> max chain depth 4, refreshQueued never left set, manualRefreshing never left set, refreshAck nil at t=400 in all seeds, max loads 64 vs 110 events
- Same model, Q2 hung round trip (load #2 never returns): tap-while-idle -> refreshAck='Refreshing…' stuck, refreshQueued=true stuck; tap-while-loading -> identical; queued-run-hangs -> manualRefreshing=true AND refreshAck='Refreshing…' stuck (control permanently .disabled)
- Same model, Q3 token+text expiry guard: 1600 exhaustive interleavings (3 consecutive outcomes from {unchanged,updated,failed,queued} x delays {0.5,1.0,2.9,3.1,5.9}s) -> 0 violations of immediate-display / survive-to-own-deadline / cleared-at-deadline
- Same model, Q4 'Refreshing…' lifetime on the queued path: rtt 0.15s -> held 0.10-0.25s then 'Up to date' 0.30-3.30s; rtt 1.0s -> 0.55-1.95s then 2.00-5.00s; rtt 5.0s -> 2.55-9.95s then 10.00-12.95s (persists exactly until the queued run completes)
- grep -rn 'GramView(' App/*.swift -> exactly 3 hosts (HerdrApp.swift:1910 detailColumn, :2323 TabView page, :7635 GramScreenshotHarness), confirming the withdrawn diagnosis; grep -n 'alignment: .top)' App/GramView.swift -> no match (modifier reverted); grep -c testTheSearchRowStaysAtTheTopOfThePage HerdrUITests/GramTests.swift -> 0 (receipt removed)
- NOT RUN, and it is the check that matters: xcodebuild build -scheme Herdr. The app target is iOS-only and this seat has no Xcode or simulator (which swift swiftc xcodebuild -> only docker; the swift:6.1 image has no Apple SDKs). The P0 rests on the undeclared-symbol evidence above, not on a compile of the app target

jerryfane:
> Runtime: `omp`
> Job: `local-review-review-router-18d60bf73782bc7f-1`

**Decision:** `approved`
**Summary:** Reviewed jerryfane/herdrup#268 at exact head c09f43c4a6d549f544a6bd59bad534b28365ec76 (git rev-parse matches). Delta from the prior reviewed head b5c70f0212fa is exactly one file, +10/-0: App/GramView.swift:1315-1324, the restored `private enum LoadOutcome { case unchanged, updated, failed }` plus its 8-line doc comment. Nothing else changed, so every other finding is re-observed at this head rather than re-derived.

P0 (f9) IS DISCHARGED, and I did not take it on reading. `grep -rn LoadOutcome --include=*.swift .` now returns two lines over the 229 tracked Swift files (:1323 declaration, :1326 use) where it returned one. I downloaded swift-6.0.3-RELEASE-ubuntu24.04 into this seat (this box had no toolchain: `command -v swift swiftc xcodebuild` exit 1) and type-checked a structural mirror that reproduces the 10 added lines verbatim inside a @MainActor struct alongside load's five return paths and manualRefresh's switch: `swiftc -typecheck -parse-as-library` exit 0 and `swiftc -emit-sil` exit 0. The mirror is discriminating, proved by five mutations: delete the declaration -> `error: cannot find type 'LoadOutcome' in scope` (reproduces the prior head exactly); add a 4th case -> `error: switch must be exhaustive` at the :1286 switch; delete `return .updated` -> `error: missing return in instance method expected to return 'GramView.LoadOutcome'`; widen `load` to fileprivate -> `error: method cannot be declared fileprivate because its result uses a private type`. So your four scrutiny points check out: the switch has no `default:` and is genuinely exhaustiveness-enforced (a future case breaks the build at the switch, which is what you want); all five return paths return (:1328 isLoading guard, :1377 digest guard, :1388 do-tail, :1396/:1408 both catch arms) and the do/catch needs no trailing return because the second catch is a catch-all; `private` is the right level and the only legal one, since widening `load` alone is a hard error.

PLACEMENT AND PRIVACY: correct and 
[truncated]
**Severity:** `P3`

**Findings**
- **RESOLVED (recorded P0) — the declaration is back, and I proved it with a discriminating type-check rather than by reading the diff** (No action. Treat `xcodebuild build -scheme Herdr` as the only gate for App/ changes; `swift build` cannot compile that target and neither `swiftc -parse` nor `swiftc -typecheck` can see this class of defect (parse is syntax-only, and missing-return is a SIL-stage diagnostic that -typecheck also passes).)
  - continues_uid: jerryfane/herdrup#268-f9
  - detail: Carried at P3 because nothing remains open on this row; it was recorded P0 and is fixed at this head. `grep -rn "LoadOutcome" --include=*.swift .` now returns TWO lines over the 229 tracked Swift files — :1323 `private enum LoadOutcome { case unchanged, updated, failed }` and :1326 the use — where at b5c70f0 it returned only the use. Executed proof, not reading: I installed swift-6.0.3-RELEASE-ubuntu24.04 (this seat had no toolchain; `command -v swift swiftc xcodebuild` exited 1) and built a structural mirror reproducing the 10 added lines verbatim inside a @MainActor struct together with load's five return paths and manualRefresh's switch. `swiftc -typecheck -parse-as-library Mirror.swift` exit 0; `swiftc -emit-sil -parse-as-library Sil.swift -o /dev/null` exit 0. The mirror discriminates — deleting just the declaration line reproduces `error: cannot find type 'LoadOutcome' in scope` (exit 1), i.e. the prior head's exact diagnostic. Also confirmed your account of the blind gates by execution: `swiftc -parse` exits 0 BOTH at this head and on the prior head's file that contains the dangling use, and `swiftc -typecheck App/GramView.swift` exits 1 at `import CoreTransferable` because App/ cannot be type-checked off-Apple. Package.swift has no App/ target and project.yml:65-72 puts App/ only in the Xcode target, so `swift build` never sees it. CI's xcodebuild at ci.yml:178-184 stays the authoritative gate; my evidence bounds the 10 added lines, not the whole file.
  - file: App/GramView.swift
  - line: 1323
  - locator: App/GramView.swift:1323 (declaration) and :1326 (`private func load(initial: Bool) async -> LoadOutcome`); controls: swiftc 6.0.3 -typecheck/-emit-sil on a verbatim mirror, exit 0 baseline vs exit 1 with the declaration removed
  - severity: P3
  - state: answered
- **Withdrawal still stands; the only thing open is the stale HerdrApp.swift:1885 citation at :476, so I am carrying the row at P3** (Repoint :476 at HerdrApp.swift:1899 so the citation names the ZStack that actually performs the top anchoring.)
  - continues_uid: jerryfane/herdrup#268-f1
  - detail: Re-checked at this head; the recorded P1 substance (the `alignment: .top` frame and its 14-line causal comment) is resolved by withdrawal and did not come back: `grep -n "alignment: .top" App/GramView.swift` returns only :958 and :2406, both `HStack(alignment: .top, spacing: 10)`, and body still runs Group -> `.background(Palette.ground.ignoresSafeArea())` at :321. Host enumeration re-confirmed: `grep -rn "GramView(" App/*.swift` gives exactly HerdrApp.swift:1910 (inside detailColumn, whose `ZStack(alignment: .topLeading)` is at :1899 and is made full-size by `Palette.groundMachine.ignoresSafeArea()` at :1900), :2323 (TabView page), :7635 (GramScreenshotHarness) — no centring parent, so withdrawing was correct. THE RESIDUAL IS UNCHANGED AND STILL WRONG: :476 asserts 'On Mac the split-view host anchors that overflow at the top (HerdrApp.swift:1885)', and I printed HerdrApp.swift:1880-1915 — line 1885 is `.tint(Palette.brand)`. The mechanism it claims is real but lives at :1899. This file's line citations are load-bearing (the P0 above was created by an edit span anchored on a doc comment), so a citation that points at a tint modifier is a live papercut, not nothing.
  - file: App/GramView.swift
  - line: 476
  - locator: App/GramView.swift:476 (the citation) vs App/HerdrApp.swift:1885 (`.tint(Palette.brand)`) and :1899 (`ZStack(alignment: .topLeading)`); withdrawal verified by `grep -n 'alignment: .top' App/GramView.swift` and :311-321
  - severity: P3
  - state: answered
- **Still RESOLVED — the withdrawn receipt is absent at this head and the harness it used is still exercised by the four surviving tests** (No action.)
  - continues_uid: jerryfane/herdrup#268-f2
  - detail: Re-checked at this head (the file is untouched by the 10-line diff, but I verified rather than assumed). `wc -l` = 162 and `grep -c testTheSearchRowStaysAtTheTopOfThePage` = 0. The remaining tests are testGramPageRenders (:13), testGramSearchFiltersTheInbox (:43), testGramReadAllClearsTheUnreadCount (:108) and testComposerGrowsUpwardThenScrollsWithoutMovingSend (:131) — all existence/hittability/text assertions the HERDR_SCREENSHOT_MOCK=gram harness can honestly support. Nothing dangles: GramScreenshotHarness (App/HerdrApp.swift:7628) is still reached from the mockView `.gram` branch at :421, so the removal left no dead code. Carried at P3 because the row is closed. Note that I could not run XCUITest here (no xcodebuild on this seat, and ci.yml has no macOS destination), which is the same limitation the owner already accepted.
  - file: HerdrUITests/GramTests.swift
  - line: 162
  - locator: HerdrUITests/GramTests.swift:13-162 (`grep -n 'func test'` = 4 tests, `grep -c testTheSearchRowStaysAtTheTopOfThePage` = 0); App/HerdrApp.swift:421 -> :7628
  - severity: P3
  - state: answered
- **Precedence residual STILL OPEN and broader than recorded: refreshAck is masked by two slots, and sendError has 14 write sites, not just failed sends** (Render the acknowledgement in its own row beside the banner, or give sendError/markAllNote bounded lifetimes so a stale failure cannot mask a live acknowledgement for the rest of the session.)
  - continues_uid: jerryfane/herdrup#268-f3
  - detail: Re-checked and re-executed at this head. `bannerView` at :577 is unchanged: `if let text = sendError ?? markAllNote ?? refreshAck ?? refreshNote`. I re-ran my state-machine model of manualRefresh (:1276-1291) / acknowledge (:1304-1313) / load (:1326-1410) against this head: with `sendError` pre-set, a manual refresh sets `refreshAck = "Up to date"` but the banner renders "Couldn't send. Try again."; with `markAllNote` pre-set it renders "1 of 3 could not be marked read." In both cases the acknowledgement the PR exists to deliver — including the immediate "Refreshing…" from f5 — is invisible, and the button is once again indistinguishable from broken. The masking surface is WIDER than the recorded finding said: `sendError` is cleared only at :1553 (start of the next send) but written at 14 sites, including attachment-skip notes (:1715, :1787), file-open failures (:1847, :1876, :1878, :1946), save failures (:2005, :2007) and delete failures (:2259, :2261) — so merely skipping an over-size attachment or failing one delete is enough to mute every later refresh acknowledgement indefinitely. `markAllNote` is likewise cleared only at :2295 and set at :2353. Line numbers shifted +10 from the prior head; the code did not change.
  - file: App/GramView.swift
  - line: 577
  - locator: App/GramView.swift:577 (`sendError ?? markAllNote ?? refreshAck ?? refreshNote`); sendError writes :1563, :1565, :1627, :1715, :1787, :1847, :1876, :1878, :1946, :2005, :2007, :2259, :2261 cleared only at :1553; markAllNote set :2353 cleared only at :2295; model output: refreshAck='Up to date' while banner shows the stale failure
  - severity: P3
  - state: answered
- **Still RESOLVED — the outcome is derived from apply(answer), and the restored doc comment now records exactly the three wrong answers a length delta gave** (No action.)
  - continues_uid: jerryfane/herdrup#268-f4
  - detail: Re-checked at this head; carried at P3 because the recorded P2 is closed. The mapping at :1286-1290 is unchanged and still sourced from the right truth: `apply(answer)`'s `changed` plus `answer.isUnchanged` via the guard at :1377, `return .updated` at :1388, `.failed` at :1396/:1408. Against GramInbox.apply's real semantics: a digest-matched answer carries nil messages and returns false -> `.unchanged` -> 'Up to date'; any answer carrying messages sets `changed = fresh != messages` and makes `isUnchanged` false -> `.updated` -> 'Refreshed'. NEW at this head and worth noting as a positive: the 8-line doc comment added above the enum (:1315-1322) now records the three cases that made the deleted length delta wrong — the 500 ceiling (`maxRefreshLimit` :1423, applied at :1369-1370), a concurrent trim/delete, and the reader's own optimistic post — so the reasoning is captured in the file instead of only in review. I re-read the comment against the code and every claim in it is accurate.
  - file: App/GramView.swift
  - line: 1286
  - locator: App/GramView.swift:1286-1290, :1315-1322 (new doc comment), :1367-1370, :1377, :1388, :1423; Sources/HerdrKit/GramInbox.swift apply()/conditionalDigest
  - severity: P3
  - state: answered
- **Still FIXED — the queued tap is acknowledged before returning; re-measured note spans at this head** (No action beyond the f3 precedence fix.)
  - continues_uid: jerryfane/herdrup#268-f5
  - detail: Re-checked and re-modelled at this head. :1277-1283 still sets `refreshQueued = true` and calls `acknowledge("Refreshing…", expiring: false)` BEFORE returning, so a tap landing inside a poll is never silent on that path. My executed model reproduces the same lifetimes as the prior round: 'Refreshing…' spans exactly the queued wait and is then replaced by the terminal note. The two caveats stand and are tracked elsewhere, not double-counted: the glyph does not change on the queued path because `manualRefreshing` is deliberately not set there (control at :639-648), and the immediate acknowledgement is invisible behind a sticky sendError/markAllNote (f3). Carried at P3, closed.
  - file: App/GramView.swift
  - line: 1281
  - locator: App/GramView.swift:1277-1283, :639-648; model: 'Refreshing...' held for the queued wait then replaced by 'Up to date' at rtt 0.15/1.0/5.0s
  - severity: P3
  - state: answered
- **Retry at :819 STILL bypasses manualRefresh, and I reproduced the swallowed tap at this head (1 load for 2 taps)** (Route :819 through `manualRefresh()`; `initial: true` is not load-bearing because :1354-1355 already handle the empty/warm-cache phase transitions.)
  - continues_uid: jerryfane/herdrup#268-f6
  - detail: Re-checked and executed at this head. The three migrated affordances are still on the acknowledged path — sidebar token `Task { await manualRefresh() }` at :335, header button at :635, and `.refreshable { await manualRefresh() }` at :927 — but the Retry button in the `.unavailable` phase is unchanged at :819: `Button { Task { await load(initial: true) } }`. Executed reproduction with my model: two taps scheduled 0.1s apart inside a 0.5s round trip produce loads_actually_run = 1 and banner = nil on the :819 path, versus loads_actually_run = 2 when the identical tap goes through `manualRefresh()`. So a Retry that lands inside the ambient poll is still discarded by the guard at :1328 and reports nothing — the same defect the PR migrated three other affordances off, in the same file. Mitigating: it only exists when the list is empty and the poll has just failed, and the failure backoff stretches the poll to 30s (:1262), so the overlap window is proportionally smaller.
  - file: App/GramView.swift
  - line: 819
  - locator: App/GramView.swift:819 (`Button { Task { await load(initial: true) } }`) vs :335/:635/:927; guard at :1328; backoff at :1262; model: 2 taps -> 1 load via :819, 2 loads via manualRefresh
  - severity: P3
  - state: answered
- **STILL OPEN — no deadline was added; a hung round trip strands 'Refreshing…' forever and, on the queued run, permanently disables the control** (Bound the manual path (a Task.sleep race or a deadline in HerdrClient.call) and clear both `manualRefreshing` and `refreshAck` on expiry; or give 'Refreshing…' a generous expiry/watchdog instead of `expiring: false` so no path can leave a permanent 'work in progress' claim, and drop `.disabled` so a second tap coalesces through `refreshQueued`.)
  - continues_uid: jerryfane/herdrup#268-f7
  - detail: Re-checked at this head and re-reproduced. No deadline exists anywhere in the path: `HerdrClient.call` is `let response = try await transport.roundTrip(line)` with no timeout, `CitadelTransport.roundTrip` (:312-331) has none either, and the only `timeout` in that file is the CONNECT budget at :62 — nothing bounds a channel that accepts the exec and then stops producing output. Executed model with the round trip never returning: (a) hang on the in-flight load -> isLoading=true, refreshQueued=true, banner='Refreshing...' at t=600s; (b) hang on the QUEUED run -> manualRefreshing=true AND banner='Refreshing...' at t=600s, which via `.disabled(manualRefreshing)` at :647 leaves the control permanently untappable while the banner insists work is in progress. `acknowledge("Refreshing…", expiring: false)` at :1281 is the single acknowledgement deliberately exempted from the 3s expiry, so it has no escape hatch and depends entirely on load's defer firing. Cancellation is NOT the hazard (re-confirmed): load is non-throwing so its defer always runs, a cancelled gramList lands in the generic catch and returns `.failed` -> `acknowledge(nil)`. On iPad the section switch remounts and resets @State; on the phone the Gram tab stays mounted, so it persists for the session. Still P3: the trigger is a wedged SSH channel and the pre-existing isLoading stick shares the root cause.
  - file: App/GramView.swift
  - line: 1281
  - locator: App/GramView.swift:1281, :1284-1285, :647, :1330-1337; Sources/HerdrKit/HerdrClient.swift call() (`try await transport.roundTrip(line)`, no deadline); Sources/HerdrKit/CitadelTransport.swift:312-331 (no timeout; only the connect budget at :62); model at t=600s: banner='Refreshing...', manualRefreshing=True
  - severity: P3
  - state: answered
- **Re-verified at this head, still CLEARED — 400-seed stress shows no orphaned flags and bounded loads; the token guard's dependence on distinct strings is unchanged** (No action. If the wording set grows, bump the token on every `acknowledge` call (including the nil and non-expiring ones) so correctness stops depending on distinct strings.)
  - continues_uid: jerryfane/herdrup#268-f8
  - detail: All items re-checked at this head. (1) RECURSION/LOOP: re-ran the stress against this head's code — 400 seeds x 50 taps crammed into 0.5s across 60 polls: zero violations, `refreshQueued` and `manualRefreshing` never left set, `refreshAck` nil at t=400 in every seed, and loads bounded at 70 against 110 scheduled events. The clear-set-spawn sequence in load's defer (:1330-1337) is still one synchronous MainActor region, so no flag can be orphaned. (2) GAP UNDER A SHORT LIST: moot — the frame is reverted and `.background(Palette.ground.ignoresSafeArea())` at :321 applies directly to the Group. (3) COMPOSER `layoutPriority(1)` at :478 and :499: untouched by this head. (4) CONTROL IDENTITY: unchanged — `.accessibilityLabel("Refresh")` at :648, spinner/glyph swap at :639-645, no receipt queries it. (5) TOKEN GUARD: unchanged at :1304-1313 and still sound, still for the same fragile reason — `acknowledge(nil)` and the non-expiring 'Refreshing…' do not bump the token, so the `refreshAck == text` comparison is load-bearing and holds only because 'Refreshing…', 'Up to date' and 'Refreshed' are pairwise distinct strings.
  - file: App/GramView.swift
  - line: 1304
  - locator: App/GramView.swift:1304-1313, :1330-1337, :321, :478/:499, :639-648; stress: 400 seeds, 0 violations, max 70 loads / 110 events
  - severity: P3
  - state: answered
- **NEW (latent): `.unchanged` now means both 'ran and nothing moved' and 'did not run at all', which is exactly what the new doc comment promises it does not do** (Add a `case skipped` returned at :1328 and map it in manualRefresh to the same queue-and-acknowledge behaviour as :1277-1283 (or return `LoadOutcome?` with nil for 'did not run'). The switch is exhaustiveness-checked, so the compiler will point at every place that must decide.)
  - detail: Introduced by this head's 10 lines in the sense that the restored doc comment at :1315 declares the type to be 'What a load actually did, so a manual refresh can say something TRUE' — and one of its three cases is not that. The early return at :1328 (`guard !isLoading else { return .unchanged }`) reports 'nothing moved' for a load that never touched the network, and manualRefresh maps it to the reader-visible 'Up to date' (:1287). You are right that it is unreachable TODAY: manualRefresh checks `isLoading` at :1277 and, because both functions are @MainActor and the `await` at :1286 is a same-executor direct async call, no other MainActor work can interleave between the check and load's guard, so the only reachable producers of `.unchanged` are the digest guard at :1377 [INFERENCE on the suspension semantics — I cannot run Swift concurrency in this sandbox, `-emit-sil` fails with `could not build C module 'SwiftGlibc'` because /usr/include is unreadable here]. Two ways it becomes real: any future caller that reads the result (the type is `@discardableResult`, so :819/:1259/:1264 discard it silently today), or any future ungated entry point — note that the sidebar token path (:335) and `.refreshable` (:927) are NOT gated by `.disabled(manualRefreshing)` the way the header button at :647 is, so two concurrent `Task { await manualRefresh() }` bodies already exist as callers and only the no-suspension property keeps them from racing. My type-check control shows the compiler will help you here: adding a fourth case produces `error: switch must be exhaustive` at :1286, so a `.skipped` case cannot be added silently.
  - file: App/GramView.swift
  - line: 1328
  - locator: App/GramView.swift:1328 (`guard !isLoading else { return .unchanged }`) vs :1315 doc comment and :1287 ('Up to date'); ungated callers at :335 and :927 vs the gated one at :647; control B: `swiftc -emit-sil` on a 4-case mirror -> `error: switch must be exhaustive` at the :1286 switch
  - severity: P3

**Tests Run**
- git rev-parse HEAD -> c09f43c4a6d549f544a6bd59bad534b28365ec76 (matches the requested head); git diff --stat b5c70f0212faba7ee64d4778fb1cbeec9ba14dda HEAD -> 1 file changed, 10 insertions(+), 0 deletions(-), App/GramView.swift only
- grep -rn "LoadOutcome" --include=*.swift . -> 2 hits (App/GramView.swift:1323 declaration, :1326 use) across 229 tracked Swift files; was 1 hit at the prior head
- command -v swift swiftc xcodebuild sourcekit-lsp -> exit 1, no toolchain on this seat; installed swift-6.0.3-RELEASE-ubuntu24.04 from download.swift.org; swift --version -> 'Swift version 6.0.3 (swift-6.0.3-RELEASE), Target: x86_64-unknown-linux-gnu'
- swiftc -typecheck -parse-as-library Mirror.swift (verbatim copy of App/GramView.swift:1315-1323 inside a @MainActor struct with load's five return paths and manualRefresh's switch) -> exit 0
- swiftc -emit-sil -parse-as-library Sil.swift -o /dev/null (same mirror, SIL-stage diagnostics enabled) -> exit 0, no missing-return and no exhaustiveness error
- CONTROL A, declaration line deleted (= prior head b5c70f0) -> exit 1, 'error: cannot find type LoadOutcome in scope' — reproduces the P0 diagnostic
- CONTROL B, 4th enum case added -> exit 1, 'error: switch must be exhaustive' at the mirror of App/GramView.swift:1286 — the switch is genuinely exhaustiveness-enforced with no default:
- CONTROL C, 'return .updated' deleted -> exit 1, 'error: missing return in instance method expected to return GramView.LoadOutcome' — every return path in the baseline is therefore covered; note the same mutation exits 0 under -typecheck, so -typecheck alone is also an insufficient gate
- CONTROL D, load() widened to fileprivate while the enum stays private -> exit 1, 'error: method cannot be declared fileprivate because its result uses a private type' — private is the only legal level for load given a private nested result type
- CONTROL E, isLoading guard deleted -> exit 0 (confirms the guard is not what the other controls are detecting)
- swiftc -parse App/GramView.swift @ this head -> exit 0, 0 errors; swiftc -parse on `git show b5c70f0:App/GramView.swift` (which contains the dangling LoadOutcome use and no declaration) -> exit 0, 0 errors: proof that -parse cannot observe this defect class
- swiftc -typecheck App/GramView.swift -> exit 1, 'App/GramView.swift:1:8: error: no such module CoreTransferable' — App/ is not type-checkable on Linux; Package.swift declares only HerdrKit/HerdrKitTests/AgentActivityState/AgentActivityStateTests (no App/ path) and project.yml:65-72 puts App/ only in the Xcode Herdr target
- swift build -> exit 1, 'error: Error Domain=NSCocoaErrorDomain Code=513 You don't have permission' (read-only workspace, cannot create .build); NOT RUN and NOT INFERABLE: xcodebuild build -scheme Herdr and the XCUITest receipts — no Xcode on this seat. Swift codegen/linking also unavailable here: swiftc -emit-sil on the @MainActor+print variant fails with 'could not build C module SwiftGlibc' / 'cannot open /usr/include/stdc-predef.h: Permission denied'
- Executed Python state-machine model of manualRefresh (:1276-1291) / acknowledge (:1304-1313) / load (:1326-1410) / bannerView precedence (:577) / Retry (:819) at this head: f7 hang-on-queued -> banner='Refreshing...' and manualRefreshing=True at t=600s; f7 hang-on-inflight -> banner='Refreshing...', refreshQueued=True at t=600s; f3 sticky sendError -> refreshAck='Up to date' but banner="Couldn't send. Try again."; f3 sticky markAllNote -> banner='1 of 3 could not be marked read.'; f6 two taps 0.1s apart inside a 0.5s round trip -> 1 load via :819 vs 2 loads via manualRefresh
- f8 stress at this head: 400 seeds x (50 taps within 0.5s + 60 polls) -> 0 violations, refreshAck nil at t=400 in every seed, refreshQueued/manualRefreshing never left set, max 70 loads against 110 scheduled events
- grep -n 'alignment: .top' App/GramView.swift -> only :958 and :2406 (HStack row layouts); body at :311-321 goes Group -> .background(Palette.ground.ignoresSafeArea()); grep -rn 'GramView(' App/*.swift -> HerdrApp.swift:1910, :2323, :7635; sed -n '1880,1915p' App/HerdrApp.swift -> :1885 is '.tint(Palette.brand)', :1899 is 'ZStack(alignment: .topLeading)'
- wc -l HerdrUITests/GramTests.swift -> 162; grep -c testTheSearchRowStaysAtTheTopOfThePage -> 0; grep -n 'func test' -> 4 tests at :13/:43/:108/:131; GramScreenshotHarness still referenced App/HerdrApp.swift:421 -> :7628
**Diagnostics:** workflow advance failed: review approval held: no observed pull request row records a current head for jerryfane/herdrup#268, so this approval cannot be shown to be bound to it

PR #281, opened 2026-09-21
fix(tests): use the platform socket helper so the suite builds on macOS

jerryfane:
One line: `socket(AF_INET, Int32(SOCK_STREAM.rawValue), 0)` becomes `socket(AF_INET, sockStream, 0)`.

`SOCK_STREAM` is an `Int32` on Darwin and an enum with `.rawValue` on Glibc, so the file compiled on Linux CI and failed on every macOS checkout. `Sources/HerdrKit/PlatformSocket.swift` already exposes `sockStream` for this exact case — the call site simply did not use it.

Verification on the macOS builder: before, `swift test` fails to compile; after, it builds and runs 543 tests. Four then fail for macOS environment reasons unrelated to this change (a `/private/var` vs `/var` symlink comparison in GramFileCacheTests, and two GramStagingTests denied by TCC over SSH) — filed separately; this PR does not touch them.


jerryfane:
> Runtime: `omp`
> Job: `local-review-review-router-18d748cef7dfbcc2-1`

**Decision:** `approved`
**Summary:** PR #281 at exact head 14d747d856d229440bfeb20d1f6b3dea1f6049e5 (verified against GitHub API: open, base 667f386 on main). One-line test-only change: CitadelTransportTests.swift:74 replaces Int32(SOCK_STREAM.rawValue) with the existing public sockStream helper. sockStream is Int32 on both platforms (PlatformSocket.swift:14-18: Darwin SOCK_STREAM is already Int32; Glibc branch is the identical Int32(SOCK_STREAM.rawValue)), matching socket(2)'s Int32 type parameter on both. No production code changed; no other raw SOCK_STREAM spelling remains outside the helper; usage matches the established convention in HerdrKitTests.swift, LiveEnvironment.swift, PairingTests.swift. Remaining bare close/connect calls are in a private static func (no XCTestCase shadowing) and are pre-existing. No prior verdicts for #281. EVIDENCE static_only: no Swift toolchain exists in this seat (which swift empty, only Go toolchain installed), so swift build/test and the macOS compile could not be executed; verdict rests on executed git/grep/API checks plus type-level inspection of the helper and call site.

**Tests Run**
- git rev-parse HEAD -> 14d747d856d229440bfeb20d1f6b3dea1f6049e5 (matches pinned head)
- git diff 667f386..14d747d --stat -> 1 file changed, Tests/HerdrKitTests/CitadelTransportTests.swift only, +1/-1; no production files
- curl api.github.com/repos/jerryfane/herdrup/pulls/281 -> head.sha 14d747d, base 667f386 main, state open
- grep SOCK_STREAM repo-wide -> only Sources/HerdrKit/PlatformSocket.swift retains the symbol; no inline rawValue spelling remains
- grep sockStream repo-wide -> helper defined public in PlatformSocket.swift:14-18; consumed by Pairing.swift, Transport.swift, and 4 test files including the changed one
- grep for bare bind/connect/close in Tests/ -> only close/connect inside private static func blackHoles (lines 76,81) and pre-existing sites; unchanged by this PR
- which swift / toolchain scan -> absent; swift build and swift test NOT runnable, evidence downgraded to static_only

